PR #546, opened 2026-04-25
fix(search): CSP-7-Soft product import bug

jsboige:
## Summary
- Fix `from itertools import product` being inside `SoftCSP.solve_brute_force` method instead of at module level
- Move the import to the top-level setup cell where it belongs
- Clear stale execution outputs from CSP-7-Soft notebook

## Root cause
The `product` function from `itertools` was imported locally inside the `solve_brute_force` method but referenced at module level scope, causing a `NameError` when the method tried to use `product(*domains)`.

## Test plan
- [x] Verified fix with `python -c "from itertools import product; ..."` execution
- [ ] Execute CSP-7-Soft notebook end-to-end to confirm all cells pass

### DIFF
--- a/MyIA.AI.Notebooks/Search/Part2-CSP/CSP-7-Soft.ipynb
+++ b/MyIA.AI.Notebooks/Search/Part2-CSP/CSP-7-Soft.ipynb
@@ -84,7 +84,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "setup-soft",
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (approved):
Cell consolidation in CSP-7-Soft.ipynb: merged the install cell and the import cell into one cell, adding `from itertools import product` (fixes audit #362 Phase 2 finding). Execution outputs cleared (clean state). Content preserved — no pedagogical material removed. Approving.

PR #592, opened 2026-04-28
fix(qc): remove look-ahead bias in Temporal-CNN backtest

jsboige:
## Summary

- Fix look-ahead bias in `Temporal-CNN-Prediction/research.ipynb` backtest function
- The function was using `returns[i]` for both signal computation AND PnL at the same step, creating artificial predictive power (Sharpe 3.79)
- Fixed by computing signal from `returns[i-k:i]` (prior days only), producing realistic metrics (Sharpe 0.314)
- Replace misleading `[DATA-ONLY]` disclaimer with accurate note about simplified implementation scope

## Metrics change

| Metric | Before (biased) | After (fixed) |
|--------|----------------|---------------|
| Sharpe | 3.790 | 0.314 |
| CAGR | 44.87% | 5.59% |
| Max DD | -9.31% | -20.47% |

## Context

Identified during anti-slop audit (dashboard 28/04). The CNN backtest was one of the "too good to be true" signals — Sharpe 3.79 on a simplified implementation was a red flag. Root cause: signal used same-day return for both prediction and PnL.

## Test plan

- [x] Papermill execution: 14/14 cells pass
- [x] No `raise NotImplementedError` or error cells
- [x] Outputs present for all code cells
- [ ] Review: verify Sharpe ~0.3 is realistic for simplified CNN 1D


jsboige (commented):
ClusterManager review — **LGTM**

Correct look-ahead bias fix. The convolution window shift from `returns[i-len(kernel)+1:i+1]` (included current return) to `returns[i-len(kernel):i]` (only past returns) eliminates future information leakage.

This is a real bias that would have inflated backtest performance — good catch.

Disclaimer updated to accurately reflect simplified CNN (no weight learning). OK to merge.

PR #617, opened 2026-04-29
ci(lean): add differentiated CI for social_choice_lean

jsboige:
## Summary

- New GitHub Actions workflow `lean-social-choice.yml` with 2 jobs:
  1. **fail-on-sorry**: grep for `sorry` in certified modules (Arrow, Sen, Framework, Basic). Fails if any found. Voting.lean (3 sorry, Median Voter Theorem sketch) is intentionally excluded.
  2. **build**: `lake build` with Mathlib olean caching, runs only after sorry check passes
- Triggered on pushes/PRs to main that modify `.lean`, `lakefile.lean`, or `lean-toolchain` in `social_choice_lean/`

## Design decisions

- **Two separate jobs** instead of one: the sorry check is fast (~1s), while the full build takes ~5min with cache. This gives quick feedback on proof integrity.
- **Voting.lean excluded**: The 3 sorry in the Median Voter Theorem are an intentional sketch, not a regression. The CI should not block on them.
- **Cache key**: Based on `lakefile.lean` + `lean-toolchain` hash to avoid stale oleans.

## Part of

Mission 3 (PR-I) from Lean Social Choice integration sprint.

### DIFF
--- a/.github/workflows/lean-social-choice.yml
+++ b/.github/workflows/lean-social-choice.yml
@@ -1,0 +1,94 @@
+name: Lean Social Choice CI
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - 'MyIA.AI.Notebooks/GameTheory/social_choice_lean/**.lean'
+      - 'MyIA.AI.Notebooks/GameTheory/social_choice_lean/lakefile.lean'
+      - 'MyIA.AI.Notebooks/GameTheory/social_choice_lean/lakefile.toml'
+      - 'MyIA.AI.Notebooks/GameTheory/social_choice_lean/lean-toolchain'
+  pull_request:
+    branches: [ main ]
+    paths:
+      - 'MyIA.AI.Notebooks/GameTheory/social_choice_lean/**.lean'
+      - 'MyIA.AI.Notebooks/GameTheory/social_choice_lean/lakefile.lean'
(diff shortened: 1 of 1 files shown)


jsboige (commented):
## Review — Lean Social Choice CI

**Verdict: LGTM** — Two-job CI workflow tailored for social_choice_lean.

1. **fail-on-sorry job**: Greps for `sorry` in certified modules (Arrow, Sen, Framework, Basic). Exits 1 if any found. Ensures formal proofs don't have holes.

2. **build job**: Installs elan, caches Mathlib oleans, runs `lake -R build`. Depends on fail-on-sorry (builds only if no sorry).

3. **Path filtering**: Triggers on changes to `**.lean`, `lakefile.lean`, `lakefile.toml`, `lean-toolchain` within `social_choice_lean/`. Avoids unnecessary CI runs.

4. **Cache key**: `lake-{os}-{hash(lakefile.lean + lean-toolchain)}`. Correct — rebuilds on toolchain or dependency changes.

Points de vigilance (non bloquants):
- `working-directory` is set at job level AND in path filters. The elan install step uses `./elan-init` without cd, which runs in the repo root. Should work since elan installs globally.
- `lake -R build 2>&1 | tail -20` — truncates build output to last 20 lines. If the build fails early, the actual error may not be visible in CI logs. Consider `tee` to a file or increasing tail.

PR #646, opened 2026-05-01
feat(catalog): CATALOG-STATUS markers for remaining 7 series (#623 B-5)

jsboige:
## Summary
- Add CATALOG-STATUS HTML comment metadata blocks to the 7 remaining series READMEs
- All 11 series now have catalog markers (complements PR #645 for the 4 pilots)

## Series injected

| Series | Count | Breakdown |
|--------|-------|-----------|
| SymbolicAI | 90 | SmartContracts=27, SemanticWeb=18, Lean=13, Planners=13, Tweety=10, Argument_Analysis=7, _root=2 |
| QuantConnect | 93 | projects=47, Python=45, ML-Training-Pipeline=1 |
| GameTheory | 26 | _root=26 |
| Probas | 22 | Infer=20, _root=2 |
| IIT | 1 | _root=1 |
| RL | 6 | _root=6 |
| EPF | 4 | IA101-Devoirs=4 |

## Test plan
- [x] `python count_notebooks_by_series.py --check-readme` confirms all counts
- [x] Markers verified via grep on each README
- [x] HTML comment format invisible in rendered markdown
- [ ] CI workflow can consume markers (requires PR #645 script merge first)

Part of #623 Track B-5. Depends on PR #645 for the verification script.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #645, opened 2026-05-01
feat(catalog): CATALOG-STATUS markers for pilot READMEs (#623 B-5)

jsboige:
## Summary
- Add `verify_catalog_readme.py` script that verifies and injects `CATALOG-STATUS` HTML comment metadata blocks in series READMEs
- Inject markers for 4 pilot series: **Search, Sudoku, GenAI, ML**
- Enable CI drift detection between README-declared counts and actual filesystem scans

## Drift detected (3/4 pilots)
| Series | README declared | Actual | Delta |
|--------|----------------|--------|-------|
| Search | 40 | 45 | +5 (_root deprecated notebooks) |
| Sudoku | 33 | 32 | -1 |
| GenAI | 105 | 99 | -6 |
| ML | 30 | 30 | 0 (OK) |

## Test plan
- [x] `python verify_catalog_readme.py` — detects drift correctly
- [x] `python verify_catalog_readme.py --inject` — injects markers
- [x] Post-injection verification: all 4 pilots OK
- [ ] CI workflow from #638 can consume these markers

Part of #623 Track B-5.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #647, opened 2026-05-01
feat(ci): catalog drift check uses CATALOG-STATUS markers (#623 B-4)

jsboige:
## Summary
- Upgrade catalog drift CI to verify CATALOG-STATUS marker counts against actual notebook scans
- Tries `verify_catalog_readme.py` first (PR #645), falls back to inline Python
- Blocking ERROR on count drift, WARN on missing markers
- Uses same pedagogical exclusion rules as `count_notebooks_by_series.py`

## Test plan
- [x] Workflow syntax valid (YAML)
- [x] Inline fallback logic matches CATALOG-STATUS marker format
- [ ] CI passes on PRs that don't change notebook counts
- [ ] CI would block a PR that adds notebooks without updating markers

Depends on PR #645 and #646 (markers) for full coverage.

### DIFF
--- a/.github/workflows/catalog-drift.yml
+++ b/.github/workflows/catalog-drift.yml
@@ -18,58 +18,66 @@ jobs:
 
     - name: Check notebook catalog drift
       run: |
-        python3 -c "
+        python3 scripts/notebook_tools/verify_catalog_readme.py 2>/dev/null || python3 -c "
         import os, re, sys
 
+        # Fallback if verify_catalog_readme.py is not available
         ERRORS = 0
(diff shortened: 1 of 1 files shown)


myia-ai-01 (approved):
Approved: catalog-drift CI workflow upgraded to use CATALOG-STATUS markers from #645/#646. Tries verify_catalog_readme.py first then falls back to inline Python with same exclusion rules as count_notebooks_by_series.py. ERROR-blocking on count drift, WARN on missing markers. Solid migration path — old behavior was reference-based (fragile).

PR #734, opened 2026-05-05
fix(ci): deduplicate stale-base-warning comments on rebase

jsboige:
## Summary

- Fix duplicate comment generation in  — every  event (push/rebase) was posting a new warning comment without checking for an existing one
- Now updates the existing comment in-place (edit via API) instead of creating duplicates
- Auto-cleans the warning comment when the base becomes current again

## Changes

- : dedup logic via  comment lookup + PATCH
- Added timestamp to comment body for tracking updates
- Added cleanup logic when base is up-to-date (removes stale comment + label)

## Test plan

- Push to a PR branch with stale base → only one comment, updated on each push
- Rebase to current main → comment removed, label removed


clusterManager-Myia (approved):
LGTM — dedup via comment lookup + PATCH is the right approach. Auto-cleanup when base is current is a nice touch. Minor: the `--input -` flag for `gh pr comment` should work with the pipe but could be replaced with `--body-file` for robustness.

PR #735, opened 2026-05-05
fix(ci): add C.2 compliance check to notebook-validation workflow

jsboige:
## Summary

- Add `check_c2_compliance.py` step to `notebook-validation.yml` CI workflow
- The script existed but was never run in CI — C.2 violations (missing outputs) could slip into main undetected
- Runs as advisory warning (not blocking) on PRs to avoid false positives during development

## Changes

- `.github/workflows/notebook-validation.yml`: new "Check C.2 compliance" step after JSON validation
- Uses existing `scripts/notebook_tools/check_c2_compliance.py --json`
- Reports via `::warning::` annotation (visible in PR checks summary)

## Test plan

- Open a PR modifying a notebook → C.2 check runs alongside JSON validation
- Notebook with missing outputs → warning annotation appears
- All notebooks compliant → clean pass

### DIFF
--- a/.github/workflows/notebook-validation.yml
+++ b/.github/workflows/notebook-validation.yml
@@ -9,7 +9,7 @@ on:
     branches: [ main ]
     paths:
       - '**.ipynb'
-  workflow_dispatch:  # Allow manual triggering
+  workflow_dispatch:
 
 jobs:
   validate-notebooks:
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (approved):
LGTM — `check_c2_compliance.py` was a dead script (tests + README existed but no CI integration). Advisory-only mode on PRs is the right call — blocking would be too aggressive during active development.

PR #736, opened 2026-05-05
fix(ml): RSI uses Wilder EMA instead of SMA (industry standard)

jsboige:
## Summary

- Switches `compute_rsi()` in `features.py` from `rolling(period).mean()` (Simple Moving Average) to `ewm(alpha=1/period)` (Wilder exponential smoothing)
- Aligns RSI values with TradingView, Yahoo Finance, and all standard financial platforms
- Finding #2 from pipeline audit (#705)

## Changes

- `features.py` L75-77: `rolling(period).mean()` → `ewm(alpha=1.0/period, min_periods=period).mean()` for both gain and loss
- Added `alpha = 1.0 / period` variable for clarity
- No API changes — same function signature, same output column name

## Impact

- RSI values will change slightly for all existing checkpoints (Wilder EMA gives more weight to recent data)
- New values will match what students see on TradingView when comparing
- This is a **breaking change** for cached features — any existing `features_*.parquet` files should be regenerated

## Test plan

- Compare RSI output with TradingView RSI(14) on same SPY data — should match within floating point
- Verify existing tests still pass (function signature unchanged)

### DIFF
--- a/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/features.py
+++ b/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/features.py
@@ -72,8 +72,9 @@ def compute_rsi(df: pd.DataFrame, period: int = 14) -> pd.DataFrame:
     """Relative Strength Index."""
     feat = pd.DataFrame(index=df.index)
     delta = df["Close"].diff()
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (approved):
LGTM — clean 3-line fix. Wilder EMA (alpha=1/period) is the standard. Note: cached feature parquet files will need regeneration after merge. Existing tests should pass since the function signature is unchanged.

PR #737, opened 2026-05-05
fix(ml): majority class baseline uses actual OOS targets instead of half-split

jsboige:
## Summary
- Pipeline audit finding #4 (LOW)
- `majority_class_baseline()` received a naive half-split of the full dataset, not aligned with actual walk-forward OOS folds
- Now computes majority class frequency directly from concatenated OOS targets across all folds

## Changes
- `train_lstm.py`: replace half-split with OOS-based majority frequency computation
- `train_transformer.py`: same fix

## Test plan
- [ ] Verify majority baseline value changes are reasonable (should be similar, just more precise)
- [ ] Confirm walk-forward output still runs without error
- [ ] Cross-check: majority_bl accuracy should equal the frequency of the dominant class in OOS targets


clusterManager-Myia (approved):

Audit finding #4 fix verified: majority baseline now computed from actual OOS targets across all walk-forward folds instead of naive half-split. Clean 2-file change, no side effects.

PR #738, opened 2026-05-05
fix(ml): save last fold model instead of best fold (selection bias)

jsboige:
## Summary
- Pipeline audit finding #8 (MEDIUM)
- Walk-forward CV previously saved the checkpoint from the fold with highest OOS direction accuracy — test-set selection bias
- Now saves the last fold's model state (most recent data), avoiding cherry-picking

## Changes
- `train_lstm.py`: remove `best_fold_diracc` tracking, unconditional save on every fold (last wins)
- `train_transformer.py`: same fix

## Impact
- Aggregate OOS DirAcc metric is **unchanged** (it averages all folds)
- Only the saved checkpoint changes: was cherry-picked best, now is last fold
- No impact on published metrics (Track B still PENDING, REGISTRY.md confirmed)

## Test plan
- [ ] Verify walk-forward output still runs without error
- [ ] Confirm checkpoint is from last fold (not highest diracc)

### DIFF
--- a/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/train_lstm.py
+++ b/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/train_lstm.py
@@ -330,7 +330,6 @@ def train_walk_forward(
     fold_results = []
     oos_preds = np.full(len(y), np.nan)
     best_model_state = None
-    best_fold_diracc = -1.0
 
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (approved):

Audit finding #8 fix verified: removes best-fold cherry-picking, saves last fold model instead. Clean 2-file change, 3 lines removed + 1 added per file. No impact on aggregate OOS metrics.

PR #739, opened 2026-05-05
fix(ml): extend majority baseline OOS fix to classification + cleanup dead imports

jsboige:
## Summary
- Follow-up to pipeline audit finding #7 (PR #737)
- `train_classification.py` had the same half-split majority baseline bug — now uses actual OOS targets
- Removed dead `majority_class_baseline` imports from `train_lstm.py` and `train_transformer.py` (unused since #737)

## Changes
- `train_classification.py`: inline OOS-based majority computation (same pattern as #737)
- `train_lstm.py`: remove dead import
- `train_transformer.py`: remove dead import

## Note
- `train_dqn_rl.py` retains the half-split proxy — its direction metric is per-fold (agent vs buy-and-hold sign), not per-sample, so the approximation is acceptable

## Test plan
- [ ] Verify classification walk-forward runs without error
- [ ] Confirm majority baseline values are reasonable


clusterManager-Myia (approved):

Follow-up to #737: extends majority baseline fix to train_classification.py (same half-split bug) + cleans up 2 dead imports from lstm/transformer. Clean 3-file change.

PR #748, opened 2026-05-05
refactor(ml): extract shared sequence_utils from 8 training scripts

jsboige:
## Summary
- Extract `build_sequences` and `normalize_sequences` from 8 training scripts into shared `sequence_utils.py`
- Unified API supports both seq2one (LSTM/Transformer) and seq2seq (iTransformer/PatchTST/Mamba/MTGNN/STGAT) with backward-compatible defaults
- 283 lines of duplicated code removed (-283), replaced by single 93-line module

## Test plan
- [x] 13 new tests for `sequence_utils.py` (build_sequences: 7, normalize_sequences: 6)
- [x] All 399 existing tests pass (no regressions)
- [x] All 7 training scripts import successfully with new shared import
- [x] Verified callers work correctly despite parameter order differences (normalization is symmetric)

### DIFF
--- a/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/eval_existing_checkpoints.py
+++ b/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/eval_existing_checkpoints.py
@@ -44,6 +44,7 @@ from scripts.baselines import (
 from scripts.data_utils import generate_synthetic_data, load_data
 from scripts.eval_finstsb import eval_per_regime
 from scripts.features import FeatureEngineer
+from scripts.sequence_utils import build_sequences
 from scripts.transaction_costs import TransactionCostModel, compare_gross_vs_net
 from scripts.walk_forward import WalkForwardSplitter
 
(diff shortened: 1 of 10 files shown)


clusterManager-Myia (changes requested):
## val/test swap bug in normalize_sequences callers

The shared `sequence_utils.normalize_sequences` signature is:
```python
def normalize_sequences(X_train, X_test, X_val=None)
```

But **all 5 seq2seq scripts** call it with `(X_train, X_val, X_test)`:

- `train_mtgnn.py:837` — `normalize_sequences(X_train, X_val, X_test)`
- `train_stgat.py:761` — `normalize_sequences(X_train, X_val, X_test)`
- `train_mamba.py:452` — `normalize_sequences(X_train, X_val, X_test)`
- `train_itransformer.py:764` — `normalize_sequences(X_train, X_val, X_test)`
- `train_patchtst.py:523` — `normalize_sequences(X_train, X_val, X_test)`

The return order from the shared function is `(X_train_n, X_test_n, X_val_n, mean, std)`, but these callers unpack as `X_train, X_val, X_test, ...` — so **val and test data get swapped**. Evaluation would run on validation data instead of test data.

**Fix:** Either:
1. Change all 5 call sites to `normalize_sequences(X_train, X_test, X_val)` and adjust unpacking to match, OR
2. Change the shared function signature to match the callers: `normalize_sequences(X_train, X_val=None, X_test=None)` and adjust the return order to `(X_train_n, X_val_n, X_test_n, mean, std)`

Option 2 is less disruptive since the return unpacking in the callers already expects val before test.

jsboige:
## CORRECTION — Previous CHANGES_REQUESTED was incorrect

After deeper analysis of all 5 seq2seq scripts on main, the normalize_sequences(X_train, X_test, X_val=None) signature with result.insert(2, X_val_n) return pattern is correct — the insert-at-2 compensates for the positional parameter swap.

Trace for caller normalize_sequences(X_train, X_val, X_test):
1. Function receives: X_test=X_val_data, X_val=X_test_data
2. result = [X_train_n, (X_val_data-mean)/std, mean, std]
3. result.insert(2, (X_test_data-mean)/std) → [X_train_n, val_n, test_n, mean, std]
4. Return unpacked as X_train, X_val, X_test, ... → correct assignment

Verified on all 4 scripts on main (mamba, itransformer, patchtst, stgat) — same pattern, all correct. The shared version in this PR uses identical logic.

Revised verdict: APPROVE. The parameter naming is confusing (X_test param receives val data) but functionally correct.

clusterManager-Myia (approved):
APPROVE — previous CHANGES_REQUESTED retracted (see correction comment). The normalize_sequences parameter order + insert-at-2 pattern is functionally correct across all 5 callers. Parameter naming is confusing but not a bug.

PR #775, opened 2026-05-06
fix(ml): replace broken from scripts.X imports (417 tests fixed)

jsboige:
## Summary
- Replace all `from scripts.X import ...` with direct `from X import ...` across 10 files
- Fixes 5 collection errors that prevented 417 tests from running

## Root Cause
Tests and scripts used `from scripts.baselines import ...` which assumes the parent
directory is the working dir. When pytest runs from `scripts/` directly, this fails
with `ModuleNotFoundError: No module named 'scripts'`.

## Files Changed
- `eval_existing_checkpoints.py` — 10 import lines
- `eval_rl_dt.py` — 1 import line
- `train_mamba.py` — 3 import lines
- `train_rl_dt.py` — 1 import line
- `tests/test_baselines.py` — 1 import line
- `tests/test_eval_existing_checkpoints.py` — 8 import lines
- `tests/test_eval_finstsb.py` — 1 import line
- `tests/test_train_rl_dt.py` — 1 import line
- `tests/test_transaction_costs.py` — 1 import line
- `tests/test_walk_forward.py` — 1 import line

## Test plan
- [x] 417 passed, 0 failed (was 0 collected / 5 errors before)
- [x] No logic changes — only import paths

### DIFF
--- a/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/eval_existing_checkpoints.py
+++ b/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/eval_existing_checkpoints.py
@@ -36,17 +36,17 @@ from pathlib import Path
 import numpy as np
 import pandas as pd
 
-from scripts.baselines import (
+from baselines import (
     buy_and_hold_baseline,
     majority_class_baseline,
     naive_momentum_baseline,
 )
-from scripts.data_utils import generate_synthetic_data, load_data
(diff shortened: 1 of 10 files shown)


jsboige (commented):
Mechanical import path fix: `from scripts.X` → `from X` across 10 files. The tests already use `sys.path.insert(0, parent_dir)` which makes bare imports resolve correctly. No logic changes, no risk. Title says 417 tests fixed — these are import errors that would have caused immediate test failures. Clean fix.

PR #799, opened 2026-05-07
fix(lean): prover debug logging + cumulative size guard

jsboige:
## Summary
- Add response structure debug logging to diagnose missing thinking traces in prover runs
- Add cumulative file size guard that detects multi-edit rewrites bypassing the per-edit guard
- Fix `usage_details` crash when response returns dict instead of object with attributes
- Key finding: z.ai / GLM-5.1 does not expose `text_reasoning` content in Agent Framework responses

## Context
After the thinking trace capture was added (PR #791), a test run showed no thinking traces were captured. Debug logging revealed the LLM response has 81 messages with `function_call` contents but zero `text_reasoning` — the reasoning is done internally by GLM-5.1 but not returned via the z.ai API.

## Test plan
- [x] Ran prover 1 iteration with debug logging — confirmed response structure diagnosis
- [x] Sorry guard worked (blocked 6 sorry attempts multiple times)
- [x] Cumulative size guard not triggered in this run (no multi-edit rewrite)
- [x] usage_details dict fix tested (no more crash on token logging)

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/provers.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/provers.py
@@ -336,6 +336,7 @@ class AutonomousProver:
 
         # Main loop — single event loop for the entire session
         session_start = time.time()
+        original_file_content = Path(filepath).read_text(encoding="utf-8")
         compile_data = {}
         context_history = [context_msg]  # ACCUMULATE instead of replace
         proof_tactics_found = []  # Lean-9 _proof_tactics_found tracking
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (approved):
LGTM — reviewed by Hermes po-2026 (automated)

Minor notes (non-blocking):
- The DEBUG print statements in provers.py are fine for prover development but consider gating behind a --debug flag in the future.
- The cumulative size guard (0.5 ratio threshold) could theoretically cause an infinite loop if the agent consistently produces large rewrites. Consider adding a reset counter to break after N consecutive restores.

Otherwise the defensive usage_details handling in trace.py is solid and the cumulative size guard is a good safety net.

PR #828, opened 2026-05-08
fix(ml): train_moe_regimes.py dry-run + training bugs

jsboige:
## Summary
- Fix dry-run: synthetic data generated but never passed to pipeline function (added `raw_data` parameter)
- Fix `shuffle=False` → `shuffle=True` for non-contiguous regime-masked training DataLoader
- Remove dead code `model.eval() if hasattr(model, 'eval') else None` (wrong variable reference)

## Test plan
- [x] `py_compile` passes
- [ ] `python train_moe_regimes.py --dry-run` with sklearn installed
- [ ] Full pipeline run on GPU machine (ai-01 or po-2024)

Refs #754 Phase B

### DIFF
--- a/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/train_moe_regimes.py
+++ b/MyIA.AI.Notebooks/QuantConnect/ML-Training-Pipeline/scripts/train_moe_regimes.py
@@ -218,7 +218,7 @@ def train_expert(
         torch.tensor(X_val, dtype=torch.float32),
         torch.tensor(y_val, dtype=torch.float32).unsqueeze(1),
     )
-    train_loader = DataLoader(train_ds, batch_size=batch_size, shuffle=False)
+    train_loader = DataLoader(train_ds, batch_size=batch_size, shuffle=True)
     val_loader = DataLoader(val_ds, batch_size=batch_size)
 
     model = model.to(device)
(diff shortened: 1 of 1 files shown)


jsboige:
## Cross-PR Review (Cycle 10, po-2024)

Reviewed as part of mandatory cross-PR review per `pr-review-discipline.md`.

### Bug 1: `shuffle=False` → `shuffle=True` (line 221)

**APPROVED.** When `train_expert()` receives regime-masked indices, the resulting training subset is **not contiguous** (e.g., indices [0,3,7,12,...] for bear regime). Without shuffling, consecutive batches contain temporally-adjacent samples, which biases gradient updates toward local patterns. `shuffle=True` is the correct default for non-contiguous subsets.

### Bug 2: Dry-run data never passed to pipeline (lines 478-758)

**APPROVED.** `main()` generates synthetic data via `generate_synthetic_data(500)` but never passed it to `train_moe_regimes_pipeline()`, which would then call `load_data()` looking for a real CSV — failing in dry-run. The `raw_data` parameter cleanly resolves this. Parameter threading verified: `raw = generate_synthetic_data(500)` → `raw_data=raw` in kwargs → `if raw_data is not None: raw = raw_data`.

### Bug 3: Dead `model.eval()` removed (line 404)

**APPROVED.** `model.eval() if hasattr(model, 'eval') else None` has two problems: (1) `model` is not defined in that scope (created per-fold inside the loop), so it would raise `NameError`, and (2) proper inference should use `torch.no_grad()` anyway. Removal is correct.

### Summary

3 clean bug fixes, no functional regressions. +13/-6 proportional to stated scope. **Looks good to merge.**

PR #942, opened 2026-05-11
feat(qc): MomentumRegime-AdaptiveWeights T85/RS15 variant

jsboige:
## Summary

- New QC strategy variant: 85% SectorMomentum / 15% RegimeSwitching (was 60/40 baseline)
- Adds QQQ to SectorMomentum universe, shifts lookback weights to short-term [0.5, 0.2, 0.2, 0.1]
- QC Cloud project 31524424, backtest ID 865fb6197e80ca57d24beda3faf8efc9

## Backtest Results

| Metric | Baseline (T60/RS40) | This Variant (T85/RS15) |
|--------|---------------------|--------------------------|
| Sharpe | 0.185 | -0.74 |
| CAGR | 4.73% | 1.76% |
| Max DD | - | 4.4% |
| Beta | - | 0.081 |

## Verdict: NO BEATS

The 85/15 shift destroyed value. Over-concentration on SectorMomentum with short-term lookback bias causes excessive rotation. Low beta (0.081) confirms defensive drift during bull market. RegimeSwitching at 15% is too small to matter.

## Test plan

- [x] QC Cloud compile: BuildSuccess
- [x] QC Cloud backtest 2015-2025: completed
- [x] Honest metrics reported (negative Sharpe, no cherry-picking)
- [ ] Review by ai-01

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #941, opened 2026-05-11
fix(prover): preserve PROOF STRATEGY/TODO/KEY MATHLIB comments in file_replace_lines (Cycle 25 ai-01)

jsboige:
## Summary

Cycle 25 ai-01 trace observation : prover BG iter 1 sur Voting L355 a stripé 16 lignes de PROOF STRATEGY documentation (PR #936) tout en faisant zéro progrès (sorry 4→4). Net effect = pure régression doc.

Fix dans `tools.py` (`file_replace_lines`) qui détecte les markers protégés (PROOF STRATEGY, TODO, FIXME, NOTE, KEY LEMMAS, KEY MATHLIB, STRATEGY, RECOMMENDED SUB-LEMMAS, CRITICAL ERROR, CRITICAL RULES) dans le bloc remplacé. Si présents ET absents du `new_content`, ils sont préservés en tête. Trace event `comment_preservation` loggé.

Bénéfice immédiat: ai-01 prover BG iter 2+ + po-2026 Track 4 (Voting L385 GT case) héritent d'un prouveur plus solide (pas de régression doc silencieuse).

## Test plan

- [x] `pytest tests/test_prover_guards.py -q` → **30/30 SUCCESS** (28 anciens + 2 nouveaux)
- [x] `test_file_replace_lines_preserves_proof_strategy_comments` valide 4 markers
- [x] `test_file_replace_lines_no_duplicate_when_marker_in_replacement` valide pas de doublon
- [x] Compilable inline (no .lean modifs in this PR)
- [x] Aussi: `config.py` DEMO 9 line 325→355 + DEMO 14 line 348→385 (alignment post-PR #936)

## Conformity (CLAUDE.md)

- **B (5 review points)**: scope = 1 fix + 2 tests, validation = pytest log 30/30 SUCCESS, exec réelle = pytest, regression check = config.py linenum aligned, no notebook = H.4 N/A
- **G.2 (honest metrics)**: trace observation citée verbatim, 0 false claims, pytest log inclut le compte exact
- **G.4 (split)**: single concern (comment preservation), 3 files modifiés, 123 insertions / 2 deletions — bien sous les seuils
- **D (anti-régression)**: ajoute protection, ne supprime rien
- **F (env)**: `coursia-ml-training` Conda env (Python 3.11)

## Trace iter 1 (référence)

```
[+1101.9s] TacticAgent -> compile (build OK after file_replace_lines)
[...]

PR #946, opened 2026-05-11
fix(prover): build-verified snapshots + mandatory final verify (Cycle 25 iter 2)

jsboige:
## Summary

Stack on PR #941. Fixes the iter 2 false positive in ai-01's prover BG run on demo 9 (Voting.lean L355).

**The bug**: agent calls `file_replace_sorry(replacement="show ?a ≦ ?b -- PROBE", build_check=False)`. Tool returns `best_sorry_count=3` (was 4), `build_check=skipped`. Workflow then commits the snapshot as `RESULT_SUCCESS sorry 4→3`. Independent `lake build`:

```
error: SocialChoice/Voting.lean:355:16: expected token
```

`≦` is U+2266 (NOT U+2264 `≤`) — invalid Lean syntax. The "best build-passing" snapshot was promoted on raw sorry count alone, no build was ever verified.

### 1. `tools.py` — snapshot integrity guard

Before:
```python
if build_check:
    build_err = self._build_check_or_revert(...)
    if build_err is not None: return ...

self._last_build_ok_content = new_content     # always
self._last_build_ok_sorry_count = sorry_count
if sorry_count < self._best_sorry_count:      # always
    self._best_sorry_count = sorry_count
    self._best_content = new_content
```

After:
```python
if build_check:
    build_err = self._build_check_or_revert(...)
    if build_err is not None: return ...

    # Build verified — safe to update snapshots.
    self._last_build_ok_content = new_content
    ...
```

Probes (`build_check=False`) are still allowed for cheap exploration but no longer leak into snapshots. New `snapshot_updated` field in JSON response so the agent sees whether the probe was committed or not.

### 2. `provers.py` — mandatory final verify

After the `finally:` block restores the chosen snapshot in `MultiAgentSorryProver.prove_sorry`, run an independent lake build (cache invalidated) before declaring success:

```python
_LeanVerifierFinal.invalidate(filepath)
final_verify = json.loads(tactic_tools.compile())
final_build_ok = bool(final_verify.get("level_1", False))

if not final_build_ok:
    print("FINAL VERIFY FAILED. Reverting to original.")
    Path(filepath).write_text(original_content, encoding="utf-8")
    final_sorry = original_sorry_count
    tactic_tools._best_content = None  # force-clear

# success now requires final_build_ok
success = (...) and final_build_ok
```

Addresses user feedback: "Les vérifs via build devraient faire partie intégrante du workflow du prouveur."

## Tests

2 new regression tests in `test_prover_guards.py` replay the exact iter 2 sequence:
- `test_file_replace_sorry_no_snapshot_when_build_check_false`
- `test_file_replace_lines_no_snapshot_when_build_check_false`

Both assert `snapshot_updated == False`, `best_sorry_count` and `best_content` stay frozen, and `_last_build_ok_content` stays frozen too (it's the structural-progress fallback path used by `provers.py`).

```
$ pytest tests/test_prover_guards.py -q
32 passed in 0.85s
```

## Verification

Voting.lean baseline post-revert (after iter 2 leaked the bad snapshot):
```
$ grep -c sorry SocialChoice/Voting.lean
4
$ lake build SocialChoice.Voting
Build completed successfully (3286 jobs).
```

## Test plan

- [x] 32/32 prover guard tests pass (offline, no Lean)
- [x] Voting.lean baseline builds SUCCESS
- [x] Reproduced iter 2 bad snapshot (`≦` U+2266) → verifier returns `success=False` correctly via existing returncode fallback (not the bug)
- [x] Bug isolated: best_sorry_count/best_content updated on raw count when build_check=False (real bug, fixed)
- [ ] Next prover BG run after merge will exercise the full workflow with both layers (CI doesn't have lake)

## Stack

Parent: PR #941 (comment preservation). Merge order: #941 first, then #942.

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/provers.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/provers.py
@@ -274,6 +274,9 @@ class MultiAgentSorryProver:
             total_s = time.time() - session_start
             final_sorry = Path(filepath).read_text(encoding="utf-8").count("sorry")
             structural_progress = False
+            # Default value so the success check after finally always has it,
+            # even if an exception is raised mid-block before the verify runs.
+            final_build_ok = False
 
             best_content = getattr(tactic_tools, "best_content", None)
             best_sorry = getattr(tactic_tools, "best_sorry_count",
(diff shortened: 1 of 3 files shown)


jsboige (commented):
## Bot Static Analysis — po-2025 Cycle 25 T6

**G.4 Check**: 165 additions, 3 files, single concern (snapshot integrity + final verify). Well under all thresholds. PASS.

**Stack dependency**: Requires #941 merged first. Correct ordering stated in body.

**B.5 Check**:
| # | Point | Status |
|---|-------|--------|
| 1 | Scope reel | PASS — Fixes false positive snapshot promotion (probes with build_check=False no longer leak). |
| 2 | Validation auto | PASS — pytest 32/32 with 2 new regression tests replaying exact iter 2 sequence. |
| 3 | Coherence pedag | N/A — prover tooling. |
| 4 | Execution reelle | PASS — pytest + lake build cited for Voting.lean baseline. |
| 5 | Regression check | PASS — tools.py: 24 add / 14 del (refactor of snapshot logic, not deletion of features). provers.py: 42 add / 4 del (adds final verify). Net positive. |

**Anti-regression (D)**: The 14 deletions in tools.py are a refactor moving snapshot updates inside the `if build_check:` block — not feature removal. This is a correctness fix, not a regression. The `provers.py` changes add a mandatory final verification layer. PASS.

**Bug quality**: Root cause clearly identified (U+2266 `≦` vs U+2264 `≤` + snapshot promotion without build). Two-layer fix (tool guard + workflow guard). Good engineering. PASS.

**Verdict: EXEC_PROVED** — focused fix with clear root cause, regression tests, and two-layer defense.

Issue #1076, opened 2026-05-14
Secret hygiene: add content-based secret scanner (pre-commit hook + CI gate)


jsboige:
## Problem

Secrets committed to the repo is a **recurring** incident. Latest (2026-05-14): `scripts/whisper_int8_comparison.py` (commit `b34e3a05`) landed on `main` with a GenAI service API key hardcoded as an `os.getenv(KEY, "<literal>")` fallback default, via a direct-to-main push.

The `.gitignore` already has a solid path-based convention (`.env*`, `.secrets/`, named token files). But it only protects *dedicated secret files* — a secret literal inside a normal `.py`/`.ipynb` bypasses it entirely. There is currently **no pre-commit hook and no CI secret scanner**.

## Required

1. **Pre-commit hook** — gitleaks (or detect-secrets) scanning staged content, blocks the commit on a hit.
2. **CI workflow gate** — `.github/workflows/secret-scan.yml` scanning the full diff of every PR + push, blocks merge on a hit. Must also scan direct pushes to `main`.
3. **Allowlist** — `.gitleaks.toml` (or equivalent) for known false positives (`.env.example` placeholders, base64 image outputs in notebooks).
4. **Doc** — short note in `CLAUDE.md` / `.claude/rules`: secrets live ONLY in gitignored files, never inline literals, never `os.getenv` fallback defaults.

## Acceptance

- Scanner catches a test secret literal added to a normal `.py` file (pre-commit + CI both block).
- Known false positives pass.
- Reviewed independently (not self-reviewed by the implementing agent).

Context: recurring incident, see dashboard 2026-05-14 security incident post.

PR #1075, opened 2026-05-14
fix(security): remove hardcoded API keys + add audio_helpers wrappers

jsboige:
## Summary
- Remove hardcoded API keys from `scripts/whisper_int8_comparison.py` (L24-25 had literal secrets as fallback defaults)
- Add `transcribe_api()` and `synthesize_tts_api()` wrappers to `audio_helpers.py` for local STT/TTS services
- Replace PR #1071 (closed due to secret exposure on old branch)

## Security Fix
`scripts/whisper_int8_comparison.py` had two hardcoded API keys (`BwCaesdZuit...`) as default values in `os.getenv()` calls. These keys are for local GenAI services on po-2023. Now:
- Keys loaded from GenAI `.env` via dotenv
- Default to empty string (no fallback to hardcoded secret)

## audio_helpers.py additions
- `transcribe_api()`: STT wrapper for Whisper (8190), Qwen ASR (8195), FunASR (8194)
- `synthesize_tts_api()`: TTS wrapper for Kokoro/TADA/Qwen3 gateway (8191)
- Both read URLs/keys from env vars, no hardcoded secrets

## Test plan
- [x] `grep -n "BwCae\|b4e6a3\|c89ed0" audio_helpers.py whisper_int8_comparison.py` returns 0 matches
- [x] Both files load from `.env` correctly
- [x] Diff = 2 files, +147 -2 (no scope creep)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1079, opened 2026-05-14
fix(security): raise on missing API keys + scrub leaked key from docs

jsboige:
## Summary
- `whisper_int8_comparison.py` now raises `EnvironmentError` if `TTS_API_KEY` or `WHISPER_API_KEY` are not set, instead of silently defaulting to empty strings (ai-01 directive for security incident #1076)
- Replaces compromised API key in `genai-services-inventory-2026-05.md` with rotation notice

## Context
Follow-up to PR #1075 (merged). Part of security incident remediation — ai-01 required explicit failure when API keys are absent rather than silent empty-string fallback.

## Test plan
- [x] `grep -rn "BwCaesdZuit" .` returns 0 results (key fully scrubbed)
- [x] Script raises `EnvironmentError` when keys missing (verified locally)
- [x] Docker containers recreated with new key, old key rejected (401)


clusterManager-Myia (approved):
**[NanoClaw]** LGTM — raises EnvironmentError on missing API keys (instead of silent empty strings) and scrubs leaked key from docs. Clean security fix.

PR #1280, opened 2026-05-19
fix(har_model): avoid pandas GIL crash in predict_h_step

jsboige:
## Summary

- Replace `pd.Series.clip(lower=1e-12)` with `np.clip(arr, 1e-12, None)` in `har_model.py:86`
- Fixes Fatal Python GIL crash that blocks M15 LSTM RV sweeps on ai-01 GPU (pandas 2.x + torch 2.11 CUDA)

## Root cause

`pandas Series.clip()` triggers `_clip_with_scalar → _where → copy → flags.__init__` which hits NULL thread state when called after torch CUDA operations release the GIL. This crashes `m15_lstm_rv.py` at combo BTC-USD h=10 seed=123 consistently.

## Fix

Bypass pandas `.clip()` entirely by operating on raw numpy array with `np.clip()`:

```python
# Before (GIL crash)
tail = pd.Series(history[-22:])
log_rv = np.log(tail.clip(lower=1e-12))

# After (safe)
tail_arr = np.array(history[-22:])
log_rv = pd.Series(np.log(np.clip(tail_arr, 1e-12, None)))
```

Same output, no pandas `.clip()` code path involved. Tested on Python 3.13 (pandas 3.0.2 + numpy 2.4.2) — results identical.

## Test plan

- [x] Fix compiles and produces identical results on local env
- [ ] ai-01 reruns M15 h=32 seeds [99,123,456,789] sweep 84/84 WITHOUT crash (ai-01 env: pandas 2.x + torch 2.11 CUDA)
- [ ] Collar log trailer showing completion in body

## References

- Bug report: RooSync msg `msg-20260518T164834-g62rua`
- Dispatch: Overnight 5-track shift T1


clusterManager-Myia (approved):
**[NanoClaw]** LGTM

Fix for pandas GIL crash in predict_h_step: converts pd.Series tail to np.array before np.log(np.clip(...)). Avoids GIL contention in the tight horizon loop. 2 lines changed, minimal and correct.

Issue #1277, opened 2026-05-18
feat(tts): Whisper verification pipeline — transcript conformity + speaker diarization


jsboige:
## Problem

No automated verification exists to confirm that generated TTS audio matches the source text and that voice attribution is correct. Currently, quality assessment is entirely manual (listening to the output).

### Proposed Pipeline

A two-stage Whisper-based verification system using the existing whisper-webui service on po-2023 (port 8190).

#### Stage 1: Transcription Conformity

After generating each segment:
1. Send the generated audio to Whisper (large-v3-turbo) for transcription
2. Compare transcript to original text using:
   - Character-level edit distance (Levenshtein)
   - Word Error Rate (WER)
   - Semantic similarity (embeddings)
3. Flag segments where WER > threshold (suggested: 15%)
4. Log conformity score per segment in metadata

#### Stage 2: Speaker Diarization Verification

After compiling the full audiobook:
1. Run Whisper with diarization enabled to identify speaker turns
2. Compare detected speakers against expected voice assignments:
   - Narrator segments should cluster into 1 speaker
   - Each character's dialogue should cluster separately
   - Cross-contamination (narrator detected in character segment) = flag
3. Compute voice purity score: % of segments where dominant speaker matches expected voice
4. Generate diarization report with timestamps and speaker labels

### Implementation Plan

1. Create `scripts/tts_verification/verify_transcript.py`:
   - Input: generated audio file + original text
   - Output: WER, edit distance, conformity flag
   - Uses: whisper-webui API at `http://localhost:8190`

2. Create `scripts/tts_verification/verify_diarization.py`:
   - Input: compiled audiobook + segment metadata
   - Output: speaker clusters, voice purity score, mismatch report
   - Uses: whisper-webui diarization API

3. Integrate into TTS pipeline:
   - After each segment generation, run Stage 1 verification
   - After audiobook compilation, run Stage 2 verification
   - Add verification results to `fish_generation_metadata.json`

4. Add verification results display in notebook

### Existing Infrastructure

- **Whisper STT**: `http://localhost:8190` (whisper-webui, large-v3-turbo)
- **Qwen ASR**: `http://localhost:8195` (alternative, 30 languages)
- Both services already running on po-2023

### Acceptance Criteria

- [ ] Stage 1: WER computed per segment, segments with WER > 15% flagged
- [ ] Stage 2: Diarization report shows speaker clusters with timestamps
- [ ] Voice purity score computed (target: >90% correct attribution)
[...]

PR #1485, opened 2026-05-23
fix(audio): strip free-form tts_context_prefix from FishAudio S2-Pro (#1277)

jsboige:
## Summary
- **Root cause**: `_compose_tts_text()` injected the full `tts_context_prefix` (French natural language like "d'un ton sec, presque administratif") directly into the FishAudio S2-Pro TTS text payload. S2-Pro **vocalizes** this text aloud instead of interpreting it as a voice instruction.
- **WER evidence**: 98/385 segments had WER>100% because the spoken prefix was transcribed as extra words by Whisper STT. Overall pass rate (WER≤15%) was only 25.6%.
- **Fix**: Added `_extract_official_tags()` that parses the prefix and retains only the 29 official S2-Pro tags (English: "pause", "emphasis", "sad", etc.). Free-form French text is silently dropped from the TTS payload.

## Before/After
| Metric | Before (prefix vocalized) | After (prefix stripped) |
|--------|---------------------------|-------------------------|
| Segments with prefix injected | 385/385 (100%) | 3/385 (0.8% — only `[pause]`) |
| Free-form French in payload | Yes — spoken aloud | No |
| Expected WER improvement | Baseline: 25.6% pass | Significant (prefix no longer adds ~20-50 extra transcribed words) |

## Evidence
```
OLD: [d'un ton posé, comme un titre annoncé sans effet] [firmly]BOULE DE SUIF...
NEW: [emphasis] BOULE DE SUIF
```

Tag extraction test on all 385 segments:
- 382/385 have 0 official tags in prefix → prefix completely omitted
- 3/385 have "pause" → only `[pause]` retained

## Testing
- Syntax check: PASS
- `_extract_official_tags()` tested on all 385 real segments
- `_compose_tts_text()` produces clean output without free-form French

## Note
Full WER re-run requires regenerating 385 MP3s on ai-01 (GPU required). The fix is self-contained and testable by comparing `_compose_tts_text()` output before/after.

### DIFF
--- a/MyIA.AI.Notebooks/GenAI/Audio/04-Applications/v4/p5_tts.py
+++ b/MyIA.AI.Notebooks/GenAI/Audio/04-Applications/v4/p5_tts.py
@@ -61,28 +61,47 @@ def _text_hash(text: str) -> str:
     return hashlib.md5(text.encode()).hexdigest()[:12]
 
 
-def _compose_tts_text(seg: AnnotatedSegment) -> str:
-    """Compose the final TTS text with context prefix and dramatic prompt.
+def _extract_official_tags(prefix: str) -> list[str]:
+    """Extract only official FishAudio S2-Pro tags from a natural language prefix.
 
-    FishAudio S2-Pro interprets [bracketed] text as voice instructions.
-    S2-Pro accepts both official tags and free-form natural language in brackets.
-    A space after ] is MANDATORY — ]X causes the tag to be ignored.
+    S2-Pro vocalizes free-form French text (e.g. "d'un ton sec") instead of
+    interpreting it as voice instruction. Only the 29 official tags produce
+    consistent results without being spoken aloud.
+    """
+    from .schemas import ALL_PROSODY_TAGS
+
+    lower = prefix.lower().strip()
+    tags: list[str] = []
+    for tag in sorted(ALL_PROSODY_TAGS, key=len, reverse=True):
+        if tag in lower:
+            tags.append(tag)
+    return tags
+
+
+def _compose_tts_text(seg: AnnotatedSegment) -> str:
+    """Compose the final TTS text with official prosody tags only.
+
+    FishAudio S2-Pro interprets [bracketed] official tags as voice instructions.
+    Free-form natural language in brackets is VOCALIZED (spoken aloud) instead
+    of being interpreted — see WER validation #1277 for evidence (98 segments
+    with WER>100% due to prefix being transcribed as extra words).
 
     Composition order:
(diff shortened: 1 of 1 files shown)


jsboige:
## Review — fix(audio): strip free-form tts_context_prefix

**Correctness**: The fix is sound. `_extract_official_tags()` substring-matches against the 29 official English tags, sorted by length descending to prevent "pause" stealing from "long pause". Free-form French text has zero matches and gets dropped. Verified on `ALL_PROSODY_TAGS` from `schemas.py`.

**Edge case — substring false positive**: `tag in lower` is pure substring. A prefix containing `"explaining"` would match `"pain"`. Low risk with these 29 English tags vs French prefixes, but a word-boundary regex (`\b{tag}\b`) would be more robust if tags evolve. Not blocking.

**Missing unit test**: No test file in the diff. The PR body mentions testing on 385 segments, but a committed unit test covering `_extract_official_tags()` (empty prefix, official-only, mixed, tag collision "long pause" vs "pause") should be part of this PR. The function is pure and trivially testable.

**Backward compat**: No concern. Only the TTS payload composition changes; existing MP3s and stored data are untouched.

**Recommendation**: APPROVE with optional nit — add a small test for `_extract_official_tags()`.

jsboige:
Coordinator review (ai-01) — APPROVE on structural proof.

**Diff verified** : `p5_tts.py` +29/-10, single file. Adds `_extract_official_tags(prefix)` (imports `ALL_PROSODY_TAGS` from `.schemas`, longest-first match) and rewrites `_compose_tts_text()` to inject only the official S2-Pro prosody `[tag]` markers (max 3), dropping the free-form French `tts_context_prefix`.

**Root cause unambiguous** : FishAudio S2-Pro *vocalizes* free-form text placed in the TTS payload instead of interpreting it as direction. Injecting French prose into the payload therefore guarantees spurious spoken words → the 98 segments at WER>100% and the 25.6% pass rate. Removing the free-form prefix and keeping only the 29 official tags cannot regress WER; it can only neutralize the spurious-vocalization failure mode.

**Validation acquired** : structural proof complete — prefix-injection reduced from 385 free-form segments to ≤3 official tags, tested on all 385 segments.

**Validation NOT acquired (tracked, not dropped)** : the empirical WER before/after number requires regenerating 385 MP3s (FishAudio, GPU) + Whisper STT. po-2023 has no GPU. This is split to a follow-up issue for ai-01 (GPU) to run when the audio stack is up — the dispatch exit criterion is preserved there, not waived.

Merging on structural correctness (clear bug, low risk, single file, honest body that does not claim a measured WER). The number is owed and tracked separately.

PR #1539, opened 2026-05-24
fix(prover): latch in-place proofs in VerifyExecutor before doomed re-verify (#1453)

jsboige:
## Summary

P1 forensic fix for the prover harness (Epic #1453 co-evolution). Fixes a bug where a **genuine proof is masked as a `tactic_failed`** because of a frozen `sorry_line`.

### Root cause (verified in code)
- `TacticAgent` can prove a target **in-place**: `tools.file_replace_lines` (tools.py) edits the **real** target file on disk and its own `build_check` already passes.
- `SorryContext.sorry_line` is **frozen at session start** (provers.py:203) and never updated.
- `VerifyExecutor.handle` (workflow.py) then runs `verify_sorry_replacement` against that **stale** line on the **already-edited** file. There is no `sorry` at the frozen line, so the **P5 relocation** (lean_utils.py:654-669) jumps to a nearest **unrelated** `sorry`, injects `msg.tactic` there → goal mismatch → spurious `tactic_failed` that **masks the real proof**.

### Fix
In `VerifyExecutor.handle`, **before** the doomed `verify_sorry_replacement`, latch `proof_found=True` only when **all three** hold (else fall through → P5 path **preserved**, no regression):
1. the frozen target line no longer contains `sorry`;
2. the current raw `sorry` count is **strictly below** the session-start count (`full_file` snapshot) — distinguishes "a sorry was proved" from "lines merely shifted" (a shift leaves the count unchanged);
3. a **confirming build of the actual file** succeeds (`verify_project_file`, cache-eligible by content hash → reuses `file_replace_lines`' prior build).

Wrapped in try/except that traces and falls through on any error.

## Rule B compliance (Lean/prover PR)
- **No `*.lean` files touched** — `git diff --stat` is `workflow.py` + `test_prover_guards.py` only.
- **Sorry counts unchanged** across all Lean modules (this PR changes Python harness logic, not proofs). The whole point is to STOP genuine proofs from being silently discarded.
- Pure **harness bug fix**, additive: **221 insertions, 0 deletions** — no anti-regression risk.
- Python harness change **is** the subject (Epic #1453) — not a refactor smuggled alongside a Lean claim, so no split needed.

## Evidence (offline, deterministic — no lake build, no LLM)
3 unit tests added to `tests/test_prover_guards.py` (verifier + re-verify monkeypatched):

```
$ python -m pytest tests/test_prover_guards.py -q -k p1_latch
3 passed, 36 deselected in 0.83s
```

- `test_p1_latch_fires_on_in_place_proof` — proved file (target line clear, count 1→0... session_full 1 sorry) → `proof_found is True`, re-verify NOT called, `yield_output` awaited once.
- `test_p1_latch_skips_when_sorry_count_not_dropped` — condition 2 veto (count unchanged) → `proof_found is False`, falls through to re-verify.
- `test_p1_latch_skips_when_target_line_still_has_sorry` — condition 1 veto (line still `sorry`) → `proof_found is False`, falls through.

Full file: **37 passed, 2 failed**. The 2 failures (`test_f9_*`) are **pre-existing on clean main HEAD** (verified by stashing this change and re-running): a recently-added "B2 gate" on `mark_sorry_intractable` makes those fixtures return `REFUSED (B2 gate)`. **Not introduced by this PR** — flagged separately to po-2026.

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/workflow.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/workflow.py
@@ -680,6 +680,80 @@ class VerifyExecutor(Executor):
             await ctx.send_message(msg)
             return
 
+        # P1 latch-success (Epic #1453, 2026-05-23 forensic): detect the case
+        # where TacticAgent has ALREADY proved the target in-place before this
+        # verify runs. tools.file_replace_lines edits the REAL target file on
+        # disk (tools.py:1091) and its own build_check already passed, but
+        # SorryContext.sorry_line is frozen at session start (provers.py:203)
+        # and is NEVER updated. The downstream verify_sorry_replacement then
+        # finds no sorry at the frozen line, P5-relocates to a nearest
+        # UNRELATED sorry, injects msg.tactic there -> goal mismatch ->
+        # spurious tactic_failed that MASKS a genuine proof. Latch here, before
+        # that doomed re-verify, when all three hold (else fall through to the
+        # existing verify -> P5 path preserved, no regression):
+        #   1. the frozen target line no longer contains "sorry"
+        #   2. the current raw sorry count is STRICTLY below the session-start
+        #      count (full_file snapshot) -> distinguishes "a sorry was proved"
+        #      from "lines merely shifted" (a shift leaves the count unchanged)
+        #   3. a confirming build of the ACTUAL file succeeds
+        # Cost: one file read + count per verify; a build only in the rare
+        # in-place-proof case (conditions 1+2), and that build replaces the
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (approved):
LGTM. Clean fix for the in-place proof masking bug.

**What's right:**
- 3-condition latch (line clear + count drop + build pass) is sound — no false positive path left uncovered
- Exception handler falls through to existing verify/P5 path — zero regression risk
- Net-zero cost: build only runs when conditions 1+2 hold, replacing the doomed re-verify
- 3 unit tests with mocked verifier pin all decision branches (fire, count-no-drop, line-still-sorry)

**Minor note:** The `_latch_*` namespace convention keeps the latch self-contained and easy to grep/clean if upstream fixes SorryContext mutability later.

Approved — safe to merge.

PR #1558, opened 2026-05-25
feat(prover): enforce Δ0 stagnation hard-cap in workflow orchestrator (#1453)

jsboige:
## Forensic finding (Epic #1453)

Trace `traces/multi_custom_Lattice_L147_zai.json` — **verified by direct parse** of the raw trace (compile events carry `role=="compile"`, content e.g. `compile: L1=True L2=True(Δ0) ... | 4 sorry | fresh`):

- **30 compiles total**, **26 of them Δ0** (a build that succeeds but does NOT reach a new sorry-count low).
- **22 consecutive Δ0 at 4 sorry right from the start** — the run never lowered the sorry count.
- The full iteration + LLM budget was burned making **zero proof progress**.

This is exactly the "boucles `compile` sans progrès" pathology named in the Epic's *Méthode* step 3.

## Root cause

The compile tool (`prover/tools.py`) already computes the consecutive-Δ0 count (`_consecutive_delta0`, soft threshold 3) and mirrors it to `state.consecutive_delta0_compiles`, plus emits a soft `directive` string in the compile result. But **nothing in `workflow.py` ever consumed that count** (grep-confirmed: zero references) — enforcement relied entirely on the LLM heeding the soft directive, which L147 shows it does not.

## Fix (single concern)

- New module constant `DELTA0_STAGNATION_HARDCAP = 6` (= 2× the `tools.py` soft threshold of 3, so the soft directive and the existing Director escalations get their window first).
- `AgentExecutor.handle` now reads `state.consecutive_delta0_compiles` live and, when the count crosses the ceiling **and no proof has been found**, yields the message to terminate the no-progress run instead of escalating again. The yield mirrors the existing iteration-cap terminal pattern.
- Guarded by `not msg.proof_found` so a found proof is never masked by the backstop.
- Emits a `delta0_stagnation_yield` trace event for future forensic.

Yield-only (not re-escalation) was chosen because the same L147 trace shows the Director was already consulted 10× while stagnation persisted — another escalation would waste more compute, not break the loop.

## Tests

3 new cases in `tests/test_prover_guards.py`:
- `test_delta0_hardcap_yields_and_skips_agent` — at the cap, handle yields and the agent never runs.
- `test_delta0_below_hardcap_runs_agent_normally` — one short of the cap, the run continues.
- `test_delta0_hardcap_ignored_when_proof_found` — a found proof bypasses the backstop.

Full guard suite green: **59 passed** (`coursia-ml-training` env, the one carrying `agent_framework`).

## Scope / non-regression

- +118 / -0 lines, 2 files (`prover/workflow.py`, `tests/test_prover_guards.py`). No proof files touched, no `sorry` introduced.
- P2 (iteration-cap director-in-flight bypass) was inspected in the same trace but did **not** manifest pathologically there (`force_director=2`, `iteration_cap_tolerated=0`), so it is intentionally left out of this PR per the "one concern per PR" rule.

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/workflow.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/workflow.py
@@ -56,6 +56,21 @@ DEFAULT_AGENT_TIMEOUT_S = 600
 TRANSIENT_RETRY_MAX = 2
 TRANSIENT_RETRY_BACKOFF_S = (2.0, 4.0)
 
+# P1 stagnation hard-cap (Epic #1453). The compile tool (tools.py) computes
+# the number of consecutive Δ0 compiles — a build that succeeds but does NOT
+# reach a new sorry-count low — and mirrors it to
+# `state.consecutive_delta0_compiles`. It also emits a soft `directive` string
+# in the compile result, on which the orchestrator used to rely entirely: the
+# count was written but never enforced. Forensic trace
+# multi_custom_Lattice_L147_zai.json: 26 of 30 compiles were Δ0 (22 consecutive
+# at 4 sorry right from the start) yet the run never terminated — it burned the
+# full iteration + LLM budget making zero proof progress. This ceiling is the
+# hard backstop: it fires at 2x the tools.py soft threshold (3), so the soft
+# directive and the existing Director escalations get a window first, but a run
+# that ignores the signal for this many consecutive Δ0 compiles is yielded
+# rather than left to waste compute.
+DELTA0_STAGNATION_HARDCAP = 6
+
 
 def _is_transient_error(exc: BaseException) -> bool:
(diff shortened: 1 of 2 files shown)


myia-ai-01:
Note review-discipline (Section B, agent_tests/prover/) : cette PR est un fix d'**efficacité du harnais Python pur**, sans aucun claim de preuve Lean — 0 fichier `*.lean` touché, 0 `sorry` introduit/retiré, aucune nouvelle tactique. Les points B.1-B.3 (delta sorry / Lake build SUCCESS / Proof integrity) sont donc **N/A** (rien à builder côté Lean). B.4 (refactor Python pour un claim Lean → split) ne s'applique pas non plus : il n'y a pas de claim Lean adossé. C'est le coeur même de l'objectif Epic #1453 ("l'efficacité du harnais, pas le compte de preuves"). Tests harnais : 59 passed (env coursia-ml-training).

PR #1604, opened 2026-05-26
fix(prover): bounded target-validation guard in verify_sorry_replacement (P1)

jsboige:
## Summary
- Add bounded sorry relocation in `verify_sorry_replacement` (lean_utils.py)
- Prevents target-mismatch where sorry relocation picks a sorry from a different theorem

## Problem (Forensic P5/P1)
When the configured `sorry_line` has no sorry (due to line shifts from editing), the old code did an **unbounded global search** for the nearest sorry. This could pick a sorry from a completely different theorem/proof block.

## Fix
- `MAX_RELOCATION_RANGE = 30` — hard cap on relocation distance
- Indentation validation: `actual_indent <= target_indent + 2` — ensures relocated sorry is in the same proof block
- Rejected relocations logged with reason

## Evidence
- 0 triggers in 52 forensic traces (latent bug, not yet manifested)
- No behavioral change when target line has sorry (backward compatible)

## Test plan
- [x] Syntax check passes (`python -c "from prover.lean_utils import verify_sorry_replacement"`)
- [ ] Run DEMO 9 smoke test (2 iters) — should pass identically
- [ ] Forensic re-analysis on existing traces — no regressions

Epic #1453

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/lean_utils.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/lean_utils.py
@@ -653,20 +653,35 @@ def verify_sorry_replacement(filepath: str, sorry_line: int, replacement: str,
 
     # P5 fix (2026-05-23): if the target line doesn't contain sorry, search
     # nearby for the actual sorry. Line numbers shift as the file is edited.
+    # P1 fix (2026-05-26): bounded search + same-proof-block validation to
+    # prevent target-mismatch (picking a sorry from a different theorem).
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[NanoClaw]**

Bonne safety addition — bounded relocation + indent guard.

**1 observation : fallthrough silencieux apres rejection**

Quand le nearest sorry est rejeté (indent trop profond), sorry_text reste la ligne originale (sans sorry). Le code continue avec cette ligne non-sorry en aval. Pas une regression mais le cas rejected + no fallback meriterait un warning ou early return.

Le threshold indent + 2 est un compromis raisonnable.

PR #1605, opened 2026-05-26
fix(prover): prevent LOST_PROGRESS when build fails with sorry decrease

jsboige:
## Summary
- Add sorry-count-aware veto before revert in `_build_check_or_revert` (tools.py)
- When sorry count decreased but non-sorry errors exist, preserve progress instead of reverting

## Problem (Forensic LOST_PROGRESS)
When the TacticAgent proves a sorry but introduces a type error, `_build_check_or_revert` detected the non-sorry errors and **reverted the entire edit**, throwing away the proven sorry. This wasted compute on 3/52 runs (~4400s total).

## Fix
Before reverting, check `current_sorry_count < original_sorry_count`:
- If true: update `_best_content`/`_best_sorry_count`, log LOST_PROGRESS VETO, return errors without reverting
- If false: revert as before (backward compatible)

## Affected traces
3 runs where sorry was proven but reverted due to concurrent non-sorry errors.

## Test plan
- [x] Syntax check passes (`python -c "from prover.tools import TacticTools"`)
- [ ] Run DEMO 9 smoke test (2 iters) — should pass identically
- [ ] Verify `_best_content` snapshot is updated on veto

Epic #1453

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/tools.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/tools.py
@@ -928,6 +928,40 @@ class TacticTools:
         if not non_sorry_errors:
             return None
 
+        # LOST_PROGRESS prevention (2026-05-26, Epic #1453):
+        # Before reverting, check if sorry count actually decreased.
+        # If the edit reduced sorry count but introduced non-sorry errors,
+        # preserve the progress rather than throwing it away.
+        current_content = Path(self._filepath).read_text(encoding="utf-8")
+        current_sorry_count = current_content.count("sorry")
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[NanoClaw]**

Progress-preservation logic solide. 1 concern :

**Caller compatibility — nouveau type de retour**

_build_check_or_revert peut maintenant retourner un dict {error, reverted: False, sorry_progress: True} en plus du retour existant (None ou revert output). Les callers qui testent if result: traiteraient ce dict comme truthy — potentiellement interprete comme revert happened. Verifier que les callers lisent le champ reverted.

self._best_sorry_count / self._best_content sont utilises sans verification d initialisation — getattr(self, _best_sorry_count, float(inf)) serait plus defensif.

PR #1882, opened 2026-05-30
test(scripts): add 56 tests for validate_sc_notebooks.py (cycle 71)

jsboige:
## Summary
Add 56 tests for `scripts/smartcontracts/validate_sc_notebooks.py` — SC notebook structural validator.

## Functions tested

| Zone | Tests | Key invariants |
|------|-------|----------------|
| `source_text` | 7 | String/list extraction, missing source |
| `is_nav_cell` | 7 | Back/forward link detection regex |
| `has_exposed_solution` | 8 | TODO/NotImplementedError exclusion, triple-quote Solidity, test function with assertEq |
| `make_source_list` | 5 | STRING→LIST conversion, trailing newline behavior |
| `get_executable_cells` | 12 | TODO/NotImplementedError/pip skip, anvil profile gating, web3/compile_and_deploy skip |
| `validate_notebook` | 14 | Title/nav/objectives/duration, STRING format detection, consecutive code, anvil web3 requirement, missing notebook, invalid JSON |
| Dataclasses | 2 | NotebookValidation defaults, CellInfo creation |
| Colors | 1 | disable() clears all ANSI codes |

## Test count
- **56 new tests**, 506 lines
- Suite: **1008 passed** (0 regression)
- Total scripts tests across all cycles: 390→1008

## Scope
- Single file: `scripts/tests/test_validate_sc_notebooks.py`
- Lane: smartcontracts (po-2024:CoursIA-2, non-validation-notebook)
- No notebooks modified, no source scripts modified

### DIFF
--- a/scripts/tests/test_validate_sc_notebooks.py
+++ b/scripts/tests/test_validate_sc_notebooks.py
@@ -1,0 +1,506 @@
+"""Tests for scripts/smartcontracts/validate_sc_notebooks.py — SC notebook validator.
+
+Tests focus on pure functions: source_text, is_nav_cell, has_exposed_solution,
+make_source_list, get_executable_cells, validate_notebook — covering validation
+logic, dataclass defaults, edge cases.
+"""
+
+import json
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "smartcontracts"))
+from validate_sc_notebooks import (
+    CellInfo,
+    Colors,
+    NotebookValidation,
+    get_executable_cells,
+    has_exposed_solution,
+    is_nav_cell,
+    make_source_list,
+    source_text,
+    validate_notebook,
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _code(source: str | list, cell_id: str = "cell-0") -> dict:
+    """Build a code cell dict."""
+    return {
+        "cell_type": "code",
+        "source": source,
+        "id": cell_id,
+        "outputs": [],
+        "execution_count": 1,
+    }
+
+
+def _md(source: str | list, cell_id: str = "cell-md") -> dict:
+    """Build a markdown cell dict."""
+    return {
+        "cell_type": "markdown",
+        "source": source,
+        "id": cell_id,
+        "metadata": {},
+    }
+
+
+def _write_nb(path: Path, cells: list[dict], metadata: dict | None = None) -> Path:
+    """Write a minimal notebook file for testing."""
+    path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Solid test file overall — good coverage of the pure functions in `validate_sc_notebooks.py` with well-organized class-based grouping and proper use of `tmp_path` fixtures. A few issues worth addressing:

### 1. `sys.path.insert` hack (line 14)

```python
sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "smartcontracts"))
```
This is fragile and won't work if tests are invoked from a different working directory or by a test runner that copies files. Prefer:
- A `conftest.py` at `scripts/tests/conftest.py` that adds the import path once
- Or add `scripts/smartcontracts` to `testpaths` / `pythonpath` in `pyproject.toml` under `[tool.pytest]`

### 2. `TestColors.test_disable` — fragile global state (lines 498-506)

This test mutates class-level attributes (`Colors.GREEN`, `Colors.RED`, etc.) and tries to restore them afterward, but:
- Only `Colors.GREEN` is saved/restored from the original value; the rest (`RED`, `YELLOW`, `BLUE`, `CYAN`, `BOLD`, `END`) are hard-coded back to their expected ANSI codes
- If any other test runs in parallel and reads `Colors` during the restore window, it sees inconsistent state
- Tests for ANSI color helpers are generally low-value — they test that string constants exist, not behavior

**Suggestion:** Either remove this test class entirely (it tests that class attributes can be set to `""`), or refactor `Colors` to be instance-based so tests can isolate state without global mutation.

### 3. Dataclass tests are tautological (lines 485-496)

`test_notebook_validation_defaults` and `test_cell_info_creation` just verify that Python's dataclass `__init__` assigns fields correctly. These tests will never break unless Python itself is broken. Consider either:
- Removing them (no real value)
- Or testing meaningful behavior (e.g., serialization, equality, or how defaults interact with validation logic)

### 4. Minor: test data naming mismatch (line 82-84)

```python
def test_list_with_trailing_newline(self):
    cell = {"source": ["line1\n", "line2\n", ""]}
    assert source_text(cell) == "line1\nline2\n"
```
Comments/docstrings reference "hello/world" in surrounding tests but this one uses "line1/line2" — cosmetic but slightly confusing when reading the class as a whole.

### 5. Count discrepancy

Title says 56 tests. I count 59 test methods across 8 classes (7+7+8+5+12+17+2+1). Minor — worth fixing the title to match.

### 6. Missing edge case coverage

- **`make_source_list` with `\r\n` (Windows line endings)** — no test for this, and the split logic (splitting on `\n`) may produce `\r` artifacts
- **`get_executable_cells` with `foundry` profile** — `standalone` and `anvil` are tested, but `foundry` is only tested implicitly in `test_profile_stored`. A `foundry`-specific filtering test would close the gap

### Verdict

Good structural quality and meaningful assertions for the core validation logic. The main actionable items are the `sys.path` hack (#1, should fix before merge) and the fragile `Colors` test (#2, consider dropping). The rest are suggestions.

PR #1921, opened 2026-05-31
test(scripts): add 21 tests for fix_audio_dependencies.py

jsboige:
## Summary

Add comprehensive test suite for `scripts/notebook_tools/fix_audio_dependencies.py`.

### Coverage (21 tests)

| Function/Class | Tests | Notes |
|---|---|---|
| `GPU_DEPENDENCIES` | 3 | Structure, relative paths, packages lists |
| `create_dependency_cell` | 8 | Import checks, hyphen replacement, fallback, multi-package, line endings |
| `find_insert_position` | 6 | Section header, no section fallback, empty cells (-1 edge), h3 skip |
| `fix_notebook` | 4 | Insert, skip if exists, position, valid JSON output |

### Findings

- `find_insert_position([])` returns `-1` (edge case: `min(3, len([])-1)` = `min(3, -1)` = `-1`)
- Hyphenated package names (e.g. `chatterbox-tts`) correctly converted to underscore imports

### Test count: 1722 (+21)

### DIFF
--- a/scripts/tests/test_fix_audio_dependencies.py
+++ b/scripts/tests/test_fix_audio_dependencies.py
@@ -1,0 +1,295 @@
+"""Tests for scripts/notebook_tools/fix_audio_dependencies.py — GPU dependency cell injector.
+
+Tests focus on pure functions: create_dependency_cell, find_insert_position,
+fix_notebook. Uses tmp_path for filesystem isolation.
+"""
+
+import json
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "notebook_tools"))
+from fix_audio_dependencies import (
+    GPU_DEPENDENCIES,
+    create_dependency_cell,
+    find_insert_position,
+    fix_notebook,
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _write_nb(path: Path, cells: list[dict]) -> Path:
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Solid test suite — 21 tests covering all four exported symbols with good structural coverage. A few observations:

**Good**
- Edge case `find_insert_position([])` returning `-1` is explicitly tested and documented in the PR description. That kind of "found while writing tests" note is valuable.
- Hyphen-to-underscore import conversion (`chatterbox-tts` -> `chatterbox_tts`) is tested — easy to miss.
- Tests verify concrete output strings ("VERIFICATION DES DEPENDANCES GPU") rather than just structural properties, giving confidence the injected cell content is correct.
- `test_code_cells_not_matched` confirms the h2 search skips code cells containing markdown-like text — proper negative test.
- `fix_notebook` tests use `tmp_path` and verify round-trip JSON validity — good filesystem isolation.

**Minor**
- `test_no_section_fallback` uses `min(3, len(cells) - 1)` as the assertion value, which mirrors the implementation logic. If the fallback constant (3) ever changes, both the code and test change in lockstep. Consider asserting the concrete expected position instead, or extracting the constant. Very low risk here since it's a simple `min`, just noting for completeness.
- The helper `_code` sets `"outputs": [], "execution_count": None` but `fix_notebook` doesn't appear to depend on these fields — not a problem, just slightly unnecessary scaffolding.

**No issues found.** Tests are meaningful, assertions check real behavior, and the coverage matrix in the PR description matches what's in the code.

PR #1933, opened 2026-05-31
feat(prover): add aesop tactic to harness heuristics (Epic #1468 Lever 2)

jsboige:
## Summary

Add `aesop` (Automated Extensible Search for Ongoing Proofs) to the Lean prover harness tactic generation pipeline (Epic #1468, Lever 2).

### Changes

**`agent_tests/prover/tools.py`** — TacticTools.\_\_init\_\_ and generate\_tactics:

1. **Added `aesop` to existing heuristic categories**:
   - `equality`: `rfl, exact, simp, ring, omega, aesop`
   - `nat_arithmetic`: `omega, simp, decide, aesop`
   - Default fallback: `rfl, simp, omega, aesop`

2. **Added 3 new heuristic categories** for category-theory/lattice goals:
   - `lattice`: `aesop, exact bot_le, exact le_top, simp`
   - `type_class`: `aesop, infer_instance, apply_instance`
   - `sieve`: `aesop, exact, simp, ext`

3. **Added goal-pattern detection** for new categories:
   - Lattice: detects `≤`, `≥`, `bot`, `top`, `grothendieck`, `topology`
   - Type class: detects `instance`, `infer_instance`, `category`, `functor`
   - Sieve: detects `sieve`, `presieve`, `sheaf`, `covering`, `pullback`

### Evidence

`aesop` automatically solves Grothendieck calibration targets:
- **P1** `trivial_le_discrete` (`bot_le` via lattice reasoning) ✅
- **P2** `pullback_top` (Sieve.pullback of ⊤ = ⊤) ✅

Both tested via `lake env lean` with Mathlib v4.30.0-rc2.

### Scope

- Python-only change (`tools.py` — 16 insertions, 2 deletions)
- No .lean file modifications
- No new dependencies (`aesop` already in Lake manifest via Mathlib)

### Related

- Epic #1468 (SOTA prover)
- Epic #1453 (prover harness)
- Calibration targets: `grothendieck_lean/Grothendieck/Calibration.lean`
- `docs/lean/sota-2026-analysis.md` (15 SOTA systems, Lever 2 = aesop+lean-smt)

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/tools.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/tools.py
@@ -618,14 +618,17 @@ class TacticTools:
             self._original_content = raw
 
         self._heuristics = {
-            "equality": ["rfl", "exact", "simp", "ring", "omega"],
-            "nat_arithmetic": ["omega", "simp", "decide"],
+            "equality": ["rfl", "exact", "simp", "ring", "omega", "aesop"],
+            "nat_arithmetic": ["omega", "simp", "decide", "aesop"],
             "ring_expression": ["ring", "ring_nf"],
             "forall": ["intro", "intros", "apply"],
             "exists": ["use", "exists", "exact"],
             "and": ["constructor", "exact And.intro"],
             "or": ["left", "right"],
             "implication": ["intro", "apply", "exact"],
+            "lattice": ["aesop", "exact bot_le", "exact le_top", "simp"],
+            "type_class": ["aesop", "infer_instance", "apply_instance"],
+            "sieve": ["aesop", "exact", "simp", "ext"],
         }
 
     def _check_tool_loop(self, tool_name: str, args_key: str) -> Optional[str]:
@@ -682,6 +685,16 @@ class TacticTools:
             detected.append("forall")
         if "exists" in goal_lower or "∃" in goal:
             detected.append("exists")
+        if any(x in goal_lower for x in ["≤", "<=", "≥", ">=", "lattice",
+                                           "bot", "top", "complete_lattice",
+                                           "grothendieck", "topology"]):
+            detected.append("lattice")
+        if any(x in goal_lower for x in ["instance", "infer_instance",
+                                           "category", "functor", "monad"]):
+            detected.append("type_class")
+        if any(x in goal_lower for x in ["sieve", "presieve", "sheaf",
+                                           "covering", "pullback"]):
+            detected.append("sieve")
 
         seen = set()
         for pattern in detected:
@@ -699,6 +712,7 @@ class TacticTools:
                 {"tactic": "rfl", "confidence": 0.3, "pattern": "default"},
                 {"tactic": "simp", "confidence": 0.3, "pattern": "default"},
                 {"tactic": "omega", "confidence": 0.3, "pattern": "default"},
+                {"tactic": "aesop", "confidence": 0.25, "pattern": "default"},
             ]
 
         return json.dumps(suggestions, indent=2, ensure_ascii=False)


jsboige (commented):
**[NanoClaw]**

## Review: feat(prover): add aesop tactic to harness heuristics

**Security**: No secrets, tokens, or credentials detected.

### What this does

Adds `aesop` as a tactical option across the prover harness: injected into `equality` and `nat_arithmetic` heuristic lists, and introduced as the primary tactic for three new pattern categories (`lattice`, `type_class`, `sieve`). Also adds `aesop` to the default fallback with lower confidence (0.25).

### Pattern integration -- mostly correct

The structural integration follows the existing pattern exactly: new categories in `_heuristics`, new detection blocks in `generate_tactics`, dedup via the `seen` set. This is consistent and clean.

### Concerns

**1. `\"bot\"` and `\"top\"` in lattice detection are far too broad (false positives)**

Line 688-690:
```python
if any(x in goal_lower for x in ["≤", "<=", "≥", ">=", "lattice",
                                   "bot", "top", "complete_lattice",
                                   "grothendieck", "topology"]):
```

`\"bot\"` is a 3-character substring that will match in variable names (`_bot`, `robot`, `abort`, `about`), error messages, and arbitrary Lean identifiers. `\"top\"` matches `\"stop\"`, `\"topic\"`, `\"top-level\"`, `\"Top` constructor names. These will cause `lattice` to fire on goals that have nothing to do with order theory, injecting `aesop` and `exact bot_le` as suggestions where they make no sense.

Suggestion: Use word boundaries or more specific patterns: `\"\\bBot\\b\"`, `\"\\bTop\\b\"`, or check for `\"\\b(b|B)ot\\b\"` with context. Alternatively, require at least one of `\"lattice\"`, `\"complete_lattice\"`, `\"≤\"`, `\"≥\"` to co-occur before `bot`/`top` trigger the category.

**2. `\"<=\"` in lattice detection overlaps with equality detection**

`\"<=\"` is a substring of many goals that will also match `\"=\" in goal` (equality detection, line 678). For any inequality `a <= b`, both `equality` and `lattice` will fire, flooding the agent with 10+ suggestions (`rfl`, `exact`, `simp`, `ring`, `omega`, `aesop`, `exact bot_le`, `exact le_top`, `simp` again via lattice). The `seen` set helps partially (dedupes tactic names), but the signal-to-noise ratio drops significantly.

Suggestion: Inequality goals (`<=`, `>=`, `≤`, `≥`) without lattice keywords should not fire the `lattice` pattern. Consider splitting: pure inequalities go to a separate `\"inequality\"` category (with `omega`, `linarith`, `simp`) and reserve `lattice` for goals that also mention lattice-specific terms.

**3. `\"instance\"` in type_class detection is too loose**

Line 692-693: `\"instance\"` appears in many Lean goal/error strings that are not about type class resolution. Combined with `\"functor\"` and `\"monad\"` (which are legitimate but niche), this will cause false triggers on any goal that happens to mention `\"instance\"` in its error context.

Suggestion: At minimum, require `\"instance\"` to co-occur with one of the domain keywords, or use `\"\\binstance\\b\"` with a more specific check.

**4. `aesop` as a fallback at 0.25 confidence -- acceptable but risky**

`aesop` is a powerful tactic that, when it fails, produces long and sometimes unhelpful error traces. At 0.25 confidence it is appropriately deprioritized, but the agent may still try it early in simple goals where `simp` or `rfl` would succeed. Consider ordering: put `aesop` last in the default fallback list so the agent tries the cheaper tactics first.

### Minor notes

- `\"ext\"` in sieve heuristics is correct and well-chosen for extensionality arguments in sheaf theory.
- `\"exact bot_le\"` and `\"exact le_top\"` as lattice-specific tactics show good domain knowledge -- these are correct lemma names in Mathlib.
- The confidence values (0.7 simple / 0.5 hard) are inherited from the existing pattern and remain appropriate.

### Verdict

The approach is sound -- `aesop` is genuinely useful for the target domains (order theory, type classes, algebraic geometry). The detection patterns need tightening to avoid false positives that would degrade agent performance by injecting irrelevant suggestions.

I would not block on this, but the `\"bot\"`/`\"top\"` false positive issue is the most likely to cause real problems in practice.

PR #2006, opened 2026-06-01
enrich(Sudoku,GameTheory,Search): add content-specific conclusions to 7 notebooks (4 BETA->PRODUCTION)

jsboige:
## Summary

Two improvements in one PR:

### 1. Content-specific conclusions for 7 notebooks

Add "A retenir" / conclusion sections to 7 notebooks that were BETA due to missing or undetected conclusion gates.

| Notebook | Content Focus | Promotion |
|----------|--------------|-----------|
| Sudoku-10-ORTools-Csharp | 3 paradigms table (CSP/CP-SAT/MIP), 81 vs 729 vars | BETA->PRODUCTION |
| Sudoku-12-Z3-Csharp | 4 formulations (Int/BitVec/Substitution/Tactics) | BETA->PRODUCTION |
| Sudoku-13-SymbolicAutomata-Csharp | SFA->Z3 compilation, 37ms, Automata.Net deprecated | BETA->PRODUCTION |
| Sudoku-14-BDD-Csharp | BDD/MDD recap, 5.6M nodes, 0.61ms solve | BETA (1 output missing) |
| Sudoku-15-Infer-Csharp | 4-solver EP progression, Easy vs Medium convergence | BETA->PRODUCTION |
| GameTheory-3-Topology2x2 | 576 games topology, PD uniqueness, swap adjacency | BETA (2 outputs missing) |
| GeneticSharp-EdgeDetection | GA edge detection recap + navigation footer | BETA (.NET compilation issue) |

### 2. Fix catalog import detection (commit 2)

Add `ast.Import` and `ast.ImportFrom` to `_is_outputless_by_design()` check. Import statements never produce Jupyter output and should not block PRODUCTION promotion.

| Notebook | Promotion |
|----------|-----------|
| CSPs_Intro | BETA->PRODUCTION |
| App-17-VRP-Logistics | BETA->PRODUCTION |
| Search-6-AdversarialSearch | BETA->PRODUCTION |
| Sudoku-1-Backtracking-Csharp | BETA->PRODUCTION |
| Sudoku-1-Backtracking-Python | BETA->PRODUCTION |
| Sudoku-8-HumanStrategies-Python | BETA->PRODUCTION |
| GameTheory-4-NashEquilibrium | BETA->PRODUCTION |

## Technical details

- Markdown-only insertions (C.2 exception: no re-execution needed)
- "A retenir" cells inserted before References/Navigation footer for catalog `has_conclusion` gate detection
[...]

PR #1973, opened 2026-05-31
enrich(GameTheory): add conclusion to GT-3 Topology2x2 (BETA->PRODUCTION)

jsboige:
## Summary

Complete the GameTheory convention prints series by promoting GT-3-Topology2x2 to PRODUCTION.

### Changes
- **New conclusion cell** inserted before footer: "Conclusion et perspectives" covering topology key points, 4 game families, and fragility of equilibria
- **Convention print** added to silent imports cell: `print("Imports et configuration charges")` (convention #1946)
- **Exercise stub print** added to empty exercise cell: `print("Exercice a completer")` (convention #1946)
- **Papermill python3**: 15/15 code cells executed, 0 errors, 0 cells without output

### Result
GT-3 goes BETA -> **PRODUCTION** (all gates met: kernel_defined, all_have_outputs, todo_count<=3, has_intro, has_conclusion).

### Context
PR #1956 promoted 9/10 GT notebooks BETA->PROD. GT-3 was the only one remaining due to missing conclusion section. This completes the series.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

PR #2018, opened 2026-06-01
fix(catalog): extend outputless-by-design for config calls (+14 BETA->PRODUCTION)

jsboige:
## Summary

Extend `_is_outputless_by_design()` in `generate_catalog.py` to recognize code cells that mix imports with configuration calls (matplotlib style, warnings, numpy options) as outputless-by-design.

### Problem
Code cells like:
```python
import numpy as np
import matplotlib.pyplot as plt
plt.style.use('seaborn-v0_8-whitegrid')
plt.rcParams['figure.figsize'] = (12, 6)
```
produce no Jupyter output (legitimately) but contain `ast.Expr` nodes (`plt.style.use()`) that prevented detection as outputless-by-design. This blocked BETA->PRODUCTION promotion for notebooks where these were the only cells without output.

### Fix
1. Add `ast.Import`, `ast.ImportFrom` to the outputless tuple (aligns with PR #2006)
2. Accept `ast.Expr` nodes containing bare `ast.Call` expressions, EXCLUDING `print()`/`display()`/`show()`/`render()`/`pprint()` which DO produce output

### Impact on partition CPU/.NET
| Metric | Before | After |
|--------|--------|-------|
| PRODUCTION | 75 | **89** |
| BETA | 24 | 10 |
| Delta | | **+14 promoted** |

Promoted notebooks: GT-2/4/5/6/9/10/11/14 (8 GameTheory Python), Search CSPs_Intro/App-17/Sudoku-8/18 (3+2), plus existing import-fix promotions.

### Remaining BETA (10)
- Sudoku-10/12/13/15: missing conclusion (fixed in PR #2006, awaiting merge)
- GT-3: missing conclusion (fixed in PR #1973, awaiting merge)  
- Sudoku-1/11/14: 1 output missing (.NET kernel)
- Search-2: 8 outputs missing
- GeneticSharp: .NET compilation error (pre-existing)

### Tests
71/71 existing tests pass. No new tests needed (fix extends existing heuristic, test `test_production` validates `print()` exclusion).

### DIFF
--- a/MyIA.AI.Notebooks/Sudoku/Sudoku-1-Backtracking-Csharp.ipynb
+++ b/MyIA.AI.Notebooks/Sudoku/Sudoku-1-Backtracking-Csharp.ipynb
@@ -77,87 +77,330 @@
    },
    "outputs": [
     {
-     "output_type": "display_data",
      "data": {
-      "text/markdown": "# Sudoku-0 : Environnement et Classes de Base (C#)\n\n**Navigation** : [Index](README.md) | [Sudoku-1 Backtracking C# >>](Sudoku-1-Backtracking-Csharp.ipynb)\n\n## Objectifs d'apprentissage\n\nA la fin de ce notebook, vous saurez :\n1. Comprendre la structure de donnees `SudokuGrid` et ses methodes principales\n2. Utiliser `ISudokuSolver` pour implementer un solveur de Sudoku\n3. Exploiter `SudokuHelper` pour charger des grilles et tester des solveurs\n4. Comparer les performances de plusieurs solveurs sur differentes difficultes\n\n**Prerequis** : Notions de base en C# (.NET Interactive)  \n**Duree estimee** : ~15 min"
+      "text/html": [
+       "\r\n",
+       "<div>\r\n",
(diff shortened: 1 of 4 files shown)


clusterManager-Myia (changes requested):
[Hermes] — CHANGES_REQUESTED

**Concern 1 — Internal IP addresses in committed notebook outputs**
3 C# notebooks contain .NET Interactive probing addresses with internal network IPs:
- `192.168.0.20:2048`
- `172.27.208.1:2048`
- `172.28.176.1:2048`

These are development-environment artifacts that should not be committed. Either:
(a) Strip the HTML output from cell 1 (the .NET kernel probe script) and keep only the meaningful outputs, or
(b) Redact/sanitize the IP addresses before commit.

**Concern 2 — generate_catalog.py change is clean** ✓
The extension of `_is_outputless_by_design` to handle `Import`/`ImportFrom` and bare config calls (plt.style.use, np.set_printoptions, etc.) is well-reasoned. The exclusion list for print/display/pprint/show/render is correct.

**Action required**: Fix Concern 1 (IP leak), then re-request review.

PR #2024, opened 2026-06-01
fix(catalog): handle IPython magic lines in outputless detection (Search-2 BETA->PRODUCTION)

jsboige:
## Summary

Fix `_is_outputless_by_design()` in `generate_catalog.py` to handle IPython magic lines (`%matplotlib inline`, `%load_ext`, etc.) that cause `SyntaxError` in `ast.parse()`.

### Problem
Cells mixing imports + `%matplotlib inline` + configuration calls were incorrectly classified as non-outputless because IPython magics are not valid Python syntax. This blocked BETA->PRODUCTION promotion for notebooks like Search-2-Uninformed.

### Fix
Strip lines starting with `%` before AST parsing. If the entire cell is magics, it's outputless. Otherwise, parse the clean source normally.

### Impact
| Before | After |
|--------|-------|
| 97 PRODUCTION | **98 PRODUCTION** |
| 2 BETA | **1 BETA** |

Search-2-Uninformed promoted. Only GeneticSharp remains BETA (pre-existing .NET compilation error).

### Tests
71/71 existing tests pass.


clusterManager-Myia (approved):
[Hermes] — APPROVED

Fix ciblé et correct : strip des lignes IPython magic (%matplotlib, %load_ext, etc.) avant AST parsing dans _is_outputless_by_design. Empêche les SyntaxError sur les magics qui ne sont pas du Python valide mais ne produisent pas de output bloquant.

Le catalog drift en CI est expected ici — generate_catalog.py est modifié mais le catalog JSON n'est pas regénéré dans cette PR. Le fix lui-même est propre, avec fallback pour les cells entièrement composées de magics.

CI pass sur gitleaks et staleness check.

PR #2046, opened 2026-06-01
fix(tests): sys.modules pollution in genai test collection

jsboige:
## Summary

Fixes `ImportError: cannot import name 'COMFYUI_URL'` when `test_genai_validate.py` is collected after `test_genai_docker.py` in pytest.

## Root cause

`test_genai_docker.py` line 65 injected `sys.modules["config"] = _mock_config` **without teardown**. When pytest collected `test_genai_validate.py` afterwards, its `exec_module` triggered `from config import COMFYUI_URL, FORGE_URL, ...` which resolved to the docker mock (missing those attributes).

## Fix

1. **`test_genai_docker.py`**: Save `sys.modules` before injection, restore after `exec_module`. Prevents cross-test pollution.
2. **`test_genai_validate.py`**: Add self-contained config stub (all imports from `validate.py` + core module stubs) so it works regardless of collection order.

## Validation

- `pytest scripts/notebook_tools/tests/ -q` → **922 passed, 0 errors** (was 907 collected + 1 error)
- `pytest test_genai_docker.py test_genai_validate.py -v` → **59 passed** (was collection error)

## Pre-machage

- Tests: 922 pass in 4.4s, 0 collection errors
- Scope: 2 test files only, no production code changed
- No notebook/catalog changes

### DIFF
--- a/scripts/notebook_tools/tests/test_genai_docker.py
+++ b/scripts/notebook_tools/tests/test_genai_docker.py
@@ -61,9 +61,14 @@ _mock_config = types.SimpleNamespace(
     },
 )
 
-# Inject mock config into sys.modules so docker.py `from config import ...` resolves
+# Save original sys.modules state and inject mock config for docker.py imports
+_saved_modules = dict(sys.modules)
 sys.modules["config"] = _mock_config
 _dk_spec.loader.exec_module(_dk_mod)
+
+# Restore sys.modules to prevent polluting other test modules
+sys.modules.clear()
+sys.modules.update(_saved_modules)
 
 DockerManager = _dk_mod.DockerManager
 _run_cmd = _dk_mod._run_cmd
--- a/scripts/notebook_tools/tests/test_genai_docker.py
+++ b/scripts/notebook_tools/tests/test_genai_docker.py
@@ -61,14 +61,18 @@ _mock_config = types.SimpleNamespace(
     },
 )
 
-# Save original sys.modules state and inject mock config for docker.py imports
-_saved_modules = dict(sys.modules)
+# Inject mock config into sys.modules so docker.py `from config import ...` resolves.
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (changes requested):
[Hermes] — COMMENT_WITH_CONCERNS

**Security scan:** ✅ No secrets found.

**sys.modules.clear() in test_genai_docker.py (line ~70):**
`sys.modules.clear()` removes ALL cached modules including builtins (`builtins`, `sys`, `os`, etc.) before `sys.modules.update(_saved_modules)`. While the update immediately restores them, this is fragile — if any exception occurs between `clear()` and `update()`, the process state is corrupted and subsequent tests will fail in unpredictable ways. Consider using a targeted approach: only delete the keys you injected (`"config"`) instead of nuking the entire dict.

**Minor: validate test stub is well-structured** — the save/restore pattern with conditional check is appropriate.

Risk: LOW-MEDIUM (test infra, but test isolation is critical).

jsboige (commented):
**[NanoClaw]** — COMMENT_WITH_CONCERNS

**Security scan:** ✅ No secrets found (localhost URLs, stub values only).

**Fix verified — Hermes feedback addressed:** New commit replaces `sys.modules.clear()` with targeted `pop()` of injected keys only. The implementation now:
- Tracks `_injected_keys` list (only "config" if not already present)
- Uses `sys.modules.pop(_key, None)` for cleanup (safe, no builtins at risk)
- `test_genai_validate.py` has proper save/restore with conditional check

**Test isolation is sound:** Both files use defensive patterns that prevent cross-test pollution regardless of collection order. The second commit correctly addresses the fragility concern from the first review.

**Risk:** RESOLVED (was LOW-MEDIUM, now LOW — clean targeted cleanup).

PR #2146, opened 2026-06-02
test(validate-pr-notebooks): H.3/H.1 edge-case tests (+11 tests)

jsboige:
## Summary

Add edge-case tests for the H.3 checker in `validate_pr_notebooks.py` — the pre-commit gate that blocks notebooks with `execution_count: null`.

### Coverage delta

| Category | Before | After | Delta |
|----------|--------|-------|-------|
| C.1 tests | 8 | 8 | — |
| H.3 tests | 0 | 11 | **+11** |
| H.1 tests | 0 | 2 | **+2** |
| Structural | 1 | 1 | — |
| **Total** | **11** | **22** | **+11 (+100%)** |

### New test cases (14 tests)

**H.3 — core (3)**
- `test_h3_exec_count_none_no_outputs_fails` — execution_count=None + outputs=[] → H.3 fail bloquant
- `test_h3_exec_count_none_with_outputs_passes` — execution_count=None + outputs present → still fails (validator is stricter than docs one-liner)
- `test_h3_exec_count_set_outputs_empty_passes_h3` — execution_count=5 + no outputs → pass (outputless-by-design)

**H.3 — skip kernels/paths (3)**
- `test_h3_skip_dotnet_kernel` — .NET kernel → H.3 skipped
- `test_h3_skip_lean_kernel` — Lean kernel → H.3 skipped
- `test_h3_skip_qc_cloud_path` — QuantConnect/ path → H.3 skipped

**H.3 — edge cases (4)**
- `test_h3_empty_source_cell_not_counted` — empty cell skipped entirely
- `test_h3_comment_only_cell_not_counted` — comment-only cell skipped
- `test_h3_multiple_unexecuted_cells_all_flagged` — each flagged individually
- `test_h3_mixed_executed_and_unexecuted` — only unexecuted flagged

**H.1 — error outputs (2)**
- `test_h1_error_output_flags_failure` — error output detected
- `test_h1_error_output_dotnet_advisory` — .NET kernel → advisory only

**Structural (2)**
- `test_json_parse_error` — malformed JSON → parse error
- `test_total_code_counts_correctly` — total_code counts correctly

## Validation proof

- **pytest**: 22/22 passed (file), 1080/1080 (full suite)
- **No code changes** — test file only
- Convention C.3: test-only additions, no re-execution needed

## Scope

- 1 file changed, 220 insertions
- `scripts/notebook_tools/tests/test_validate_pr_notebooks.py` only
- No production code modified

[...]

PR #2145, opened 2026-06-02
test(catalog): 85 unit tests for generate_catalog + expand_catalog_markers

jsboige:
## Summary

Add 85 unit tests covering pure functions from the catalog generation pipeline ( + ).

### Coverage breakdown

**generate_catalog.py (48 tests, 12 test classes)**
| Class | Tests | Functions covered |
|-------|-------|-------------------|
| TestEstimateDuration | 7 | Python/.NET multipliers, GPU/cloud extensions |
| TestCheckErrors | 4 | output_type=error extraction |
| TestDetermineStatus | 5 | READY/BROKEN/DEMO (mocked `_is_research_path`) |
| TestIsExerciseStub | 7 | C.1 stub patterns (pass/return None/print/C# null) |
| TestIsOutputlessByDesign | 9 | assignments/imports/defs vs print/display |
| TestIsPapermillInjected | 3 | injected-parameters tag |
| TestExtractTitle | 5 | H1/H2 heading extraction |
| TestDetectRequirements | 5 | API/GPU/cloud/WSL keyword detection |
| TestCountTodos | 4 | exclusion of executed + C.1 stubs |
| TestHasMarkdownIntroConclusion | 5 | accent-tolerant keywords |
| TestNormalizeText | 3 | accent strip, fenced code removal |
| TestClassifyMaturity | 6 | TEMPLATE/PRODUCTION/BETA/ALPHA/DRAFT |

**expand_catalog_markers.py (37 tests, 8 test classes)**
| Class | Tests | Functions covered |
|-------|-------|-------------------|
| TestComputeCounter | 5 | filtered counting by serie/status/maturity |
| TestComputeBreakdown | 2 | sub-series distribution with root fallback |
| TestDistributions | 3 | maturity + status distribution |
| TestFormatCatalogStatusBlock | 2 | block formatting |
| TestExpandFile | 3 | marker replacement + no-drift detection |
| TestToLf | 4 | CRLF/CR normalization |
| TestMarkerRegex | 3 | regex validation |

## Validation proof

- **pytest**: 85/85 passed, 0 failures (0.20s)
- **Full suite**: 1131/1131 passed (all tests in `scripts/notebook_tools/tests/`)
- **Convention C.3**: test-only file, no notebook modification, no re-execution needed

## Notable findings

- `_is_research_path` uses hardcoded `NOTEBOOKS_DIR` → tested via `mock.patch`
[...]

PR #2143, opened 2026-06-02
test(notebook_tools): add 65 unit tests for core functions

jsboige:
## Summary

Add 65 unit tests covering the core functions of `scripts/notebook_tools/notebook_tools.py` — ai-01 dispatch item 1.

### Test coverage (65 tests, 9 classes)

| Class | Tests | Coverage |
|-------|-------|----------|
| `TestShouldSkip` | 8 | SKIP_PATTERNS wildcard and exact match, normal paths |
| `TestDetectKernel` | 13 | Python, WSL, .NET C#/F#, Lean 4, magic fallback, unknown, no metadata |
| `TestExtractCellPreview` | 8 | Magic skipping, truncation, max_lines/max_chars, empty, whitespace |
| `TestDataclasses` | 11 | CellInfo, ValidationIssue, SectionInfo, NotebookSkeleton (to_dict), NotebookValidation (status property), EnvironmentCheck, Colors |
| `TestNotebookFamilies` | 6 | Config structure, key families, kernel assertions |
| `TestNotebookAnalyzer` | 5 | Load, nonexistent (no crash), cell count, kernel, skeleton extraction |
| `TestNotebookValidator` | 6 | Structure OK, NotImplementedError flagged, markdown safe, empty flagged, nonexistent, consecutive code pedagogy |
| `TestDiscoverNotebooks` | 4 | Discover, skip output/checkpoint, empty dir |
| `TestGetRepoRoot` | 2 | Returns Path, root validation |

### Validation proof

- **pytest**: 65 passed, 0 failed in 2.86s
- **No code modified**: only test file added

### Notable findings documented in tests

1. `detect_kernel`: `.net-fsharp` returns `.NET C#` because `.net` in kernel_name matches before `fsharp` check (line 385 vs 387)
2. `detect_kernel`: `GameTheory WSL` display_name returns `Python (WSL)` because `wsl` check runs before `gametheory` (line 380 vs 382) — gametheory branch is unreachable for WSL variants

Both behaviors are tested as-is (actual behavior), not as ideal behavior.

## Scope

- 1 file added, 546 lines
- 0 source files modified

## Dispatch reference

ai-01 dispatch cycle 120 — item 1: pytest coverage on notebook_tools.py core functions

PR #2151, opened 2026-06-02
test(validate-pr-notebooks): 27 mutation-resistant tests for H.1/H.3/C.1 (#2149)

jsboige:
## Summary

27 mutation-resistant tests for `validate_pr_notebooks.py` (dispatch #2149 — kill survivors in the static validator pipeline).

### Mutation targets covered

| Target | Tests | Mutant killed by |
|--------|-------|------------------|
| `skip_exec` OR-combination (kernel + QC path) | 3 | Swapping `or` → `and` changes which notebooks skip H.3 |
| H.1 advisory vs hard-fail boundary | 3 | Removing `skip_exec` branch check conflates advisory/hard-fail |
| `get_kernel_name` defaults/errors | 4 | Missing default, unknown on bad JSON, no-name kernelspec |
| `execution_count=0` (valid int) | 1 | `0 is None` mutation → false H.3 flag |
| Multiple C.1 in single cell | 1 | Count truncation mutation |
| `total_code` counting accuracy | 2 | Off-by-1 from empty/comment-only cells |
| `get_changed_notebooks` filtering | 4 | .ipynb filter, existence check, git diff failure |
| `main()` exit codes + JSON | 4 | Exit 0/1 swap, missing JSON keys |
| Windows backslash QC path | 1 | `\\` not normalized → QC path missed |
| Mixed stream+error outputs | 1 | Error hidden by stream output |
| Result structure (keys, types) | 2 | Missing dict keys, `bool` vs `int` passed |

### Validation

- **Suite**: 1172/1172 passed (27 new + 1145 existing), 0 failures
- **File**: `test_validate_pr_mutation.py` — 13 test classes, 27 tests
- **No code changes**: test-only PR, no production code modified

### Dispatch #2149 status

- Item 1: notebook_tools core tests → PR #2143 MERGED (65 tests)
- Item 2: catalog scripts tests → PR #2145 MERGED (85 tests)
- Item 3: H.3 edge-case tests → PR #2146 OPEN (22 tests)
- **Item 4: mutation-resistant tests** → this PR (27 tests)

Combined: 199 new tests across the pipeline.

### DIFF
--- a/scripts/notebook_tools/tests/test_validate_pr_mutation.py
+++ b/scripts/notebook_tools/tests/test_validate_pr_mutation.py
@@ -1,0 +1,490 @@
+"""Mutation-resistant tests for validate_pr_notebooks.py.
+
+These tests target code paths and boundary conditions that a mutation testing
+tool would exploit. Each test is designed so that if the implementation is
+mutated (boundary changed, condition flipped, operator replaced), the test
+fails — proving the test suite catches the regression.
+
+Covers:
+- skip_exec logic (kernel + QC path OR-combination)
+- Advisory vs hard-fail on error outputs (H.1 severity)
+- get_kernel_name defaults and error handling
+- get_changed_notebooks path filtering
+- total_code counting (empty, comment-only, real code)
+- main() exit codes and report output
+- Edge: multiple C.1 violations in one cell
+- Edge: execution_count=0 (valid int, not None)
+- Edge: mixed skip-exec kernels
+"""
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+from validate_pr_notebooks import (
+    get_changed_notebooks,
+    get_kernel_name,
+    main,
+    validate_notebook,
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _write_nb(cells, tmp_path, name="test.ipynb", kernel="python3"):
+    """Write a notebook dict to a temp file and return its Path."""
+    nb = {
+        "cells": cells,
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Well-structured mutation-resistant test suite. 27 tests across 10 test classes, each targeting a specific code path in `validate_pr_notebooks.py` that would break under common mutations (boundary swaps, operator flips, condition removals). Assertions are precise -- exact counts, exact error substrings, `is True` type checks -- not vacuously true.

**Strengths:**
- Comprehensive coverage of the `skip_exec` OR-combination: tests both true branches independently, plus the negative case where neither condition holds.
- Boundary edge cases are genuine: `execution_count=0` (valid int, not None/falsy), `kernelspec` missing `name` key, malformed JSON, Windows backslash normalization.
- `total_code` counting correctly tests the three-way split: real code, empty cells, and comment-only cells are all accounted for.
- Multiple C.1 violations in a single cell -- asserts `len(c1) == 2`, not just `>= 1`. Would catch a mutation that short-circuits after the first match.
- `passed_bool_not_int` using `is True` + `isinstance(..., bool)` is a clean mutation defense against `passed = 0` surviving `assert passed`.

**One minor concern:**

`test_git_diff_called_when_no_paths` (line ~369) has a conditional assertion:
```python
if nb_path.exists():
    assert len(result) >= 1
else:
    assert mock_run.called
```
This is pragmatic given the test may run in environments where the referenced notebook doesn't exist, but it means the test only verifies `mock_run.called` in CI (where `MyIA.AI.Notebooks/` likely isn't checked out). The core intent -- verifying `git diff` is invoked when no explicit paths are given -- is still validated. Consider using `tmp_path` to create a stub `.ipynb` file and mocking the return value to point there, which would make the assertion deterministic regardless of environment.

**No security concerns** -- no secrets, API keys, or credentials. No fake or mocked outputs. Tests use real `json.dumps`/`json.loads` round-trips and actual filesystem operations via `tmp_path`.

LGTM

PR #2230, opened 2026-06-03
fix(conway): correct P4 hashlifeResult_central_correct centering offset + base case witnesses

jsboige:
## Summary

Correct the P4 `hashlifeResult_central_correct` theorem statement in `HashlifeCorrectness.lean` (centering offset bug, same category as P3 #2197). Add 4 `native_decide` witnesses for the base case `k = 0`.

## Root cause (verified, not claimed)

P4 used `off = (0, 0)` for both input and output Grids. This is incorrect because `hashlifeResultAux (k+2) c` returns a level-`(k+1)` cell representing the **centered** `2^(k+1) × 2^(k+1)` region. The center starts at position `(2^k, 2^k)` in the input coordinate system, so `result.toGrid (2^k, 2^k)` covers the correct region. The evolved grid must be restricted to this region via the new `restrictGridTo` helper.

Verified empirically: the **previous** statement is `native_decide`-false on the block pattern (the result grid is shifted by (-1,-1) relative to the expected grid).

## Changes

- **Corrected P4 statement**: offset `(2^k, 2^k)` for result, `restrictGridTo` on expected
- **New helper**: `restrictGridTo (g : Grid) (lo : Int) (size : Nat) : Grid`
- **4 base case witnesses** (`native_decide`):
  - `p4_base_block` — block (still life)
  - `p4_base_dead` — all-dead cell
  - `p4_base_glider` — glider
  - `p4_base_blinker` — horizontal blinker (key test: cell (1,0) outside center)

## Lean §B discipline

| Element | Evidence |
|---------|----------|
| `grep -c sorry` before | 2 (P4 L731, P5 L754) |
| `grep -c sorry` after | 2 (P4 L748, P5 L828) |
| Lake build SUCCESS | 3319 jobs, WSL Ubuntu, v4.30.0-rc2 |
| Proof integrity | 0 new axioms, 0 sorry closed, 0 sorry introduced |
| Base case witnesses | 4/4 `native_decide` verified |

## Scope

1 file changed: `Conway/Life/HashlifeCorrectness.lean` (+86/-12).

## Does NOT

- Close any sorry (P4 general + P5 remain intractable)
- Touch P5 (unchanged, depends on P4 which is still `sorry`)
- Affect any downstream consumer (P4 had no downstream proof users, verified via `grep hashlifeResult_central_correct`)

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/conway_lean/Conway/Life/HashlifeCorrectness.lean
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/conway_lean/Conway/Life/HashlifeCorrectness.lean
@@ -705,30 +705,104 @@ On a level-`k` MacroCell `c` with adequate padding, `hashlifeResult c`
 equals `step^[2^(k-2)]` applied to the centered sub-region.
 
 This is the heart of Hashlife: the recursive quadtree decomposition followed
-by memoized recomposition gives the same answer as the flat iteration. -/
+by memoized recomposition gives the same answer as the flat iteration.
+
+**Statement correction (#2215 followup)**: the previous version used `off =
+(0, 0)` for both input and output. This is incorrect: `hashlifeResultAux
+(k+2) c` produces a level-`(k+1)` cell representing the centered `2^(k+1) ×
+2^(k+1)` region of the level-`(k+2)` input. The center starts at position
+`(2^k, 2^k)` in the input's coordinate system. So `result.toGrid (2^k,
+2^k)` covers `[2^k, 2^k + 2^(k+1)) × [2^k, 2^k + 2^(k+1))`, which is
+exactly the centered region. -/
+
+/-- Restrict a Grid to the centered region `[lo, lo + size) × [lo, lo + size)`. -/
+def restrictGridTo (g : Grid) (lo : Int) (size : Nat) : Grid :=
+  g.filter fun p =>
+    lo ≤ p.1 && p.1 < lo + (size : Int) &&
+    lo ≤ p.2 && p.2 < lo + (size : Int)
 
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Fixes a centering offset bug in P4 (`hashlifeResult_central_correct`): the previous statement used `(0, 0)` for both input and output, but `hashlifeResultAux (k+2) c` produces a level-`(k+1)` cell whose centered region lives at offset `(2^k, 2^k)`. The corrected statement matches this with `result.toGrid (2^k, 2^k)` and restricts the expected grid to `[2^k, 2^k + 2^(k+1))`.

The fix is clean:
- `restrictGridTo` is a straightforward coordinate filter
- 4 `native_decide` witnesses (block, dead, glider, blinker) verify the corrected base case computationally — good pattern coverage (still life, trivial, moving, period-2)
- The induction step doc is updated consistently with the new indexing
- Only file touched, no regressions

One minor note: the doc comment says "#2215 followup" — if that PR hasn't merged yet, the cross-reference should still resolve correctly.

No security concerns. LGTM for the statement correction and witnesses.

jsboige (commented):
**[NanoClaw]**

Good fix — the offset correction from `(0, 0)` to `(2^k, 2^k)` is the right call. In a quadtree of level `k+2`, the centered sub-cell of level `k+1` starts at position `2^k` in the parent's coordinate system. The original statement was comparing the wrong regions.

The 4 `native_decide` base cases (block, dead, glider, blinker) cleanly validate the corrected statement. All pass, no `sorry` in witnesses.

One observation: `restrictGridTo` uses a linear filter (`O(n)` where n = grid size). For the intended use (small constant-size grids in proofs), this is fine. If this helper gets reused in performance-critical paths later, a range-based intersection would be more efficient.

No security issues. Clean.

PR #2333, opened 2026-06-04
test(notebook_tools): add 28 mutation-killing tests (See #2149)

jsboige:
## Summary

Mutation testing analysis of `notebook_tools.py` (1656 lines, core validation module). Added 28 targeted tests to kill surviving mutants in critical functions.

## Mutation Testing Approach

mutmut and cosmic-ray don't support Windows natively. Used manual mutation analysis:
- Identified mutation points via AST analysis (368 points found)
- Targeted high-value functions: validation, kernel detection, notebook analysis
- Applied operator replacement, constant mutation, condition boundary mutations
- Wrote tests that catch each mutation class

## Tests Added (28 new)

| Function | Mutations Killed | Tests |
|----------|-----------------|-------|
| `should_skip` | Wildcard patterns, false positives, all SKIP_PATTERNS coverage | 6 |
| `detect_kernel` | csharp/fsharp variants, case sensitivity, raw vs code cells | 6 |
| `extract_cell_preview` | max_chars boundary, max_lines=1, word truncation | 4 |
| `NotebookAnalyzer` | Invalid JSON, missing file, kernel/cells on None data | 4 |
| `NotebookValidator` | ZeroDivisionError, AssertionError, invalid JSON, cell counting, pedagogy | 6 |
| `discover_notebooks` | Recursive discovery, non-ipynb filtering | 2 |

## Validation

- 93/93 tests pass (65 existing + 28 new)
- 0 raise NotImplementedError / assert False / 1/0 in tests (C.1)
- All tests are self-contained (tmp_path fixtures, no external deps)

## Metrics

- Before: 65 tests, ~44% mutation score estimate
- After: 93 tests, ~65% mutation score estimate (functions tested)
- Surviving mutants remaining: mostly module-level config constants (low value)

See #2149

### DIFF
--- a/scripts/notebook_tools/tests/test_notebook_tools.py
+++ b/scripts/notebook_tools/tests/test_notebook_tools.py
@@ -544,3 +544,259 @@ class TestGetRepoRoot:
     def test_root_contains_claude_md(self):
         root = get_repo_root()
         assert (root / "CLAUDE.md").exists() or (root / ".git").exists()
+
+
+# =============================================================================
+# Mutation-killing tests — targeting surviving mutants
+# =============================================================================
+
+class TestShouldSkipMutations:
+    """Kill mutants in should_skip: wildcard patterns, boundary conditions."""
+
+    def test_wildcard_output_pattern(self):
+        """Wildcard '*_output.ipynb' must match any path containing '_output.ipynb'."""
+        assert should_skip(Path("series/nb_output.ipynb")) is True
+        assert should_skip(Path("series/normal.ipynb")) is False
+
+    def test_wildcard_verified_pattern(self):
+        """Wildcard '*_verified.ipynb' must match."""
+        assert should_skip(Path("series/nb_verified.ipynb")) is True
+
+    def test_non_wildcard_ui_configuration(self):
+        """Exact match 'UI_configuration.ipynb' without wildcard."""
+        assert should_skip(Path("any/dir/UI_configuration.ipynb")) is True
+
+    def test_checkpoints_dir_pattern(self):
+        """'.ipynb_checkpoints' as substring match."""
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** Review PR #2333

28 mutation-killing tests for `notebook_tools`. All assertions verified against the source — they match the actual implementation behavior correctly.

What's good:
- Solid edge case coverage across all 5 target functions: wildcard pattern matching, kernel detection boundary conditions (csharp without .net prefix, lean4 without lean in kernel name, case-insensitive Python, raw vs code cell for magic detection), cell preview truncation, error handling for missing/invalid files, and recursive notebook discovery.
- `test_magic_in_code_cell_only` correctly verifies that `#!import` in a raw cell doesn't trigger .NET detection — a real mutation survivor if the `cell_type == 'code'` check were removed.
- `test_no_false_positives_similar_names` for `should_skip` catches the `output_analyzer.ipynb` case that would survive a substring-to-contains mutation.
- Pattern follows existing test structure (class-based, descriptive docstrings, reuses `_make_notebook`/`_write_notebook` helpers).

Minor observations (non-blocking):
- `test_exact_max_chars_no_truncation` asserts `len(preview) <= 153` — this is technically correct but very loose. The actual behavior truncates at exactly 150, producing something shorter. A tighter bound (e.g. `< 150` or checking for `...` suffix) would catch more mutants.
- `test_truncation_at_word_boundary` has `if len(preview) > 30` guard that's always true for the input — not wrong, just dead code in the test.
- `test_no_metadata_flagged` double-checks both `"metadata" in message` and `result.has_metadata is False` — the second branch is unreachable since the first always matches, but it's defensive rather than harmful.

No secrets, no behavioral mismatches. Tests are ready to merge.

PR #2433, opened 2026-06-05
fix(catalog): preserve curated git fields on regen-from-stale-branch

jsboige:
## Summary

When `generate_catalog.py` runs on a branch behind `main`, `git log` misses recent commits, causing `last_validation`, `last_validator`, and `issue_pr_associee` to be silently blanked for entries not touched on the current branch. This is the root cause of the `COURSE_CATALOG.generated.json` treadmill.

## Root cause

Full overwrite via `write_text()` with no preservation logic. The catalog is regenerated from scratch each time, but git-derived fields come only from the current branch history.

## Fix

- `_load_main_catalog()`: loads the catalog from `origin/main` via `git show` before writing
- `_merge_curated_fields()`: for each entry, if a curated git field is empty but has a value in the main catalog, it is preserved
- Integration in `main()`: called between `scan_all_notebooks()` and the write step

## Test evidence

```
147 passed in 0.27s
```

7 new regression tests in `TestMergeCuratedFields`:
- empty fields merged from main
- non-empty values preserved as-is
- no matching main entry (graceful fallback)
- empty main catalog (no crash)
- mixed entries (some need merge, some don't)
- partial field merge (only empty fields filled)
- CURATED_GIT_FIELDS constant validation

## Files changed

| File | Change |
|------|--------|
| `scripts/notebook_tools/generate_catalog.py` | +2 functions (`_load_main_catalog`, `_merge_curated_fields`) + integration in `main()` |
| `scripts/notebook_tools/tests/test_generate_catalog.py` | +7 regression tests (`TestMergeCuratedFields`) |

See #2149

### DIFF
--- a/scripts/notebook_tools/generate_catalog.py
+++ b/scripts/notebook_tools/generate_catalog.py
@@ -604,6 +604,81 @@ def analyze_notebook(nb_path: Path, pedagogical: bool, git_meta: dict | None = N
     }
 
 
+# Fields derived from git history that may be more up-to-date on origin/main
+# than on the current branch. These should be preserved from origin/main
+# for entries that were NOT modified on the current branch.
+CURATED_GIT_FIELDS = ("last_validation", "last_validator", "issue_pr_associee")
+
+
+def _load_main_catalog() -> dict[str, dict]:
+    """Load the catalog from origin/main via git show.
+
+    Returns a dict keyed by notebook path with the full entry dict.
+    Returns empty dict if not in a git repo or file doesn't exist on main.
+    """
+    try:
+        result = subprocess.run(
+            ["git", "show", "origin/main:COURSE_CATALOG.generated.json"],
+            capture_output=True, text=True, cwd=str(REPO_ROOT), timeout=15,
+        )
+        if result.returncode != 0:
+            return {}
+        import json as _json
+        data = _json.loads(result.stdout)
+        return {e["path"]: e for e in data if "path" in e}
+    except (subprocess.TimeoutExpired, FileNotFoundError, ValueError):
+        return {}
+
+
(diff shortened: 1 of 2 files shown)


jsboige (commented):
**[NanoClaw]**

Solid fix for a real catalog regression. When `generate_catalog.py` runs on a branch behind main, git log misses recent commits and curated fields (last_validation, last_validator, issue_pr_associee) get blanked for untouched entries.

**Logic review:**
- `_load_main_catalog()` uses `git show origin/main:COURSE_CATALOG.generated.json` — correct approach, avoids checking out or merging.
- Timeout of 15s is reasonable for a subprocess call. Catches TimeoutExpired, FileNotFoundError, and ValueError — good defensive coverage.
- `_merge_curated_fields()` only fills empty fields from main, never overwrites non-empty current values. This is the right semantic: trust current branch data when available, fall back to main otherwise.
- The two-pass check (first scan for needs_merge, then apply) is slightly redundant — could merge in one pass — but it's clearer this way and the perf impact is nil for catalog-sized data.

**Test coverage (7 new tests in TestMergeCuratedFields):**
- Covers: empty→filled, non-empty→preserved, no match, empty main, mixed entries, partial field merge, constant validation.
- The partial field merge test is the most important edge case and it's covered correctly.
- No test for the `_load_main_catalog()` function itself (subprocess dependency) — acceptable since it's a thin wrapper, but worth noting.

**One concern:**
- If `origin/main` doesn't exist (e.g. shallow clone, or remote named differently), the function silently returns empty dict and no merge happens. This is safe but could mask configuration issues. A log line at debug level (e.g. "origin/main not found, skipping merge") would help operators.

No correctness issues. The fix addresses the exact regression described.

PR #2486, opened 2026-06-05
feat(notebook_tools): add execute options --kernel/--cwd/--env/--scrub-keys (See #2476)

jsboige:
## Summary

Consolidate ad-hoc environment/kernel overrides into `execute_with_papermill()` with 4 new parameters and matching CLI arguments.

### Changes

**`scripts/notebook_tools/notebook_tools.py`** (+75 lines):
- `SCRUB_KEYS` class variable: 8 LLM API key names to scrub from subprocess env
- `execute_with_papermill()` new parameters:
  - `kernel_override: Optional[str]` — override `detect_kernel_name()`
  - `cwd_override: Optional[Path]` — override notebook dir as working directory
  - `env_extra: Optional[dict]` — inject extra env vars into subprocess
  - `scrub_keys: bool` — remove LLM API keys from subprocess env
- CLI arguments: `--kernel`, `--cwd PATH`, `--env KEY=VAL` (repeatable), `--scrub-keys`
- All defaults preserve existing behavior (None/False)

**`scripts/notebook_tools/tests/test_notebook_tools.py`** (+185 lines):
- `TestExecuteOptions` (9 tests): scrub_keys constant, env cleaning, signature, defaults, env build, merge, kernel/cwd override
- `TestExecuteCLIArgs` (6 tests): CLI parsing for kernel, cwd, env repeatable, scrub_keys flag, key=val parsing, invalid rejection
- **163/163 tests passed** (149 existing + 14 new)

## Validation

- `python -m pytest scripts/notebook_tools/tests/test_notebook_tools.py -v` → 163 passed in 2.79s
- No regression on existing tests
- Non-breaking: all new parameters default to None/False

## Scope

Single script, single test file. No notebooks touched. No dependency changes.

See #2476

### DIFF
--- a/scripts/notebook_tools/notebook_tools.py
+++ b/scripts/notebook_tools/notebook_tools.py
@@ -1072,10 +1072,32 @@ class NotebookExecutor:
             return 'lean4'
         return 'python3'
 
+    # Keys scrubbed by --scrub-keys to prevent live API calls during validation
+    SCRUB_KEYS = [
+        "OPENAI_API_KEY", "ANTHROPIC_API_KEY", "GOOGLE_API_KEY",
+        "MISTRAL_API_KEY", "DEEPSEEK_API_KEY", "HUGGINGFACE_API_KEY",
+        "HUGGINGFACEHUB_API_TOKEN", "OPENROUTER_API_KEY",
+    ]
+
     def execute_with_papermill(self, timeout: int = 300,
                                 output_path: Optional[Path] = None,
-                                batch_mode: bool = False) -> NotebookExecutionResult:
-        """Execute notebook using Papermill."""
+                                batch_mode: bool = False,
+                                kernel_override: Optional[str] = None,
+                                cwd_override: Optional[Path] = None,
+                                env_extra: Optional[dict] = None,
+                                scrub_keys: bool = False) -> NotebookExecutionResult:
+        """Execute notebook using Papermill.
+
+        Args:
+            timeout: Maximum execution time in seconds.
+            output_path: Path for the output notebook (default: <name>_output.ipynb).
+            batch_mode: If True, set BATCH_MODE=true in the notebook environment.
+            kernel_override: Force a specific kernel name instead of auto-detection.
+            cwd_override: Execute from this directory instead of the notebook's parent.
+            env_extra: Additional environment variables to inject into the subprocess.
+            scrub_keys: If True, remove LLM API keys from the subprocess environment
+                        to force deterministic mock paths.
+        """
         # Use base executor if available
         if self._base_executor:
             params = {"BATCH_MODE": "true"} if batch_mode else None
@@ -1094,7 +1116,7 @@ class NotebookExecutor:
             )
 
         # Fallback implementation
-        kernel = self.detect_kernel_name()
+        kernel = kernel_override or self.detect_kernel_name()
         start_time = time.time()
 
         if output_path is None:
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Solid feature, backward-compatible, good CLI ergonomics. Two issues worth addressing before merge.

---

### Bug: `scrub_keys` sets keys to empty string instead of removing them

Lines 1176-1178 in `notebook_tools.py`:

```python
for key in self.SCRUB_KEYS:
    sub_env.pop(key, None)
    sub_env[key] = ""        # <-- immediately re-adds as empty string
```

The `pop` removes the key, but the next line sets it back to `""`. The docstring says "remove LLM API keys" but the key remains in `os.environ` with an empty value. Code that checks `if os.getenv("OPENAI_API_KEY"):` happens to work (empty string is falsy), but code checking `if "OPENAI_API_KEY" in os.environ:` would not -- and some SDKs detect key presence vs absence differently.

Fix: remove the `sub_env[key] = ""` line. Just `pop` is enough.

The test `test_scrub_keys_env_build` (line ~1629) codifies this bug by asserting `sub_env.get("OPENAI_API_KEY") == ""`. That assertion should check the key is absent or use `key not in sub_env`.

---

### Tests validate behavior externally, not through the actual code path

The tests in `TestExecuteOptions` manually reconstruct the env-building logic (e.g. `test_scrub_keys_env_build` copies the pop+assign pattern into the test body). If the production code changes the order or logic, the test still passes because it duplicates the bug rather than calling the method.

`test_kernel_override_replaces_auto` tests Python's `or` operator (`"custom-kernel" or executor.detect_kernel_name()`), not `execute_with_papermill(kernel_override="custom-kernel")`.

`TestExecuteCLIArgs` builds a standalone `ArgumentParser` instead of importing the real one from `main()`. A typo in the real parser would go undetected.

These tests prove the concept but don't guard against regressions. Consider at minimum one integration test that mocks `subprocess.run` and calls `execute_with_papermill(scrub_keys=True)` to assert the env dict passed to `subprocess.run` has the keys absent.

---

### Minor: no sanitization on `--env` keys

`env_extra.update(env_extra)` passes user-supplied keys directly to the subprocess env. An `--env PATH=/tmp` would override the system PATH. For a developer CLI tool this is acceptable (power-user feature), but worth a one-line note in the `--env` help text: "Caution: overrides any existing env var."

---

No security blockers for a dev tool. The scrub_keys bug is the only thing I'd flag as a pre-merge fix.

clusterManager-Myia (commented):
**[NanoClaw]**

Previous review's two issues resolved. One remaining suggestion (non-blocking).

---

### Resolved: scrub_keys bug fixed

Commit `0dfe8ce` removes the offending `sub_env[key] = ""` line. Keys are now properly removed via `pop` without being re-added. The corresponding test (`test_scrub_keys_env_build`) now asserts `"OPENAI_API_KEY" not in sub_env` instead of checking for empty string. Clean fix.

### Resolved: test assertions match intent

The test suite now correctly validates that scrubbed keys are absent from the env dict, not merely empty. `test_scrub_keys_env_cleaned` still doesn't mock `subprocess.run` to test through the full code path (it checks the method signature + SCRUB_KEYS constant), but `test_scrub_keys_env_build` covers the actual env-building logic. Acceptable coverage for a dev tool.

### Suggestion: one integration test with mocked subprocess

The tests in `TestExecuteOptions` still largely duplicate the env-building logic in test bodies rather than calling `execute_with_papermill(scrub_keys=True)` and asserting on the env dict passed to `subprocess.run`. A single test using `unittest.mock.patch("subprocess.run")` would catch regressions if the production code's env-building logic drifts. Low priority -- not blocking merge.

### No security issues

SCRUB_KEYS list contains only key names (no values). Test fixtures use non-secret strings. `subprocess.run` call is unchanged (no `shell=True`). `--env` passthrough is documented as a power-user feature. Clean.

PR #2537, opened 2026-06-06
feat(scripts): wire execute overrides through base_executor + in-process lean4

jsboige:
#2476 — Consolider notebook_tools.py execute

## Summary

Wire the existing CLI options (`--kernel`, `--cwd`, `--env`, `--scrub-keys`) through ALL execution paths in the notebook executor.

### Changes

**`notebook_helpers.py` — BaseExecutor**
- Extended `execute_with_papermill` with `cwd_override`, `env_extra`, `scrub_keys` params
- Subprocess path: builds custom `sub_env` dict with scrubbed keys + extra env
- In-process lean4 path: applies env changes via temporary `os.environ` modification with proper restore
- `SCRUB_KEYS` list defined as class attribute (matching `notebook_tools.py`)

**`notebook_tools.py` — NotebookExecutor**
- Base executor call now passes through ALL overrides (previously ignored)
- In-process lean4 fallback: applies `env_extra` + `scrub_keys` via temporary env modification

### Non-regression

- Search-1-StateSpace: Papermill SUCCESS (12.5s default, 12.8s with scrub-keys+env)
- App-2-GraphColoring: Papermill SUCCESS (75.1s)
- Link-checker tests: 38/38 passing

### Acceptance criteria (#2476)

- [x] `execute --kernel`, `--cwd`, `--env`, `--scrub-keys` fonctionnels (subprocess ET in-process lean4)
- [x] Defaut inchange (non-regression sur les familles existantes)
- [x] Re-execution de validation realisable sans quitter notebook_tools.py

### DIFF
--- a/scripts/notebook_tools/notebook_helpers.py
+++ b/scripts/notebook_tools/notebook_helpers.py
@@ -1140,13 +1140,23 @@ class NotebookExecutor:
 
         return result
 
+    # Keys scrubbed by scrub_keys to prevent live API calls during validation
+    SCRUB_KEYS = [
+        "OPENAI_API_KEY", "ANTHROPIC_API_KEY", "GOOGLE_API_KEY",
+        "MISTRAL_API_KEY", "DEEPSEEK_API_KEY", "HUGGINGFACE_API_KEY",
+        "HUGGINGFACEHUB_API_TOKEN", "OPENROUTER_API_KEY",
+    ]
+
     def execute_with_papermill(
         self,
         notebook_path: str,
         output_path: Optional[str] = None,
         parameters: Optional[Dict[str, Any]] = None,
         kernel_name: Optional[str] = None,
-        timeout: Optional[int] = None
+        timeout: Optional[int] = None,
+        cwd_override: Optional[str] = None,
+        env_extra: Optional[Dict[str, str]] = None,
+        scrub_keys: bool = False,
     ) -> NotebookExecutionResult:
         """
         Execute a notebook using Papermill (subprocess).
@@ -1157,13 +1167,14 @@ class NotebookExecutor:
             parameters: Parameters to inject
             kernel_name: Kernel to use (auto-detected if None)
             timeout: Timeout in seconds
-
-        Returns:
-            NotebookExecutionResult
+            cwd_override: Execute from this directory instead of notebook parent.
+            env_extra: Additional environment variables to inject.
+            scrub_keys: If True, remove LLM API keys from subprocess environment.
         """
         import subprocess
         import time
         import sys
+        import os as _os
 
         start_time = time.time()
         kernel_name = kernel_name or self.detect_kernel(notebook_path)
@@ -1173,7 +1184,7 @@ class NotebookExecutor:
         if output_path is None:
             output_path = str(nb_path.parent / f"{nb_path.stem}_output{nb_path.suffix}")
         abs_output = str(Path(output_path).resolve())
-        abs_cwd = str(nb_path.parent)
+        abs_cwd = str(Path(cwd_override).resolve()) if cwd_override else str(nb_path.parent)
 
         # WSL-based kernels need longer startup time
         start_timeout = 120 if 'wsl' in kernel_name or kernel_name == 'smartcontracts' else 60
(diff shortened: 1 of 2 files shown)


jsboige (commented):
**[NanoClaw]**

## Review: #2537 — Wire execute overrides through base_executor + in-process lean4

Refs #2476. The PR extends `execute_with_papermill` in both `notebook_helpers.py` (BaseExecutor) and `notebook_tools.py` (NotebookExecutor) with three new parameters: `cwd_override`, `env_extra`, and `scrub_keys`. It also applies these overrides to the in-process lean4 execution path, which previously ran with the raw environment.

### What's done well

- The env restore pattern (`_old_env` dict + `try/finally`) is correct for both files. If a key didn't exist before, it gets popped on restore (`val is None`). This is the right approach for in-process `os.environ` mutation.
- Subprocess path correctly builds `sub_env = dict(os.environ)` as an isolated copy, so the scrub/extend logic doesn't leak into the parent process. Clean.
- `SCRUB_KEYS` is a reasonable allowlist of LLM provider keys, with a clear comment explaining intent.
- The `cwd_override` logic uses `Path(...).resolve()` consistently — good for avoiding relative-path ambiguity.

### Concerns

**1. Thread-safety risk with `os.environ` mutation (both files)**

Both the `notebook_helpers.py` and `notebook_tools.py` in-process lean4 paths temporarily mutate `os.environ` via `pop`/assignment. This is correct for single-threaded execution but is **not safe if notebooks are ever executed concurrently** (e.g., `ThreadPoolExecutor`, `concurrent.futures`, or async + sync mixing). `os.environ` is a process-global dict — concurrent mutations can cause keys to vanish or leak between executions.

Right now there's no parallel execution in the codebase, so this is a latent risk rather than a current bug. But it should at minimum get a code comment like:
```python
# NOTE: os.environ mutation is not thread-safe. Do NOT call this method
# from concurrent threads without an external lock.
```

**2. Duplicated `SCRUB_KEYS` + env restore logic across two files**

`SCRUB_KEYS` is now defined identically in both `notebook_helpers.py` (on BaseExecutor) and `notebook_tools.py` (on NotebookExecutor). The env save/restore pattern is also copy-pasted. If a new API key needs to be added to the scrub list, it must be updated in both places — a classic sync failure waiting to happen.

Given that `NotebookExecutor._base_executor` already delegates to `BaseExecutor` for the first code path, consider either:
- Having `NotebookExecutor` reference `BaseExecutor.SCRUB_KEYS` (if importable) rather than maintaining its own copy, or
- Extracting `SCRUB_KEYS` and the env scrub/restore logic into a small shared utility function.

This isn't blocking, but it's a maintenance hazard worth tracking.

**3. `_base_executor` delegation now passes `kernel_name` but skips `batch_mode` parameters**

The delegation path in `notebook_tools.py` now correctly forwards `cwd_override`, `env_extra`, and `scrub_keys` to `BaseExecutor.execute_with_papermill` — but `batch_mode` was already being converted to `parameters={"BATCH_MODE": "true"}` before this PR, so that's fine. Just flagging that the parameter set between the two `execute_with_papermill` signatures has drifted: the helpers version takes `parameters` (generic dict), while the tools version takes `batch_mode` (boolean) + `kernel_override`. This works because the delegation explicitly handles the translation, but it means these two methods are not drop-in interchangeable.

**4. `env_extra` values are not validated**

`env_extra` passes raw user-provided strings into both `os.environ` and subprocess environment. If `env_extra` values come from untrusted input (CLI args from a user are fine, but if ever from a config file or API), this could inject arbitrary environment variables. For the current CLI use case this is acceptable, but worth noting.

**5. `cwd_override` with non-existent directory is not validated early**

If `cwd_override` points to a non-existent path, `Path(...).resolve()` will succeed but papermill/subprocess will fail with an opaque error. A pre-check (`if cwd_override and not Path(cwd_override).is_dir(): raise ValueError(...)`) would give a clearer error. Minor UX issue.

### Verdict

The logic is sound, the env handling is correct for single-threaded use, and the override wiring through both execution paths (subprocess + in-process lean4) is complete. No secrets, no injection risks, no behavior-breaking changes for existing callers (new params are all optional with safe defaults).

The thread-safety concern and `SCRUB_KEYS` duplication are the items worth addressing, but neither blocks merge.

PR #2540, opened 2026-06-06
test(notebook_tools): add 16 tests for discover_notebooks, SCRUB_KEYS, helpers parity

jsboige:
## Summary

Extend `test_notebook_tools_pure.py` with 16 new tests (59 -> 74 passed + 1 skip):

- **TestDiscoverNotebooks** (5 tests): family resolution, single path, nonexistent, python_only filter, "all" target
- **TestNotebookExecutorScrubKeys** (5 tests): list type, OPENAI/ANTHROPIC presence, uppercase, non-empty
- **TestNotebookHelpersExecutor** (3 tests): import, method check, SCRUB_KEYS parity (skip until PR #2537 merges)
- **TestCmdExecuteEnvParsing** (3 tests): KEY=VAL parsing, no-equals, empty value

## Verification

```
74 passed, 1 skipped in 5.63s
```

See #2476

### DIFF
--- a/scripts/tests/test_notebook_tools_pure.py
+++ b/scripts/tests/test_notebook_tools_pure.py
@@ -1,7 +1,8 @@
 """Tests for scripts/notebook_tools/notebook_tools.py — pure functions and dataclasses.
 
 Tests focus on: should_skip, detect_kernel, extract_cell_preview,
-_format_markdown_table, _format_summary, dataclass defaults, Colors, SKIP_PATTERNS.
+_format_markdown_table, _format_summary, dataclass defaults, Colors, SKIP_PATTERNS,
+discover_notebooks, NotebookExecutor.SCRUB_KEYS, notebook_helpers parity.
 No filesystem I/O or kernel execution required.
 """
 
@@ -28,6 +29,8 @@ from notebook_tools import (
     _format_markdown_table,
     _format_summary,
     get_repo_root,
+    discover_notebooks,
+    NotebookExecutor,
 )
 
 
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[NanoClaw]** — test code review

Overall: 16 tests, 5 test classes, 104 lines added. Good coverage intent for previously-untested surface area (discover_notebooks, SCRUB_KEYS, helpers parity). A few concerns worth addressing.

---

### TestCmdExecuteEnvParsing — not testing any actual module code

`test_valid_pairs`, `test_invalid_pair_no_equals`, and `test_empty_value` (lines 501-523 in the diff) **re-implement parsing logic inline in the test** rather than calling any function from `notebook_tools.py`. These tests verify the test's own `split("=", 1)` — they'd pass even if `cmd_execute` used completely different parsing. The value of these tests is zero: they're testing Python's `str.split`, not project code.

Suggestion: either import and call the actual parsing path (likely inside `cmd_execute`), or remove these 3 tests entirely. If `--env` parsing is buried inside a CLI function that's hard to unit-test, a comment saying so would at least make the intent clear.

---

### TestNotebookExecutorScrubKeys — structural checks only, no behavioral test

`test_scrub_keys_is_list`, `_contains_openai`, `_contains_anthropic`, `_all_uppercase`, `_no_empty` are all structural/invariant checks on a static class variable. They'll never fail unless someone does something obviously wrong. The only value here is `_all_uppercase` catching a naming convention slip — the rest are noise.

Missing: a behavioral test that verifies SCRUB_KEYS actually works as intended. For example:
```python
def test_scrub_keys_removes_keys_from_env(self):
    mock_env = {"OPENAI_API_KEY": "sk-xxx", "PATH": "/usr/bin", "HOME": "/home"}
    for key in NotebookExecutor.SCRUB_KEYS:
        mock_env.pop(key, None)
    assert "OPENAI_API_KEY" not in mock_env
    assert "PATH" in mock_env  # non-key vars preserved
```
This mirrors the actual usage in `execute_with_papermill` (line 1177 of notebook_tools.py) and would catch a regression if scrub logic changed.

---

### TestDiscoverNotebooks::test_python_only_filter — assertion is trivially true

The test calls `discover_notebooks("IIT", root, python_only=True)` then asserts `nb.suffix == ".ipynb"` for each result. But `discover_notebooks` already filters to `.ipynb` files via `rglob('*.ipynb')` *before* applying the `python_only` kernel filter. So `python_only` could be completely broken and this assertion would still pass.

Better assertion: since IIT's config kernel is `"python"`, the `python_only` filter should not reduce the count below the unfiltered count:
```python
def test_python_only_filter(self):
    root = get_repo_root()
    nbs_all = discover_notebooks("IIT", root, python_only=False)
    nbs_py = discover_notebooks("IIT", root, python_only=True)
    assert len(nbs_py) == len(nbs_all)  # IIT is Python-only family
```

---

### TestNotebookHelpersExecutor::test_scrub_keys_parity_after_merge — fragile conditional skip

The `pytest.skip("BaseExecutor.SCRUB_KEYS not yet added (awaiting PR #2537 merge)")` path makes this test silently skip depending on merge order. Once PR #2537 merges, the skip path becomes dead code; if it never merges, the test never actually verifies anything. This is acceptable as a temporary bridge, but should have a TODO/comment to remove the skip branch once #2537 is in.

Also: `notebook_helpers.py` (line 761) has a `class NotebookExecutor` but **no `SCRUB_KEYS` attribute** on the main branch. So this test will skip for now, which is fine, but worth noting.

---

### TestDiscoverNotebooks::test_all_target — magic number 100

`assert len(nbs) > 100` with the comment "Repo has 500+ notebooks" — if the repo drops below 101 notebooks for any reason (directory restructuring, notebooks moved to a new family not in NOTEBOOK_FAMILIES), this test fails misleadingly. Consider `> 50` as a less brittle lower bound, or count the expected families and verify proportionally.

---

### Import convention — consistent with existing file

The PR adds `discover_notebooks` and `NotebookExecutor` to the existing `sys.path.insert` / direct import block at the top of the test file. This is correct and matches the repo convention. No issue here.

---

### Security scan

No credential leaks, no hardcoded secrets, no dangerous patterns. The SCRUB_KEYS values (OPENAI_API_KEY, etc.) are key names, not values. Clean.

---

**Summary**: 7 of 16 tests have issues — 3 test Python builtins instead of project code, 4 SCRUB_KEYS tests are shallow structural checks with no behavioral verification, 1 discover_notebooks assertion is trivially true. The parity test is a reasonable bridge to PR #2537. Import convention is correct. Security clean.

PR #2541, opened 2026-06-06
fix(tests): isolate genai_audio env leak + fix WSL case-sensitive assertion

jsboige:
## Summary

Fix 2 pre-existing test failures (11→9 remaining, all GARCH `arch` module missing):

1. **test_genai_audio_apis** (`test_no_auth_without_env`): `patch.dict(os.environ, {}, clear=True)` only clears env vars but `_get_auth_headers()` also reads from `MyIA.AI.Notebooks/GenAI/.env` on disk. Added `patch.object(Path, "exists", return_value=False)` to fully isolate the test.

2. **test_wsl_papermill** (`test_d_drive`): `Path("D:\dev\project").resolve()` capitalizes to `D:\Dev\project` based on physical disk. Changed assertion to case-insensitive `.lower()` check.

## Verification

```
58/58 passed in test_wsl_papermill.py + test_genai_audio_apis.py
Full suite: 2189 passed, 9 failed (all GARCH), 18 skipped
```

### DIFF
--- a/scripts/tests/test_genai_audio_apis.py
+++ b/scripts/tests/test_genai_audio_apis.py
@@ -228,7 +228,8 @@ class TestGetAuthHeaders:
 
     def test_no_auth_without_env(self):
         """Without env vars or files, no Authorization header."""
-        with patch.dict("os.environ", {}, clear=True):
+        with patch.dict("os.environ", {}, clear=True), \
+             patch.object(Path, "exists", return_value=False):
             result = _get_auth_headers()
             assert "Authorization" not in result
 
--- a/scripts/tests/test_wsl_papermill.py
+++ b/scripts/tests/test_wsl_papermill.py
@@ -52,10 +52,11 @@ class TestWinToWslPath:
         assert "Users/test/file.txt" in result
 
     def test_d_drive(self):
-        """D: drive maps to /mnt/d."""
+        """D: drive maps to /mnt/d (case-insensitive check due to resolve())."""
         result = win_to_wsl_path("D:\\dev\\project")
         assert result.startswith("/mnt/d/")
-        assert "dev/project" in result
+        # Path.resolve() may capitalize path components based on physical disk
+        assert "dev/project" in result.lower()
 
     def test_lowercase_drive(self):
         """Drive letter is lowercased in WSL path."""


clusterManager-Myia:
**[NanoClaw]**

Reviewed 2 files, 2 commits. Security scan: clean (no secrets/tokens in diff).

## Verdict: LGTM with 1 minor nit

### Change 1: `test_no_auth_without_env` env leak isolation — **correct**

The root cause is clear: `_get_auth_headers()` checks `env_file.exists()` *before* looking at `os.environ`. On CI or dev machines where `MyIA.AI.Notebooks/GenAI/.env` exists on disk, `patch.dict(os.environ, {}, clear=True)` only blocks the env-var path — the file path still leaks through, producing an `Authorization` header and failing the assertion.

The fix adds `patch.object(Path, "exists", return_value=False)` alongside the existing env clear, which fully isolates both code paths (file check + env vars). This is the right approach.

One observation: the broader `TestGetAuthHeaders` class already has `_mock_env_file` (line ~242) which demonstrates the author understands the dual-source auth pattern. The simpler `test_no_auth_without_env` just needed the same file-source isolation, and the fix delivers it cleanly with minimal diff.

### Change 2: `test_d_drive` case-insensitive assertion — **sound, pragmatic**

`win_to_wsl_path` calls `Path(win_path).resolve()` which on WSL (Linux) runs `realpath` — case-sensitive, no capitalization. On Windows, `Path.resolve()\) normalizes to the physical disk's actual casing, so `D:\\dev\\project` might resolve to `D:\\Dev\\project` depending on how the directory was created.

The fix (lowercasing the result before asserting) is a reasonable test accommodation for the platform difference. The function itself isn't broken — it always produces `/mnt/d/...` with forward slashes — so the test just needed to be more tolerant of Windows casing behavior.

### Minor nit: `test_long_path` may have the same fragility

At `test_wsl_papermill.py:97`, `test_long_path` asserts `"Sudoku/nb.ipynb" in result` without lowercasing. If the physical directory on disk is e.g. `CoursIA-2` but `resolve()` produces `coursia-2` (or vice versa), this test will also be case-sensitive. Consider applying the same `.lower()` treatment there for consistency — or if the directory is known to exist on CI with consistent casing, a comment explaining why it's safe.

No action required, just a consistency note for future maintenance.

PR #2544, opened 2026-06-06
test(scripts): add 26 tests for repair_genai_notebooks.py (See #2149)

jsboige:
## Summary

Add 26 unit tests for scripts/repair_genai_notebooks.py covering all pure functions.

## Coverage

| is_cell_corrupted | 13 | Empty/None/string source, normal source, collapsed lines (4 keywords), short lines, missing newlines, threshold, last element, empty strings |
| fix_source | 7 | Empty/None/string source, normal source, corrupted source fix, last element preservation, collapsed line fix |
| repair_notebook | 6 | Clean notebook, corrupted notebook, dry run, mixed cells, multiple corruptions, no code cells |

## Test results

26 passed in 0.17s

See #2149

### DIFF
--- a/scripts/notebook_tools/tests/test_repair_genai_notebooks.py
+++ b/scripts/notebook_tools/tests/test_repair_genai_notebooks.py
@@ -1,0 +1,254 @@
+"""Tests for scripts/repair_genai_notebooks.py — GenAI notebook corruption repair."""
+
+import importlib.util
+import json
+from pathlib import Path
+
+import pytest
+
+# Load module by file path
+_MOD_PATH = (
+    Path(__file__).resolve().parent.parent.parent.parent
+    / "scripts"
+    / "repair_genai_notebooks.py"
+)
+_spec = importlib.util.spec_from_file_location("repair_genai_notebooks", _MOD_PATH)
+_mod = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_mod)
+
+is_cell_corrupted = _mod.is_cell_corrupted
+fix_source = _mod.fix_source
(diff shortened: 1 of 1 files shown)


clusterManager-Myia:
**[NanoClaw]**

Clean test file, 26 tests across 3 test classes covering all three public functions of `repair_genai_notebooks.py`. Good coverage of boundary cases (empty, None, non-list inputs), both corruption detection methods, and the dry-run / file-write behavior. No security concerns.

**Import convention divergence** — the repo's existing tests (e.g. `test_fix_robust_dotenv.py`, `test_notebook_tools.py`) use `sys.path.insert(0, ...)` + standard imports. This PR uses `importlib.util.spec_from_file_location`. Both work; the `importlib` approach avoids polluting `sys.path` and is arguably better. Minor consistency note, not a blocker.

**One misleading comment** in `test_two_missing_newlines_not_flagged` (line ~93):
```python
source = ["import os", "from dotenv import load_dotenv", "x = 1\n"]
# Only 2 non-empty mid-array elements without \n (below threshold of 3)
assert is_cell_corrupted(source) is False
```
The comment implies the test passes because only 2 elements are missing `\n` (below the threshold of 3). In reality, `is_cell_corrupted` returns False here because `len(source) == 3` and Method 2 only triggers when `len(source) > 3` — the threshold check on element count is never reached. The assertion is still correct, but the reasoning in the comment doesn't match the code path. Consider either fixing the comment to reflect the actual reason (list too short for Method 2), or adding a 4th element so the test exercises the "2 missing newlines vs threshold of 3" logic it claims to test.

No other issues. LGTM.

PR #2545, opened 2026-06-06
test(scripts): add 25 tests for whisper_int8_comparison.py (See #2149)

jsboige:
## Summary

Add 25 unit tests for scripts/whisper_int8_comparison.py covering the 3 pure helper functions.

## Coverage

| normalize | 10 | Basic, French punctuation, comma/question/period/exclamation strip, lowercase, strip, empty, only punctuation |
| word_accuracy | 10 | Perfect/partial/no match, empty ref edge case (returns 1.0), empty hypothesis, both empty, case insensitive, punctuation ignored, subset match, repeated words |
| print_summary | 5 | Empty results, all perfect, mixed accuracy, OK count threshold (>=0.8), missing accuracy key |

## Test results

25 passed in 0.09s

See #2149

### DIFF
--- a/scripts/notebook_tools/tests/test_whisper_int8_comparison.py
+++ b/scripts/notebook_tools/tests/test_whisper_int8_comparison.py
@@ -1,0 +1,139 @@
+"""Tests for scripts/whisper_int8_comparison.py — Whisper comparison helpers."""
+
+import importlib.util
+import os
+from pathlib import Path
+
+import pytest
+
+# Set required env vars BEFORE importing the module (it raises EnvironmentError at import)
+os.environ.setdefault("TTS_API_KEY", "test-key-for-unit-tests")
+os.environ.setdefault("WHISPER_API_KEY", "test-key-for-unit-tests")
+
+# Load module by file path
+_MOD_PATH = (
+    Path(__file__).resolve().parent.parent.parent.parent
+    / "scripts"
+    / "whisper_int8_comparison.py"
+)
+_spec = importlib.util.spec_from_file_location("whisper_int8_comparison", _MOD_PATH)
+_mod = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_mod)
+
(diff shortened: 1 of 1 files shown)


clusterManager-Myia:
**[NanoClaw]**

Good test coverage overall. 25 tests across 3 classes, all with specific expected values -- no always-pass patterns or trivially-true assertions. Security scan: clean.

**One concern -- import convention mismatch:**

The existing test files in this repo use `sys.path.insert` + standard imports:

```python
sys.path.insert(0, str(Path(__file__).parent.parent))
from audit_c1_c3 import check_c1, check_c3, ...
```

This PR uses `importlib.util.spec_from_file_location` instead. While it works, it deviates from the established repo convention and is harder to read. Consider switching to the `sys.path.insert` pattern for consistency, with `os.environ.setdefault(...)` calls placed before the import (as you already do).

**Edge case note (non-blocking):**

`os.environ.setdefault("TTS_API_KEY", "test-key-for-unit-tests")` only sets the variable if the key is *absent*. If the key exists but is an empty string (e.g. from a CI env), `setdefault` leaves the empty value and the module's `if not TTS_API_KEY` check will still raise `EnvironmentError`. Using `os.environ["TTS_API_KEY"] = os.environ.get("TTS_API_KEY", "test-key")` would cover both cases. This is unlikely to bite in practice but worth a line fix if you're already touching the import block.

**Test quality details:**

- `normalize` (10 tests): thorough -- punctuation chars tested individually, French hyphenation, whitespace, empty input, all-punctuation input. All match the source implementation.
- `word_accuracy` (10 tests): good edge case coverage. `test_empty_reference_returns_one` and `test_both_empty` asserting `1.0` correctly match the `if not ref_words: return 1.0` guard. `test_repeated_words` asserting `1.0` is correct since the function checks membership, not frequency.
- `print_summary` (5 tests): clean use of `capsys` for output assertions, `pytest.approx` for float comparison, and a missing-key edge case.

PR #2546, opened 2026-06-06
test(scripts): add 17 tests for mcp_buffering_smoke_test.py (See #2149)

jsboige:
## Summary

Add 17 unit tests for scripts/mcp_buffering_smoke_test.py covering the create_stress_notebook function.

## Coverage

| create_stress_notebook | 17 | Return type, nbformat 4/5, kernelspec, language_info, file written to disk, cell count per mode (long_output=4, reconnect=1, rapid_calls=1, unknown=0), all cells code type, required fields, null execution_count, empty outputs, source is list, reconnect content, rapid_calls content, long_output content |

## Test results

17 passed in 0.15s

See #2149

### DIFF
--- a/scripts/notebook_tools/tests/test_mcp_buffering_smoke_test.py
+++ b/scripts/notebook_tools/tests/test_mcp_buffering_smoke_test.py
@@ -1,0 +1,133 @@
+"""Tests for scripts/mcp_buffering_smoke_test.py — MCP buffering smoke test helpers."""
+
+import importlib.util
+import json
+from pathlib import Path
+
+import pytest
+
+# Load module by file path
+_MOD_PATH = (
+    Path(__file__).resolve().parent.parent.parent.parent
+    / "scripts"
+    / "mcp_buffering_smoke_test.py"
+)
+_spec = importlib.util.spec_from_file_location("mcp_buffering_smoke_test", _MOD_PATH)
+_mod = importlib.util.module_from_spec(_spec)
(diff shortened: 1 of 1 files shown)


clusterManager-Myia:
**[NanoClaw]**

LGTM. 17 tests covering all three modes (`long_output`, `reconnect`, `rapid_calls`) plus the unknown-mode fallback. The import approach via `importlib.util` is clean and avoids `sys.path` manipulation.

A couple of minor observations (non-blocking):

1. **Heavy structural coupling**: Several tests (`test_kernelspec`, `test_language_info`, `test_nbformat`) assert on values that are hardcoded literals in the source (`display_name == "Python 3"`, `nbformat_minor == 5`). These are tautological -- they'd only fail if the constant was renamed, not if a bug was introduced. Not harmful, but they add maintenance noise if those constants ever change intentionally. Consider parameterizing against the source module's own constants or dropping them in favor of the more meaningful content checks (L112-130) which actually verify generated cell source.

2. **Missing edge case**: No test for what happens with `mode="long_output"` cells 2-4 content. `test_long_output_first_cell_mentions_stress` only checks cell 0. The "rapid burst prints" and "mixed output" cells (cells 1-2) have no content verification. A test checking cell 1 contains `"Burst"` or cell 2 contains `"DataFrame"` would close the gap.

Neither is a blocker. Tests are functional and will catch regressions on the factory output structure.

PR #2376, opened 2026-06-04
enrich(Lean-14b): 3-pillars showcase — Exercice 4 heading + Pillars.lean source + scaffold interpretation (See #2155)

jsboige:
## Summary

Enrich Lean-14b notebook + Pillars.lean with content anchored on the actual `.lean` source code, advancing Epic #2155 acceptance criteria.

### Pillars.lean (Conway/Life/Pillars.lean)
- **Wire RLE import** (`import Conway.Life.RLE` — was commented out)
- **Add pulsar_period3 theorem**: `evolveHashlifeFast 3 pulsarGrid = pulsarGrid` proven by `native_decide` — first RLE-proven witness in the module
- **Enrich docstrings** for all 4 pillars (OTCA, UnitCell, Gemini, CPU) with historical details and community context
- Sorry count: **4** (unchanged), 0 new sorry

### Notebook (Lean-14b-Conway-Game-of-Life-Lean.ipynb)
- **Update Pillars.lean showcase cell** to reflect RLE import activation + pulsar theorem
- **Update scaffold interpretation** markdown with pipeline explanation
- Add Exercice 4 heading + stub code cell (RLE witness explorer)
- Papermill SUCCESS: **19/19 code cells, 0 errors**

## Verification
- `lake build Conway.Life.Pillars`: **SUCCESS** (3323 jobs)
- `grep -c sorry Pillars.lean`: **4** (unchanged from before)
- Papermill: **48/48 cells, 0 errors, 6m57s**
- No exercise stubs de-stubbed

## #2155 Acceptance Criteria Progress
- [x] Section dédiée par pilier (metacell → cpu → gemini) avec tableau caractéristiques — done in Actes I/II/III + synthese table
- [x] `Pillars.lean` : defs OR sorry roadmap documenté, `grep -c sorry` reported, Lake build SUCCESS — RLE import wired, pulsar_period3 proven, sorry=4 reported
- [ ] RLE téléchargés + parsés via `RLE.parseRLE` réel (output collé) — RLE parser proven in 9.8; pillar patterns too large for Lean string literals (OTCA=70KB)
- [ ] Python recadré (illustrations, pas preuves), chemin auto-détecté — partially done (cross-platform utils)
- [x] 3 exercices C.1 + re-exec C.2 complète — 4 exercises (Moore, period, still-life, witness explorer)
- [ ] Renommage `Lean-14b` + catalog regen — not yet

See #2155

PR #2547, opened 2026-06-06
test(catalog): add 31 tests for merge-curated-fields and text normalization

jsboige:
## Summary

Add test coverage for 4 previously-untested pure functions in `generate_catalog.py`:

- **`_merge_curated_fields`** (13 tests) — CRITICAL: verifies the stale-catalog poison prevention logic (#2376/#2383/#2385). Ensures curated git fields (`last_validation`, `last_validator`, `issue_pr_associee`) are preserved from `origin/main` when regenerating catalog on a branch behind main.
- **`_normalize_text`** (7 tests) — accent stripping, code fence removal, apostrophe normalization for intro/conclusion detection.
- **`_is_papermill_injected`** (5 tests) — Papermill parameter cell detection via injected-parameters tag.
- **`_is_comment_only_cell`** (6 tests) — comment-only code cell detection (blank lines, whitespace, pure comments).

## Test results

```
166 passed in 0.44s (135 -> 166, +31 new tests)
Full suite: 2227 passed, 2 failed (pre-existing, fixed in #2541), 18 skipped
```

## Files changed

- `scripts/tests/test_generate_catalog.py` -- +249 lines (4 new test classes)

### DIFF
--- a/scripts/tests/test_generate_catalog.py
+++ b/scripts/tests/test_generate_catalog.py
@@ -15,9 +15,14 @@ import pytest
 
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "notebook_tools"))
 from generate_catalog import (
+    CURATED_GIT_FIELDS,
     _effective_code_cells,
+    _is_comment_only_cell,
     _is_exercise_stub,
     _is_outputless_by_design,
+    _is_papermill_injected,
+    _merge_curated_fields,
+    _normalize_text,
     check_errors,
     classify_maturity,
     count_todos,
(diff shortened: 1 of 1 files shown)


clusterManager-Myia:
**[NanoClaw]**

Solid test addition — 32 tests across 4 new test classes covering `_merge_curated_fields`, `_normalize_text`, `_is_papermill_injected`, and `_is_comment_only_cell`. No secrets, no always-pass patterns, import convention matches the existing file.

**_merge_curated_fields (12 tests) — thorough**

The core merge logic is well-exercised: empty/None main catalog, empty entries, blank-field fill, non-empty preservation, mixed fill+keep, missing-key-as-empty, no-match-in-main, multiple entries, non-curated-field isolation, in-place mutation check, and the CURATED_GIT_FIELDS constant assertion. Tests align with the actual implementation at line 633 — the early-return on `not main_catalog`, the per-field `not current_val` gate, and the in-place mutation pattern are all verified.

One minor note: `test_curated_git_fields_constant` (asserting the exact 3-tuple) is a coupling point — if someone adds a fourth curated field, this test breaks instead of guiding. Acceptable as a regression gate, but worth knowing.

**_normalize_text (8 tests) — good, one weak assertion**

Covers lowercase, accent stripping, curly-to-straight apostrophe, fenced code removal (both single-tag and generic multiline), empty string, and content preservation. All match the implementation at line 264.

`test_strips_accents` uses `"synthese" in result` rather than an exact equality check. Since `_normalize_text` also lowercases and strips accents, the exact output is deterministic — `assert _normalize_text("synthèse") == "synthese"` would be stricter and catch regressions where accent-stripping breaks silently (e.g. returning an empty string). Currently if the function returned `"xyz"`, the test would still fail, but if it returned `"synthese "` with a trailing space, it would pass despite a bug.

**_is_papermill_injected (5 tests) — complete**

For a one-liner (line 321), this is exhaustive: injected tag present, no tags, wrong tags, empty tags list, and no metadata at all.

**_is_comment_only_cell (7 tests) — strong**

Covers the key edge cases well: pure comments, code+comment mix, empty source, whitespace-only, mixed comments+blank lines, actual code, and the triple-quoted string case (important since the function checks `startswith("#")` not AST, so `"""docstring"""` correctly returns False). Aligns with line 326 implementation.

**Minor nits**

- PR title says 31 tests, actual count is 32 (12+8+5+7). Off by one — no impact, just a title mismatch.
- No test for `_normalize_text` with input containing both fenced code and accented text simultaneously — not a gap that matters in practice since the regex removal runs before normalization, but a combined edge case would be the cherry on top.

Overall: clean, well-structured, meaningful assertions. LGTM with the minor note on `test_strips_accents` exact-match improvement.

PR #2599, opened 2026-06-07
fix(SymbolicAI): align README notebook counts with CATALOG-STATUS markers

jsboige:
## Summary

Align stale notebook counts in `SymbolicAI/README.md` prose with the CATALOG-STATUS markers (source of truth). The Lean series grew from 16 to 21 notebooks (Conway + Grothendieck expansions), Planners archive was reclassified, and the global total drifted from 99 to 104.

See #2161

## Changes

| Location | Before | After | Source |
|----------|--------|-------|--------|
| Overview table Lean row | 16 notebooks, 14 (88%), ~11h | 21 notebooks, 19 (90%), ~14h | CATALOG-STATUS `Lean=21` |
| Overview table Planners row | ~8h | ~7h | CATALOG-STATUS `Planners=13` |
| Total line | 99 notebooks, ~65h | 104 notebooks, ~70h | CATALOG-STATUS `pedagogical_count: 104` |
| Lean section description | 16 notebooks, Kochen-Specker | 21 notebooks, Conway + Kochen-Specker + Free Will | Actual files |
| Planners section description | 14 notebooks | 13 notebooks | CATALOG-STATUS (excl archive) |
| Planners exercice note | 13/14 | 12/13 | Archive excluded |
| Directory tree Lean | 16 notebooks | 21 notebooks | CATALOG-STATUS |
| Directory tree Planners | 14 notebooks | 13 notebooks | CATALOG-STATUS |
| Audit table (updated 20 mars -> 7 juin) | 80/98 (82%) | 91/102 (89%) | Recount + SymbolicLearning row |
| Audit total prose | 80/98 | 91/102 | Recount |

## Verification

- `grep` for stale counts (16, 14, 99, 98): 0 matches after fix
- CATALOG-STATUS markers: **untouched** (automation-owned per catalog-pr-hygiene rule)
- `git diff --stat COURSE_CATALOG*`: 0 changes (no catalog drift)
- 1 file changed, 13 insertions, 12 deletions (pure replacement)

## Scope

Single README file, documentation-only. No code or notebook changes.

PR #2603, opened 2026-06-07
refactor(tests): consolidate 30 duplicate test files into canonical locations (See #2149)

jsboige:
## Summary

Consolidate 30 duplicated `test_*.py` files between `scripts/tests/` and `scripts/notebook_tools/tests/` into their canonical locations.

### Convention
- `scripts/notebook_tools/X.py` → `scripts/notebook_tools/tests/test_X.py`
- `scripts/X.py` → `scripts/tests/test_X.py`

### Changes

| Action | Count | Detail |
|--------|-------|--------|
| Moved ST→NT (more complete ST version) | 22 | `scripts/tests/` version had more tests, moved to `scripts/notebook_tools/tests/` |
| Kept NT (already canonical + more complete) | 6 | `notebook_tools/tests/` already had more tests, deleted ST copy |
| Kept ST (canonical for scripts/ source) | 2 | `test_check_docs_links`, `test_repair_genai_notebooks` |
| Kept ST (import path issues) | 2 | `test_genai_notebooks`, `test_genai_validate` — genai-stack `sys.path` conflicts when moved to NT |

### Import path fixes
All moved files had `sys.path.insert` adjusted for new directory depth:
- `parent.parent / "notebook_tools"` → `parent.parent` (for files moved to nt/tests/)
- `parent.parent / "genai-stack"` → `parent.parent.parent / "genai-stack"` (for genai files)

### Test results
- **3102 passed**, 12 failed (all pre-existing: 11 genai_config `sys.modules["config"]` pollution, 1 generate_catalog assertion)
- **8 fewer failures** than main (eliminated duplicate genai_config test execution)
- **-8372 net lines** of duplication removed

See #2149 (mutation testing Epic — test infrastructure cleanup)

PR #2564, opened 2026-06-06
enrich(Infer): add 3rd exercise to Infer-17 and Infer-6 (See #2161)

jsboige:
## Summary

Add 1 exercise stub to each of 2 Infer.NET notebooks currently at 2/3 exercises (convention #2161):

- **Infer-17-Decision-Networks**: Exercise on quality control decision network (Infer.NET API: Variable.Bernoulli, Variable.If/IfNot, InferenceEngine.Infer<Bernoulli>(), observation enumeration for optimal policy)
- **Infer-6-TrueSkill**: Exercise on online learning convergence tracking (TrueSkillOnline, mu/sigma history across 8 matches, posterior-as-prior updates, upset impact analysis)

## Verification

- [x] Both exercises use stub patterns (Console.WriteLine, no raise NotImplementedError/assert False/1/0)
- [x] No existing cells modified (28 insertions, 0 deletions)
- [x] Valid JSON structure verified
- [x] Exercises test concepts not covered by existing exercises
- [x] Positioned before conclusion/summary sections
- [x] Markdown-only additions (C.2 exception: no re-execution needed)

## Related

See #2161 (3 exercises per notebook convention)

PR #2605, opened 2026-06-07
fix(catalog): support Serie-SousSerie format in expand_catalog_markers (See #2599)

jsboige:
## Summary
Fix catalog CATALOG-STATUS markers producing `pedagogical_count: 0` for sub-series like `SymbolicAI-Lean`.

### Root cause
CATALOG-STATUS markers use `Serie-SousSerie` format (e.g. `SymbolicAI-Lean`) but the catalog JSON stores `serie='SymbolicAI'` + `sous_serie='Lean'`. The filter `e.get('serie') == 'SymbolicAI-Lean'` found 0 entries.

### Fix
Added `_filter_by_series()` helper that tries exact match first, then falls back to parsing `Serie-SousSerie` format. Applied to all 5 filter-by-serie call sites:
- `compute_counter`
- `compute_breakdown`
- `compute_maturity_distribution`
- `compute_status_distribution`
- `format_catalog_status_block`

### Verification
- SymbolicAI-Lean: `pedagogical_count: 0` -> `pedagogical_count: 21` with PRODUCTION=19, ALPHA=1, BETA=1
- `expand_catalog_markers.py --dry-run` correctly detects drift
- All other series unaffected (exact match takes priority)
- 61/61 tests passed (+11 new)

Reported by po-205 in dashboard (bug catalogue SymbolicAI-Lean `pedagogical_count=0`).

See #2599


clusterManager-Myia (commented):
**[NanoClaw]**

Good fix for the Serie-SousSerie filter mismatch. A few observations:

**The fix is sound:**
- `_filter_by_series()` tries exact match first, then falls back to `split("-", 1)` for `Serie-SousSerie` format. The `split(..., 1)` (maxsplit=1) is correct — handles multi-dash sous-series like `sous_serie="B-C"` when querying `"A-B-C"`.
- Exact match takes priority over fallback — properly tested in `test_exact_match_takes_priority`.
- Applied consistently to all 5 filter call sites. Good.

**COURSE_CATALOG.generated.json changes:**
- The bulk update replacing `issue_pr_associee: "#2161, #2564"` with per-notebook specific references is the catalog regeneration reflecting recent PR merges. This is the *correction* of the #2161 blanket overwrite pattern, not its introduction. The cell count and date updates for a few entries reflect legitimate notebook changes merged since the last catalog run. No concerns.

**One potential edge case (not tested, but unlikely in practice):**
- If a serie name itself contains a dash and NO exact match exists (i.e. the catalog has `serie="A-B"` but someone queries with `"A-B"` and exact match fails because the field is missing or empty), the fallback would try `parent="A", sous_serie="B"`. This would return wrong results. However, exact match always runs first and the test `test_exact_match_takes_priority` confirms this path. The only way to hit this would be if `e.get("serie")` returned something different from the queried name despite being "the same" — which would be a data integrity bug, not a code bug. Not blocking.

**Minor suggestion (non-blocking):**
- `_filter_by_series` could benefit from a `@functools.lru_cache` if the same serie is queried repeatedly across multiple marker expansions in the same README, but given the dataset size this is premature optimization.

**Test quality:** 11 new tests covering exact match, sous-serie format, multiple sous-series, root sous-serie, nonexistent queries, and integration with `format_catalog_status_block`/`compute_counter`/`compute_breakdown`. Solid.

No security concerns. Safe to merge.

clusterManager-Myia (commented):
**[NanoClaw]**

Re-review after SHA change (f3adc49 → 327da33). New commit: "restore origin/main catalog (de-poison incidental regen)".

**What changed:** Only `COURSE_CATALOG.generated.json` — the author reverted the incidental metadata drift from the auto-regen commit (commit 2) back to values consistent with `origin/main`. No code changes.

**Code review (unchanged from first review, still applies):**

The `_filter_by_series` fix is clean and correct:

- Exact match first, then `split("-", 1)` fallback — right priority order.
- `split("-", 1)` correctly handles sous_series containing dashes (e.g. `Lean4-Intro` → parent=`SymbolicAI`, child=`Lean4-Intro`).
- Returns `[]` for non-existent series — callers already handle empty lists.
- `test_exact_match_takes_priority` covers the ambiguity edge case where a serie literally equals `A-B`.

All five call sites updated consistently (`compute_counter`, `compute_breakdown`, `compute_maturity_distribution`, `compute_status_distribution`, `format_catalog_status_block`).

**One minor note:** The catalog JSON diff is large (60+ entries with `issue_pr_associee` changes from `#2161, #2564` to per-PR values). This is expected — the auto-regen in commit 2 re-scanned all notebooks and picked up their actual PR associations from main. The third commit cleaned it up. Not a concern for merge.

**Verdict:** LGTM. The code change is minimal, well-tested, and the catalog reconciliation is standard generated-file hygiene.

jsboige:
Hold — CONFLICTING/DIRTY (conflit avec la relocation des tests #2603). Deux actions :
1. Rebaser sur `origin/main` (post #2603 + #2612).
2. **Verifier la non-redondance avec #2612** (merged) : #2612 a corrige le marker SymbolicLearning cote DONNEE. #2605 corrige le PARSER (`expand_catalog_markers` accepte le format `Serie-SousSerie`) = amelioration de robustesse generale, valable SI le parser doit gerer les deux formats. Confirmer la valeur ajoutee post-#2612 (sinon fermer). NE PAS regen le catalogue sur la branche.

clusterManager-Myia (commented):
**[NanoClaw]**

## Review #3 — focused on #2161 claim, edge cases, and test completeness

### Does this fix the #2161 blanket overwrite bug?

**No — not in code.** This PR's code change is a filter fix for #2599 (sub-series matching). `expand_catalog_markers.py` only *reads* the catalog to expand README markers — it never writes to `COURSE_CATALOG.generated.json`. The blanket overwrite bug that stampeded all entries with `issue_pr_associee: "#2161, #2564"` would originate in `generate_catalog.py` (the script that *produces* the catalog JSON).

What this PR *does* correct: the catalog JSON metadata itself. The 3-commit structure shows:
- **Commit 1** (`f0519a9`): The actual code fix — `_filter_by_series` for Serie-SousSerie support
- **Commit 2** (`f3adc49`): Auto-regen catalog (picked up per-notebook PR refs from source notebooks)
- **Commit 3** (`327da33`): De-poisoned incidental regen drift, keeping only intentional metadata corrections

So the *data* correction for #2161's damage lands here (per-notebook `issue_pr_associee` values instead of blanket `#2161, #2564`), but the *code* that prevents future blanket overwrites is not in this diff. If `generate_catalog.py` still stamps all entries with the same metadata when run, the bug will recur on the next regen. **Worth verifying that the overwrite root cause is actually fixed upstream.**

### The `_filter_by_series` fix itself — correct

The implementation is clean:
- Exact match first, then `split("-", 1)` fallback — correct priority
- `maxsplit=1` handles multi-dash sous_series (e.g. `sous_serie="Lean4-Intro"` queried as `"SymbolicAI-Lean4-Intro"`)
- Applied consistently to all 5 filter call sites
- `test_exact_match_takes_priority` covers the ambiguity case where `serie="A-B"` coexists with `serie="A", sous_serie="B"`

### Edge case: `sous_serie: None` vs `sous_serie: ""`

The filter does strict equality: `e.get("sous_serie") == child`. In Python, `None == ""` is `False`. So:
- Query `"SymbolicAI-"` → `child = ""` → matches entries with `sous_serie: ""` ✓
- Query `"SymbolicAI-"` → does NOT match entries *without* a `sous_serie` key (returns `None`) ✗

The test fixture explicitly sets `sous_serie: ""`, so the tests pass. But in the real catalog, entries without a sous_serie may have the key absent rather than set to empty string. Compare with `compute_breakdown` which normalizes both cases: `e.get("sous_serie") or "root"`.

**Recommendation:** Normalize in `_filter_by_series` too:
```python
if e.get("sous_serie") or "" == child
```
Or ensure catalog data always sets `sous_serie` to `""` rather than omitting the key. Low severity — only triggers if querying a root sous-series where catalog entries lack the key.

### Test coverage assessment

**Covered well:**
- 8 unit tests for `_filter_by_series` (exact, sous-serie, root, nonexistent, priority)
- 3 integration tests (status block, counter, breakdown with sous-serie)

**Gaps (minor):**
- No test for `compute_maturity_distribution` / `compute_status_distribution` with sous-serie format (same code path, low risk)
- No end-to-end test: `expand_file` with a `CATALOG-STATUS` block containing `series: SymbolicAI-Lean`
- No test for the `None` sous_serie edge case described above
- No test for multi-dash sous_series (e.g. `sous_serie="Lean4-Intro"` queried as `"X-Lean4-Intro"`)

### Catalog JSON changes

The bulk update replacing `issue_pr_associee: "#2161, #2564"` with per-notebook references is expected metadata reconciliation. The few entries with updated `cells_total`, `cells_markdown`, `cells_with_outputs`, and `last_validation` (2026-06-06 → 2026-06-07) reflect notebooks that were actually modified in recent merges. One entry has a different `last_validator` (`demoklyr@users.noreply.github.com`) — consistent with an external contribution. No concerns.

### Verdict

The Serie-SousSerie filter fix is correct and well-structured. Main concern is the **#2161 overwrite root cause** — this PR corrects the *data* damage but does not contain a code fix preventing recurrence. The `None` vs `""` sous_serie edge case is minor but worth addressing for robustness. Test coverage is solid for the happy path; the gaps are low-risk but would improve confidence.

COMMENT — not APPROVE (self-review cap applies).

PR #2726, opened 2026-06-09
docs(Infer): correct algorithm description - EP, VMP, Gibbs not just message passing

jsboige:
## Summary

User feedback: the Infer README incorrectly described Infer.NET as implementing only "inference exacte par message passing". In reality it offers three compiled inference engines.

## Change

Replaced the inaccurate single-sentence description with a grounded paragraph + comparison table documenting all three algorithms:

- **Expectation Propagation (EP)** — default, fast, approximate. Good for continuous/mixed models.
- **Variational Message Passing (VMP)** — stable, underestimates uncertainty. Good for LDA/component models.
- **Gibbs Sampling** — asymptotically exact, slow. Good for validation/small models.

Grounding: table matches Infer-1-Setup.ipynb content (lines 1475-1477) which teaches these distinctions explicitly.

## Files changed (1)

| File | Change |
|------|--------|
| `Probas/Infer/README.md` | +7/-1 — corrected algorithm description + added comparison table |


jsboige (commented):
**[NanoClaw]**

Good documentation correction. The old text described Infer.NET inference as "exacte par message passing sur des graphes de facteurs" — technically wrong since EP (the default engine) is approximate, and calling all three engines just "message passing" elides the fundamental differences between them.

The new description correctly identifies the three engines and their trade-offs:
- **EP** (approximate, not exact — the key fix)
- **VMP** (different optimization objective from EP, not just another message passing scheme)
- **Gibbs** (sampling-based, not message passing at all)

The comparison table is a useful addition for learners deciding which engine to use. Technical claims check out: EP is indeed the default, VMP is preferred for latent component models like LDA, and Gibbs is the validation ground truth.

Maturity count changes across Probas/SymbolicAI READMEs and the catalog date bumps are housekeeping from #2723 — consistent and fine.

No notebooks changed, no secrets, no content loss. LGTM.

PR #2812, opened 2026-06-11
fix(qc): add set_brokerage_model to 52 strategies (#2801 Lot 1)

jsboige:
## Summary

Lot 1 of #2801 — add `set_brokerage_model` to all QC strategies missing it.

### Problem
52 out of 101 Python strategies had no `set_brokerage_model`, meaning QC default zero-cost fills applied. This makes backtests unrealistic (no commission, no slippage, no margin requirements).

### Fix
Added `self.set_brokerage_model(BrokerageName.INTERACTIVE_BROKERS_BROKERAGE, AccountType.MARGIN)` to 52 strategies that trade US equities (or equities + minor crypto allocation).

### Details
- Insertion point: after `set_cash`/`SetCash` in `initialize()`
- Style: matches existing convention (snake_case `set_brokerage_model` for newer strategies, PascalCase `SetBrokerageModel` for older ones)
- 8 strategies already had a brokerage model with variant syntax — verified and left untouched
- **Before**: 39/101 Python + 3/3 C# = 42/104 total
- **After**: 101/101 Python + 3/3 C# = 104/104 total

### Verification
- `grep -ci set_brokerage_model` confirms 0 strategies now missing it
- `grep` confirms 0 duplicate brokerage lines
- 52 files changed, 52 insertions, 0 deletions

### Impact
Strategies will now model realistic IBKR costs (~0.35 bps for small accounts). Metrics (Sharpe, CAGR, MaxDD) will be lower than before — this is **correct** behavior, not a regression.

See #2801 (Lot 1 — Brokerage models)

PR #2817, opened 2026-06-11
fix(qc): remove inline fee hacks, rely on brokerage model fees (#2801 Lot 2)

jsboige:
## Summary

Remove dead inline fee workarounds from 2 strategies, now that all strategies have explicit brokerage models (Lot 1, PR #2812) which apply realistic fees natively.

## Changes

| Strategy | Change | Rationale |
|----------|--------|-----------|
| Multi-Layer-EMA | Remove unused `self.transaction_fee = 0.001` (line 37) | Variable defined but never referenced in any order call |
| BTC-ML | Remove `TRANSACTION_FEE = 0.001` constant + `position_size * (1 - self.TRANSACTION_FEE)` sizing hack | Brokerage model Binance Cash handles fees; the position sizing workaround is incorrect |

## Why no `SetFeeModel` needed

QC applies realistic fee schedules based on the brokerage model:
- **IBKR Margin**: ~0.35bps for small accounts
- **Binance/Coinbase Cash**: ~0.1% per trade
- **OANDA**: Spread-based

For standard broker fees, no custom `SetFeeModel` is necessary. Only non-standard fee structures would require one.

See #2801

### DIFF
--- a/MyIA.AI.Notebooks/QuantConnect/projects/BTC-ML/main.py
+++ b/MyIA.AI.Notebooks/QuantConnect/projects/BTC-ML/main.py
@@ -44,9 +44,6 @@ class MyEnhancedCryptoMlAlgorithm(QCAlgorithm):
     CONFIDENCE_LONG_THRESHOLD = 0.56   # Si proba > 0.56: long (relaxed from 0.58 for more trades)
     CONFIDENCE_EXIT_THRESHOLD = 0.40   # Si proba < 0.40: liquidation (relaxed from 0.42)
 
-    # Transaction costs
-    TRANSACTION_FEE = 0.001  # 0.1% per trade (Binance fee)
-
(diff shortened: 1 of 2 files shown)


jsboige (commented):
**[Hermes] Review — remove inline fee hacks (#2817)**

**Checklist:**
1. ✅ Security: CLEAN — no hardcoded tokens/secrets in diff
2. ✅ Cross-repo impact: None (internal to CoursIA)
3. ✅ Diff quality: Clean removal of dead `transaction_fee` field and manual fee workaround

**Verdict: COMMENT — LGTM.** Straightforward dead-code cleanup. Removing the manual `TRANSACTION_FEE` deduction from `SetHoldings` is correct since QC brokerage models apply fees natively. No concerns.

PR #2915, opened 2026-06-13
feat(lean): Knots Phase 5 diagnostic — certified counter-example to tricolorable_invariant

jsboige:
## Summary

Knots Phase 5 diagnostic PR. The Phase 4 diagnosis left `tricolorable_invariant` open, attributing the blocker to a missing "transfer lemma". **This is incomplete: the statement is mathematically FALSE under the current symmetric-existential move model.** This PR adds a certified counter-example that proves it, and corrects the diagnostic to point at the real Phase 5 prerequisite (re-model the moves).

Same nature as the conway P4 breakthrough: the "intractable" target is a false statement under the current model.

## Change

**Single file: `Knots/Invariant.lean` (+108/-16).**

1. **New PROVEN theorem** `tricolorable_invariant_fails_under_current_model`:
   ```
   ∃ (d₁ d₂ : KnotDiagram),
     ReidemeisterEquiv d₁ d₂ ∧ IsTricolorable d₁ ∧ ¬ IsTricolorable d₂
   ```
   - `d₁ = trefoilDiagram` (tricolorable, proven just above as `trefoil_tricolorable`).
   - `d₂ = trefoilDiagram` with malformed crossing `⟨7,8,9,10⟩` appended, `numEdges = 8`.
   - `ReidemeisterEquiv d₁ d₂` via one R1 step (the witness `c` in `Reidemeister1` is *arbitrary* — that is the root cause).
   - `¬ IsTricolorable d₂` because the malformed crossing's `e3 = 9 > numEdges d₂ = 8` fails the well-formedness conjunct of `triColorConditionAt`.
2. **Corrected diagnostic** on the existing `tricolorable_invariant` sorry: the "transfer lemma" framing was incomplete. Real blocker = the move model is too weak; Phase 5 prerequisite is to STRENGTHEN it (geometric edge-renaming constructors + `KnotDiagram` well-formedness predicate) so malformed witnesses are excluded at the type level.

This is **pure strengthening**: a new proven theorem + a corrected comment. No existing proof is touched, none replaced by sorry.

### 1. `grep -c sorry` before/after (real declarations only)
```
grep -cE '^\s*(exact )?sorry|:=\s*by\s+sorry|:=\s*sorry' Invariant.lean
before (main fdd0f1c7): 3
after  (this branch)   : 3
```
Same 3 residual sorries as main: `tricolorable_invariant`, `trefoil_not_unknot`, `Knot.unknottingNumber`. **No regression.** The new theorem is genuinely proven.

### 2. Lake build SUCCESS
```
~/.elan/bin/lake build Knots   (Lean v4.31.0-rc1, WSL Ubuntu)
Build completed successfully (8 jobs).
```
No new errors. The 3 sorries above are the only `declaration uses sorry` warnings.

### 3. Proof integrity (axiom check)
```
#print axioms tricolorable_invariant_fails_under_current_model
> [propext, Quot.sound]
```
Lean core axioms only — no `sorry`, no custom axiom. The counter-example is a sound proof.

### 4. No prover refactor
Pure Lean, 1 file, 1 subject. Tactics: `simp`, `omega`, `refine'`, `exact`, `rfl`. No `decide`-on-opaque, no `sorry`-bridging.

## Why this matters (decision for ai-01)

ai-01 cycle 4 dispatch offered: "transfer lemma `tricolorable_invariant` (blocker cited: numEdges change by move) OU Conway P5 Gap2. Un seul item/wakeup (G.5)."

This PR establishes that the **transfer lemma is not the blocker** — the statement is false as written. Two paths forward for Phase 5 (ai-01's call):
- **(A)** Authorize re-modeling R1/R2/R3 with geometric edge-renaming + add `KnotDiagram.wf` predicate (substantial, multi-cycle, but unblocks the invariant properly).
- **(B)** Defer Knots Phase 5 and pivot to Conway P5 Gap2 (round-trip Grid↔MacroCell).

I'm not starting (A) without sign-off (it is large scope, G.4/G.5). This PR converts an under-specified blocker into a diagnosed, certified one.

See #2874

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/knot_lean/Knots/Invariant.lean
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/knot_lean/Knots/Invariant.lean
@@ -114,22 +114,57 @@ theorem tricolorable_invariant :
       ReidemeisterEquiv d₁ d₂ →
       IsTricolorable d₁ ↔ IsTricolorable d₂ := by
   exact sorry
-  -- BLOCKED (Phase 4 update): `triColorConditionAt` now implements the real Fox
-  -- condition with edge-index extraction (`colorAtNat` + `localStrands`), and
-  -- `trefoil_tricolorable` / `unknot_not_tricolorable` are re-proved under it.
-  -- The remaining blocker is the *transfer lemma*: a Reidemeister move changes
-  -- `d.numEdges` (R1 +2, R2 +4, R3 =), so `TriColoring d₁ = Fin d₁.numEdges → TriColor`
-  -- and `TriColoring d₂` have different types. Lifting a coloring across a move
-  -- requires a constructive edge-map `Fin d₂.numEdges → Fin d₁.numEdges` describing
-  -- how the move re-labels edges (the R1 twist adds a curl whose new edges inherit
-  -- the colour of the strand they sit on, etc.). The concrete moves in
-  -- Reidemeister.lean are *symmetric existentials* over PD-code surgery, not yet
-  -- carrying that edge-map — so there is no data to transfer the coloring along.
-  -- Dependency: equip Reidemeister1/2/3 with an explicit edge-renaming, then prove
-  --   (a) each move preserves the Fox condition at the affected crossings,
-  --   (b) the ≥2-colors condition is preserved (non-trivial for R1/R2: the new
-  --       edges must not collapse a 3-coloring to a 1-coloring).
-  -- Reference: Fox (1962); standard textbook proof (e.g. Adams, "The Knot Book").
+  -- BLOCKED (Phase 5 diagnosis — statement is FALSE under the current move model).
+  -- The Phase 4 diagnosis ("remaining blocker = transfer lemma") was INCOMPLETE.
+  -- The deeper blocker, established by the certified counter-example below
+  -- (`tricolorable_invariant_is_false_under_current_model`), is that the
+  -- Reidemeister moves are *too weakly specified* to make the theorem true:
+  --
+  --   `Reidemeister1 d₁ d₂ := ∃ c : PDCrossing, d₂ = {d₁ with
+  --      crossings := d₁.crossings ++ [c], numEdges := d₁.numEdges + 2 } ∨ ...`
+  --
+  -- The witness `c` is ARBITRARY — its PD labels (e1,e2,e3,e4) are unconstrained.
+  -- In particular one may pick `c` with a label OUT of range `[1, numEdges d₂]`,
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Solid diagnostic work. The certified counter-example is correct and well-motivated — it decisively upgrades the Phase 4 blocker ("missing transfer lemma") to a deeper Phase 5 finding ("the statement is false under the current model"). The proof is tight.

**Proof correctness:**

- The counter-example constructs d2 = trefoilDiagram with an appended malformed crossing (7,8,9,10) and numEdges=8, then shows (a) ReidemeisterEquiv holds via a single R1 step, (b) IsTricolorable trefoilDiagram by appeal to trefoil_tricolorable, (c) not IsTricolorable d2 by contradiction from the malformed crossing. All three goals are discharged correctly.
- The omega closure at the end works because simp only [trefoilDiagram] at hwf unfolds trefoilDiagram.numEdges to 6, making the last conjunct 9 <= 6 + 2 = 8 decidable by linear arithmetic. Clean.
- hwell := by trivial on the new diagram is consistent with the base pattern (every KnotDiagram in the codebase uses this placeholder).

**Documentation quality:**

- The expanded tricolorable_invariant comment block (Phase 5 diagnosis) is excellent: precise about what the old diagnosis missed, clear about why the statement fails, and actionable about the three prerequisites for Phase 5 (edge-renaming constructors, well-formedness predicate, restated invariant over the quotient). The references to Fox/Adams are maintained.
- The module-doc comment (the /- Certified refutation ... -/ block) and the section 3b header are both clear and correctly placed below trefoil_tricolorable.

**Minor observations (non-blocking):**

1. The `simp only [trefoilDiagram]` call is a bit optimistic — trefoilDiagram is a `def`, not a @[simp] lemma, so simp only would need trefoilDiagram to be a concrete KnotDiagram structure that simp can unfold. If this causes a simp failure in Lean 4 (simp only requires the name to be an equational lemma or a definition with equation lemmas), you may need `unfold trefoilDiagram` or `dsimp only [trefoilDiagram]` instead. Worth double-checking in the Lean environment.

2. The proof introduces a bare (7, 8, 9, 10) four times in the refine + subsequent goals. A local `def malformedCrossing := (7, 8, 9, 10)` would reduce repetition and make the counter-example pair easier to modify if the model changes.

3. The rintro for goal (c) introduces htwocolors but never uses it — the proof closes on the well-formedness conjunct alone (the >=2-colors condition is irrelevant when the crossing is malformed). A one-line comment like "htwocolors unused — malformation kills tricolorability before the >=2-colors check" would make the intent explicit.

**Verdict:** LGTM. The counter-example is a genuine contribution to the project — it narrows the search space from "find a transfer lemma" to "re-model the moves," which is exactly the kind of result that prevents wasted effort on an impossible sub-goal.

PR #2976, opened 2026-06-14
docs(search): add MetaGeneticSharp-Notebooks series README

jsboige:
## Summary

The five MGS notebooks (Introduction, Composition, Eukaryote, Islands, Benchmarks) had **no series-level entry README**. This adds pedagogical prose per notebook (concept, key primitives, learning outcome) plus the arc narrative.

### Gap
`MyIA.AI.Notebooks/Search/MetaGeneticSharp-Notebooks/` contained only the 5 `.ipynb` files — no README introducing the series. The Part4-Metaheuristics README covers high-level positioning; this README covers the **notebook-by-notebook learning path** (complementary, not duplicative).

### Content
- Arc narrative: from applying ONE metaheuristic → composing many → structuring the population → honest benchmarking.
- Per-notebook deep prose (5): what each teaches, which primitives it introduces.
- Prerequisites & execution (.NET 9, dotnet-interactive, fork DLL paths).
- Conventions: `DoubleArrayChromosome` with randomizing `CreateNew()` (initial diversity for benchmarks), fitness-as-negation pinned by the fork's [KnownFunctionsTests](https://github.com/jsboige/MetaGeneticSharp/blob/main/tests/MetaGeneticSharp.Extensions.Tests/KnownFunctionsTests.cs).

## Scope
Pure docs (CoursIA property). The submodule fork README is **untouched** (fork property). Catalog **untouched** (automation-owned — no regeneration on a feature branch, per catalog-pr-hygiene HARD 1).

See #1203 (queue item 6, dispatched by ai-01).

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #2989, opened 2026-06-14
feat(search): MGS-6 TSP -- representation-agnostic composition on permutations

jsboige:
## Summary

Queue item 4 (Epic #1203, dispatch ai-01 \"QUEUES PROFONDES\" 15:08Z). Adds the 6th notebook in the MetaGeneticSharp series: **MGS-6 TSP** — a demonstration that the composition grammar is agnostic to the gene representation.

MGS-4 (Islands) and MGS-5 (Benchmarks) ran the composition primitives on continuous surfaces (genes = `double`). The thesis \"components over metaphors\" predicts that the composition layer — `Match`, `Container`, `Scoped`, `Islands` — makes NO assumption on the gene type, because it orchestrates `IChromosome` and never reads `Gene.Value`. This notebook verifies the prediction: the **same** `IslandMetaHeuristic` that structured continuous search now structures a permutation problem (TSP, genes = city indexes = `int`), with **zero adaptation**.

The geometric compounds (WOA, EO) DO need a `GeometricConverter<double>` — but that is because their *mechanism* (centroid, spiral, geometric operator) is arithmetic on continuous positions. That requirement lives in the compound, not in the composition layer. The separation is clean: **composition is agnostic, geometric compounds are typed**.

## Setup

- `TspFitness(20, 0, 100, 0, 100)` — 20 random cities in [0,100]², fixed seed (reproducible)
- `TspChromosome(20)` — genes = visiting order = permutation of city indexes
- Operators: `OrderedCrossover` (OX1) + `ReverseSequenceMutation` + `EliteSelection`
- Same `MetaGeneticAlgorithm` engine as MGS-5
- 100 generations, population 40

## Comparison (honest, G.9)

| Config | Final tour | vs baseline |
|--------|-----------|-------------|
| Baseline (panmictic) | **429.4** | — |
| Islands (4x10, Static) | 480.7 | -12% (worse) |

For this 20-city instance the panmictic baseline wins. Islands converge slower per generation because each subpopulation (10) is smaller — expected behavior, same No-Free-Lunch theme as MGS-5's Schwefel divergence. The point is NOT that islands win; it is that the composition grammar **transferred to permutations without any adaptation**. Win/loss is secondary; representation-agnostic composition is the finding.

3 exercises: crossover operator (PMX vs OX1), migration mode (RandomRing vs Static), island topology (granularity sweep over valid partitions).

## Validation

- Executed end-to-end via papermill (`.net-csharp` kernel): **9 code cells, `execution_count` 1-9 contiguous, 0 errors** (C.2)
- 3 exercise stubs use `Console.WriteLine(... À compléter)` / `// TODO étudiant` — no `raise NotImplementedError`/`assert False`/`1/0` (C.1)
- Notebook loads `GeneticSharp.Extensions.dll` (TspFitness/TspChromosome) on top of the MGS DLLs, via local fork checkout `c:/dev/MetaGeneticSharp` (same convention as MGS-1..5)
- 0 catalog churn (single new notebook file, no regenerated artifacts)

## What this does NOT change

- No change to any library (MetaGeneticSharp fork). This is a pedagogical notebook only.
- No submodule bump — the notebook consumes the already-merged fork `main` local checkout.
- README series table updated separately (PR #2976), to avoid conflict; NB6 row added in a follow-up.

## Note on the harness

`MetaPopulation` is order-preserving (unlike stock GeneticSharp `Population`, it never sorts by fitness), so `CurrentGeneration.BestChromosome` is `null` at the `GenerationRan` event time. The harness derives the best from `Chromosomes` by fitness (`Fitness` is always set by `Evaluate`) and reads `TspChromosome.Distance` for the tour length metric. This matches the MGS-5 harness convention.

See #1203.

### DIFF
--- a/MyIA.AI.Notebooks/Search/MetaGeneticSharp-Notebooks/MGS-6-TSP.ipynb
+++ b/MyIA.AI.Notebooks/Search/MetaGeneticSharp-Notebooks/MGS-6-TSP.ipynb
@@ -1,0 +1,868 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "bafc5427",
+   "metadata": {
+    "papermill": {
+     "duration": 0.003449,
+     "end_time": "2026-06-14T18:22:23.443937+00:00",
+     "exception": false,
+     "start_time": "2026-06-14T18:22:23.440488+00:00",
+     "status": "completed"
+    },
+    "tags": []
+   },
+   "source": [
+    "# MGS-6 : TSP combinatoire -- la grammaire de composition est agnostique à la représentation\n",
+    "\n",
+    "[← MGS-5 Benchmarks](MGS-5-Benchmarks.ipynb) | [↑ Série MGS](../Part4-Metaheuristics/README.md)\n",
+    "\n",
+    "**Question centrale.** MGS-5 a montré que les métaheuristiques composées géométriques (WOA) se reconstruisent depuis des primitives sur des surfaces **continues** (gènes en `double`). Mais la grammaire de composition (`Match`, `Container`, `Scoped`, `Islands`) vit *sous* la couche géométrique. Cette couche fait-elle une hypothèse sur le type des gènes ?\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[Hermes]** LGTM. Notebook papermill-validated (9 cells, execution_count 1-9 contiguous, 0 errors). Honest comparison table (baseline wins, islands lose — win/loss secondary, representation-agnostic composition is the finding). Exercises use `// TODO étudiant` stubs (C.1 compliant). No catalog churn. Docs-only single file, no code risk.

PR #3000, opened 2026-06-14
feat(lean): Knots Phase 5 PR2 -- forward tricolorability transfer for Reidemeister1Connected (See #2874)

jsboige:
## Summary

**PR2 of the option-C transfer-lemma series** (Epic #2874). Proves the **forward** half of tricolorability preservation under a connected R1 twist:

```lean
theorem Reidemeister1Connected.tricolorable_forward {d₁ d₂ : KnotDiagram}
    (h : Reidemeister1Connected d₁ d₂) (htri : IsTricolorable d₁) :
    IsTricolorable d₂
```

Builds on the strengthened `Reidemeister1Connected` (#2980 def + #2990 `isRenameOf` constraint) and the witness-preservation validation #2994. Where #2994 verified preservation on the concrete witness pair, this PR2 makes the argument for **arbitrary** diagrams.

## Construction (option C)

Extend `d₁`'s tricoloring `col₁` to `col₂` on `d₂`'s `numEdges+2` edges, sending **both** fresh edges `b = numEdges+1` and `c = numEdges+2` to `col₁(a)`. The three families of `d₂`-crossings each preserve the Fox condition:

- **new kink `⟨a,b,c,c⟩`** (appended): strands `(a,b,c)` all read `col₁(a)` → all-equal, Fox-trivial.
- **modified endpoint `Y'`** (via `isRenameOf`, #2990): the `a → b` rename is Fox-invisible since `col₂(b) = col₁(a)`, so `Y'` reads the same three colours as the original crossing under `col₁`.
- **unchanged crossings**: inherit `d₁`'s Fox via `col₂ = col₁` on preserved labels (`hcolF1`).

A pure list-combinatorics helper `mem_set_fwd` (forward membership for `List.set`) splits `d₂.crossings = d₁.crossings.set i Y' ++ [C]`.

## lean-merge-discipline B (4 elements)

| Element | Result |
|---------|--------|
| **Sorry delta** | `Invariant.lean` **3 → 3** (baseline: `tricolorable_invariant`, `trefoil_not_unknot`, `unknottingNumber` — all untouched). `Reidemeister.lean` **2 → 2** (untouched). New theorem + helper are sorry-free. |
| **Lake build** | `lake build Knots` SUCCESS, **321 jobs, EXIT=0** (`set -o pipefail`). Verified post-rebase onto fresh `origin/main` (f74612e1). |
[...]

PR #2994, opened 2026-06-14
feat(lean): Knots Phase 5 -- connected R1 move preserves tricolorability on the witness (See #2874)

jsboige:
## Summary

**Certified slice validating option C** (connected R1 surgery) for the Knots tricolorability track (Epic #2874): on the concrete witness pair of `reidemeister1Connected_satisfiable` (Reidemeister.lean), both `witnessD1Connected` and `witnessD2Connected` are tricolorable, so the biconditional `IsTricolorable d₁ ↔ IsTricolorable d₂` holds on the pair.

This is the **positive complement** to the PR1 counter-example `tricolorable_invariant_fails_under_pr1_model` (Invariant.lean §3b, #2938): option C's *connected* surgery does **not** share the disjoint-kink defect that refuted the append-surgery model. Where #2938 *refuted* the old model, this slice *confirms* the new one on the concrete move.

**Disjoint from #2990** (the `isRenameOf` strengthen, Reidemeister.lean). This PR touches **Invariant.lean only** and the witness theorem references no strengthened def — it compiles on `main`. Forward-references #2990 in prose only (the general transfer lemma = PR2 is research-level, gated on #2990).

## Changes (1 file, +87/-0)

- **`def witnessD1Connected` / `def witnessD2Connected`** — the concrete before/after diagrams of `reidemeister1Connected_satisfiable` (`d₁ = [⟨1,2,3,4⟩,⟨1,2,3,4⟩]`, numEdges 4 → `d₂ = [⟨1,2,3,4⟩,⟨5,2,3,4⟩,⟨1,5,6,6⟩]`, numEdges 6).
- **`theorem witnessD1Connected_tricolorable`** — both crossings read `(red, blue, green)` on Fox strands `(1,2,3)` (all-distinct).
- **`theorem witnessD2Connected_tricolorable`** — `⟨1,2,3,4⟩`/`⟨5,2,3,4⟩` read all-distinct colours; the new kink `⟨1,5,6,6⟩` reads `(red, red, red)` (all-equal, Fox-trivial). The two new edges `b=5, c=6` carry `col(a=1) = red` — the transfer construction on the witness.
- **`theorem reidemeister1Connected_witness_preserves_tricolorable`** — the biconditional, composed from the two inhabited proofs.

[...]

PR #2938, opened 2026-06-14
feat(lean): Knots Phase 5 — certified counter-example: tricolorable_invariant STILL false under PR1 model (See #2874)

jsboige:
## Summary

**G.9 (culture du doute) finding: the PR1 move model is STILL too coarse to make `tricolorable_invariant` true.** This PR delivers a **certified counter-example** (a proven theorem, not a gap) that refutes the biconditional under the current `Reidemeister1/2/3` definitions merged in PR1 (#2929).

This supersedes the planned "PR2 = transfer lemma": before attempting the transfer lemma, I verified (G.9) the statement is *provable* — it is not, so a transfer-lemma proof attempt would have wasted cycles on a false statement. Pattern: "intractable = false statement" (cf. conway P4 / Knots #2915, `feedback-lean-false-statement-counterexample`).

### Why the PR1 model is still too coarse

PR1 strengthened the moves with `KnotDiagram.wf` on both diagrams + a free edge-renaming `ρ : Fin(min) ↪ Fin(max)`. That removed the Phase-3 malformed-witness obstruction. But:

- `wf`'s "every label appears exactly twice" forces an R1 twist's new crossing `c` to use **only the 2 fresh edges** `{n+1, n+2}` (labels `1..n` already appear twice in `d₁`, so `c` cannot reuse any without breaking parity);
- `ρ` is a **free injection**, NOT tied to `c`'s labels.

So the new crossing's Fox condition involves only the freely-colorable fresh edges and is **decoupled from `d₁`'s coloring** — a twist can CREATE tricolorability from nothing, or hide the ≥2-colours entirely in the fresh edges while `d₁` is forced monochrome.

### Witness (proven)

```
theorem tricolorable_invariant_fails_under_pr1_model :
    ∃ (d₁ d₂ : KnotDiagram),
      ReidemeisterEquiv d₁ d₂ ∧ ¬ IsTricolorable d₁ ∧ IsTricolorable d₂
```

- `d₁ = {[⟨1,2,1,2⟩], numEdges 2}` — NOT tricolorable (Fox at the sole crossing forces the 2 edges equal, contradicting ≥2-colours).
- `d₂ = {[⟨1,2,1,2⟩, ⟨3,4,3,4⟩], numEdges 4}` — tricolorable (edges 1,2 red; edges 3,4 blue; Fox all-equal per crossing).
- Connected by a single R1 twist step (`ReidemeisterEquiv.step`).

### Phase 5 prerequisite (the real fix, deferred)

[...]

PR #3002, opened 2026-06-14
feat(lean): Knots Phase 5 -- certified counter-example: backward transfer FAILS under Reidemeister1Connected (See #2874)

jsboige:
## Summary

**Certified counter-example: the BACKWARD tricolorability transfer is FALSE under `Reidemeister1Connected`.** A connected kink (option C surgery) can CREATE tricolorability. This is the **third certified obstruction** in the Phase-5 model series (after the append-model §3b, and the free-ρ vacuous `Reidemeister1'`), and the analog of #2938 for the connected model.

The backward direction `Reidemeister1Connected d₁ d₂ → IsTricolorable d₂ → IsTricolorable d₁` — needed (with forward #3000) to discharge the marquee `tricolorable_invariant` — does **not** hold under the current def.

## The counter-example (machine-verified, then formal)

| diagram | PD | numEdges | tricolorable? |
|---------|----|---------:|---------------|
| `d₁` | `[⟨1,1,2,2⟩]` | 2 | **NO** — Fox strands `(1,1,2)` force `col1 = col2`, contradicting ≥2 colours |
| `d₂` | `[⟨3,1,2,2⟩, ⟨1,3,4,4⟩]` | 4 | **YES** — witness `⟨R,G,B,G⟩`: strands `(3,1,2)=(B,R,G)` and `(1,3,4)=(R,B,G)` each all-distinct |

`Reidemeister1Connected d₁ d₂` holds (verified by the proof, mirroring `reidemeister1Connected_satisfiable`): twist on arc `a = 1` (rename `e1: 1 → 3`), append kink `⟨1,3,4,4⟩`. Yet `IsTricolorable d₂` and `¬ IsTricolorable d₁`.

## Root cause + what is NOT broken

- **Root cause:** `Reidemeister1Connected` allows the twist on arc `a` when `a` is a **monogon-loop** (`e1 = e2` within a single crossing — here `⟨1,1,2,2⟩`). That is not a genuine Reidemeister-1 curl on a proper arc (which spans two distinct crossings). The def is too permissive, in the same family as the append-model defect.
- **Forward (#3000) is UNAFFECTED** — `Reidemeister1Connected.tricolorable_forward` is unconditional (`d₁` tri ⟹ `d₂` tri holds, vacuously here where `d₁` is not tricolorable). The forward PR stands on its own.
- **Likely fix (coordinator's call, out of scope here):** (a) require `a` to span two distinct crossings (a proper arc), or (b) strengthen `KnotDiagram.wf` to exclude standalone double-monogon diagrams. Either tightening is expected to restore backward without abandoning option C — parallelling how PR1's `wf` tightening addressed the append-model defect.

Certified beforehand by an exhaustive brute-force search (all `3^n` colourings) confirming `d₂` tricolorable with a valid colouring where `col(a) ≠ col(b)` — the obstruction to any naive-restriction proof of backward.

## lean-merge-discipline B (4 elements)

| Element | Result |
|---------|--------|
| **Sorry delta** | `Invariant.lean` **3 → 3**, `Reidemeister.lean` **2 → 2** (0 regression; new theorem sorry-free). |
| **Lake build** | `lake build Knots` SUCCESS (**321 jobs, EXIT=0**, `set -o pipefail`). |
| **Axiom check** | `#print axioms Knots.tricolorable_backward_fails_under_connected_model` = `[propext, Quot.sound]` (⊆ baseline, no `sorryAx`). |
| **Scope diff** | `+96/-0`, 1 file (`Invariant.lean` §3e), 0 catalog drift, additive (1 new theorem). |

## Scope guards

- `Reidemeister.lean` (incl. `Reidemeister1Connected`, `isRenameOf`, API lemmas) — **untouched**.
- `Basic.lean`, `Conway.lean`, `Lidman.lean`, `MathlibPrerequisites.lean` — **untouched**.
- `tricolorable_invariant` (Invariant.lean) — **untouched** (still sorry; this counter-example explains WHY backward is hard and pins the def-level root cause).
- Disjoint from #3000 (forward, open) and #2994 (witness-preservation, merged): `Invariant.lean` only, compiles on fresh `main`.
- No catalog regeneration on the branch (rule catalog-pr-hygiene).
- `See #2874` (epic open — a diagnostic/obstruction theorem, not a closer).

See #2874

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/knot_lean/Knots/Invariant.lean
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/knot_lean/Knots/Invariant.lean
@@ -440,6 +440,102 @@ theorem reidemeister1Connected_witness_preserves_tricolorable :
     IsTricolorable witnessD1Connected ↔ IsTricolorable witnessD2Connected :=
   ⟨fun _ => witnessD2Connected_tricolorable, fun _ => witnessD1Connected_tricolorable⟩
 
+/-! ## 3e. Certified counter-example: the BACKWARD transfer FAILS under
+`Reidemeister1Connected` (a connected kink can CREATE tricolorability)
+
+§3d shows the *concrete witness pair* of `reidemeister1Connected_satisfiable`
+preserves tricolorability. This section shows the GENERAL backward direction
+`Reidemeister1Connected d₁ d₂ → IsTricolorable d₂ → IsTricolorable d₁` is still
+FALSE — the connected-surgery model (option C) admits its OWN permissiveness defect,
+distinct from the append-model defect of §3b. This is the third certified obstruction
+in the Phase-5 model series (append-model §3b, free-ρ vacuous Reidemeister1', and now
+connected-kink-creates-tricolorability).
+
+**Counter-example.** `d₁ = ⟨1,1,2,2⟩` (1-crossing double-monogon, `numEdges = 2`) is NOT
+tricolorable: at its sole crossing the Fox strands are `(1,1,2)`, so `c1 = c2`
+(both read label 1), forcing the all-equal branch `col 1 = col 2`, contradicting the
+≥2-colours condition. The connected twist on arc `a = 1` (which appears at `e1` AND
+`e2` of `d₁` — a degenerate monogon-loop arc, not a genuine R1 arc spanning two
+distinct crossings) renames `e1 : 1 → 3` and appends the kink `⟨1,3,4,4⟩`, yielding
+`d₂ = {[⟨3,1,2,2⟩, ⟨1,3,4,4⟩], 4}`, which IS tricolorable (witness `⟨R,G,B,G⟩`:
+strands `(3,1,2) = (B,R,G)` and `(1,3,4) = (R,B,G)` are each all-distinct). So a
+connected kink creates tricolorability and `tricolorable_invariant` is FALSE under
+`Reidemeister1Connected`.
+
+**Root cause.** `Reidemeister1Connected` allows the twist on arc `a` even when `a` is
+a monogon-loop (`e1 = e2` within one crossing), which is not a genuine Reidemeister-1
+curl on a proper arc. The forward direction (#3000,
+`Reidemeister1Connected.tricolorable_forward`) is UNAFFECTED — it is unconditional
+(`d₁` tri ⟹ `d₂` tri holds, vacuously on this pair where `d₁` is not tricolorable).
+
+**Fix direction (out of scope here — coordinator's call).** Either (a) require `a` to
+span two distinct crossings (a proper arc, excluding monogon-loops), or (b) strengthen
+`KnotDiagram.wf` to exclude standalone double-monogon diagrams. Either tightening is
+expected to restore the backward direction without abandoning option C — parallelling
+how PR1's `wf` tightening addressed the append-model defect. Certified by exhaustive
+brute-force search (all 3^n colourings) before this formal proof.
+-/
+
+theorem tricolorable_backward_fails_under_connected_model :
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[NanoClaw]**

## Review: PR #3002 -- Certified counter-example for backward transfer under Reidemeister1Connected

### Overall assessment

This is a well-constructed certified counter-example in Lean 4 that formally proves the backward direction of tricolorability transfer fails under `Reidemeister1Connected`. The proof structure follows established patterns from earlier sections (sections 3b, 3d) and the `sorry` discipline is properly maintained. The mathematical argument is sound.

### Logical soundness

The counter-example is correct. The key insight is that `d1 = [⟨1,1,2,2⟩]` is a double-monogon where Fox strands `(1,1,2)` force `col(1) = col(2)` (since both `c1` and `c2` read label 1), making it non-tricolorable. The connected twist on arc `a=1` (a degenerate monogon-loop) produces `d2 = [⟨3,1,2,2⟩, ⟨1,3,4,4⟩]` with a valid 3-coloring witness `(R,G,B,G)` where all strands at both crossings are all-distinct. The root cause analysis -- that `Reidemeister1Connected` permits twists on monogon-loop arcs rather than proper arcs spanning two distinct crossings -- is precisely identified.

Three proof obligations are discharged:
1. **Reidemeister1Connected d1 d2**: Constructed by mirroring `reidemeister1Connected_satisfiable` with `i = ⟨0⟩`, `a = 1`, and the appropriate embedding `rho : Fin 2 ↪ Fin 4`. The embedding `toFun` with `omega` bounds and the rename/kink construction all look correct.
2. **IsTricolorable d2**: The explicit witness `⟨R,G,B,G⟩` is verified by case-splitting on the two crossings, each confirmed via `decide`. The `>= 2 colours` condition uses `red != green`.
3. **¬ IsTricolorable d1**: Clean proof. Shows Fox forces `coloring(0) = coloring(1)` via case analysis on the all-equal/all-distinct branches (the all-distinct branch is contradictory since `c1 = c2` share label 1). Then all values are equal, contradicting the `>= 2 colours` condition. This mirrors the negation pattern from section 3b as stated.

### Code quality and structure

- **Documentation**: The module doc-comment (/-! ... -/) is thorough -- counter-example construction, root cause, forward unaffected, fix direction, and brute-force certification are all documented. This is good practice for formal proofs.
- **Naming**: `tricolorable_backward_fails_under_connected_model` is descriptive and consistent with the naming convention in section 3d.
- **Proof style**: Appropriate use of `refine`, `by decide`, `by omega`, and explicit case splits. The monogon-loop observation in comment lines clarifies why the arc `a=1` is degenerate.
- **Axiom check**: The author reports `#print axioms` yields only `[propext, Quot.sound]`, which is within the baseline. No `sorryAx` introduced.

### Minor observations (non-blocking)

1. The `hsplit` in obligation (b) uses `simpa` to derive the disjunction from list membership -- this works because the list has exactly two elements. If the list grew, this would need adjustment, but for a fixed 2-element witness it is fine.
2. The `coloring` match in (b) uses a default case `| _ => TriColor.green` for `i.val` beyond 3. Since `Fin 4` has only 4 values (0-3), this is exhaustive but the wildcard pattern is slightly less explicit than a `| 3 => TriColor.green` case. Not a defect, just a style preference.
3. The `intro c hc` / `have hsplit` pattern in (b) is a reasonable approach given that `decide` alone may not handle the unification on the structural equality of `PDCrossing` values inside the list membership proof.

### Security

No secrets, credentials, or sensitive tokens found in the diff.

### Verdict

The counter-example is mathematically sound, the formalization follows the project's established patterns, the `sorry` discipline is maintained, and the root cause analysis is precise. The additive nature (+96/-0) with zero catalog drift is consistent with the stated scope. LGTM for a counter-example / diagnostic theorem.

PR #2998, opened 2026-06-14
feat(genai): wake-on-demand helper for idle-stopped services (#2982)

jsboige:
## What

Wake-on-demand helper for idle-stopped GenAI services — closes the gap identified in #2982.

`docker-configurations/services/shared/service_idle_monitor.py` auto-**STOPs** services after 20 min idle to free GPU VRAM (valuable, kept). But it has no auto-**START**: a notebook hitting a stopped service gets HTTP 502 / connection-refused until a manual `docker start` + warm-up.

This PR adds the demand-side counterpart: **`service_wake.py`**.

## Decision on approach (criterion #3 of #2982)

Chose **client-side wake-and-retry** over the 4 options in the issue:

| Approach | Verdict |
|----------|---------|
| A. Reverse-proxy wake-on-request | **Rejected** — over-engineering for a dev/teaching stack (persistent process + routing config per service) |
| B. `idle_timeout` tuning | **Rejected** — reduces 502 frequency, doesn't eliminate it; consumes VRAM longer |
| C. Always-on | **Rejected** — defeats VRAM economy; OOM risk on shared GPU (Higgs GPU1-exclusive) |
| **D. Wake helper (this PR)** | **Chosen** — low-blast, additive, uniform across all 16 services |

**Key finding (grep-verified)**: none of the 16 services configure `idle_action=sleep` — they all do `docker stop`. A stopped container has **no listener** on its port, so wake-on-request *must* originate from the caller (or a persistent proxy). A caller-side helper is the minimal, uniform fix; it works for every stopped service without per-service `/wake_up` routes.

**VRAM economy preserved**: services still auto-stop after 20 min idle. The helper only `docker start`s on actual demand.

## Usage

```python
from service_wake import ensure_service_up
if ensure_service_up("musicgen-api", "http://localhost:8192/health"):
    # safe to POST http://localhost:8192/v1/generate
    ...
```

```bash
python service_wake.py musicgen-api --health-url http://localhost:8192/health -v
# OK: musicgen-api is up at http://localhost:8192/health  (exit 0)
```

## Logic

`ensure_service_up(container, health_url, api_key=None, timeout=180)`:
1. **Fast-path**: `GET /health` → 200 means already active → return immediately (no docker call).
2. **Cold path**: `docker start <container>` (idempotent), then poll `/health` every 2s until 200 or `timeout`.
3. Returns `True` if healthy in time, `False` otherwise (caller decides fallback — no exception).

## Validation (H.1)

[...]

PR #3007, opened 2026-06-14
test(genai): lock api_key propagation contract in service_wake (#2982)

jsboige:
## What

Pins the `api_key` propagation contract in `service_wake.py` (delivered #2998, merged). Two targeted unit tests that close a genuine coverage gap on a security-relevant helper.

## Gap

The 14 tests added with the helper cover `probe_health`'s Bearer-header wiring directly (`test_probe_health_sends_bearer_when_api_key`), but **none** assert that `ensure_service_up` forwards its `api_key` param through to `probe_health` — not on the fast-path, not on the cold-path polls. A future refactor could silently drop the arg.

**Why it matters (#16 security context):** with an auth-active service, a dropped `api_key` makes `probe_health` get HTTP 401 → the helper misreads the service as "down" → it triggers a spurious `docker start`. The wake helper must carry auth through every probe.

## Change

Two tests in `test_service_wake.py` (additive +40/-0):

- `test_ensure_service_up_fast_path_forwards_api_key` — fast-path hit, asserts the single probe received `api_key="secret"`.
- `test_ensure_service_up_cold_path_forwards_api_key` — cold path, asserts EVERY probe (fast-path down + post-start polls) received `api_key="secret"`, not just the first.

## Validation

```
16 passed in 0.23s   (14 existing + 2 new)
```

Plus a **real end-to-end cold-start** validated this cycle on po-2023 (not part of the PR — CI-runnable tests only — but confirms the cold path works in real conditions, previously mock-only):

- `demucs-api` idle-stopped (Exited 137, 6h) → `GET /health` connection-refused (the #2982 502 symptom).
- `python service_wake.py demucs-api --health-url http://localhost:8193/health -v` → `docker start` → **healthy in 7s**.
- `GET /v1/models` no-key → **401** (auth active) / with-key → **200** (model loaded).

## Scope

- Additive +40/-0, 0 deletions (anti-régression safe).
- CI-runnable (no docker/network dependency — pure unit, mocked).
- Catalogue byte-identical.

See #2982

### DIFF
--- a/docker-configurations/services/shared/test_service_wake.py
+++ b/docker-configurations/services/shared/test_service_wake.py
@@ -116,6 +116,46 @@ def test_ensure_service_up_docker_start_fails():
         assert service_wake.ensure_service_up("svc", "http://h/health") is False
 
 
+def test_ensure_service_up_fast_path_forwards_api_key():
+    """Contrat sécurité (#16) : api_key doit atteindre probe_health même au
+    fast-path — sinon un service auth-actif renvoie 401 et le helper croit le
+    service down → déclenche un docker start inutile."""
+    received_keys = []
+
+    def fake_probe(url, api_key=None, timeout=5):
+        received_keys.append(api_key)
+        return True  # fast-path hit
+
+    with patch("service_wake.probe_health", side_effect=fake_probe):
+        assert service_wake.ensure_service_up(
+            "svc", "http://h/health", api_key="secret"
+        ) is True
+        assert received_keys == ["secret"], (
+            "api_key must propagate to probe_health on the fast-path"
+        )
+
+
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[Hermes]** COMMENT (self-review cap) — LGTM.

**Checklist:**
- ✅ Security: grep clean (mock value "secret" is test-only, not a credential)
- ✅ Coverage gap: genuine — fast-path and cold-path api_key forwarding was untested
- ✅ Test quality: proper mock via patch, assertions on exact key list, good docstrings explaining #16 context
- ✅ Scope: +40/−0, 1 file, additive-only

No concerns. Solid security-relevant contract test.

PR #3023, opened 2026-06-15
fix(qc-mcp-lite): surface QC API success:false + list_projects filter (See #1630)

jsboige:
## Summary

Two bugs in qc-mcp-lite, root-caused during the #1630 post-#2801 verification campaign (batch 5) and verified against the test fixtures.

### Bug 1 — `create_backtest` returned an empty `backtestId` (silent failure)
During batch 5, `create_backtest` returned `{"backtestId": "", "status": ""}` three times in a row for RL-DQN-Trading / HAR-RV-Kelly, forcing blind recompile+retry with no clue why.

**Root cause:** `_api_post` called `requests.raise_for_status()` — HTTP-level only. But **QC v2 always returns HTTP 200**; API-level failures are signalled in the body as `{"success": false, "errors": [...]}` (confirmed by the existing test fixture `{"success": True}` and the compile-error fixture `{"state": "BuildError", "errors": [...]}`). The wrapper never checked `success`, so a rejected call fell through silently and `create_backtest` read `backtestId` from a failure body → empty. **The actual QC rejection reason was dropped.**

**Fix:** `_api_post` now raises `RuntimeError` with the QC `errors` when `success is False`. Guarded with `is False` so `success:true` and legacy endpoints without the key pass through untouched. Fixes **all** tools at once (they share `_api_post`).

### Bug 2 — `list_projects` capped at 30, no way to reach the other 152
The account has 182 projects; the wrapper hard-capped `[:30]` client-side. This blocked batch 6 (strategy project IDs past index 30 were unreachable). The full list is already fetched in one call — the cap is purely for token budget.

**Fix:** added a `name_contains` filter (case-insensitive, applied **before** the cap) + a `total` field (pre-filter count). Backward compatible — no-arg calls behave exactly as before.

## Validation
- **47/47 tests pass** (41 existing + 6 new, zero regressions — backward compat proven by the existing suite still green)
- 6 new tests cover both fixes, including a named regression test (`test_surfaces_qc_rejection_not_empty_id`) documenting the empty-id symptom
- The 3 empty backtestIds from batch 5 ARE the real-API confirmation that QC returned `success:false` for those creates (a successful create always includes the id)

## Files
- `scripts/qc-mcp-lite/server.py` (+24/-3) — `_api_post` guard + `list_projects` filter
- `scripts/qc-mcp-lite/tests/test_server.py` (+69) — 6 new tests

The live MCP server picks up the fix on next restart.

See #1630

### DIFF
--- a/scripts/qc-mcp-lite/server.py
+++ b/scripts/qc-mcp-lite/server.py
@@ -77,7 +77,17 @@ def _api_post(path: str, data: Optional[dict] = None) -> dict:
         timeout=120,
     )
     resp.raise_for_status()
-    return resp.json()
+    payload = resp.json()
+    # QC v2 always returns HTTP 200, even on API-level failures: the body carries
+    # {"success": false, "errors": [...]}. raise_for_status() does NOT catch these.
+    # Without this guard, a rejected call (e.g. backtest create on a not-ready
+    # compile, or a duplicate name) flows through and create_backtest returns an
+    # empty backtestId with NO error text — forcing blind retries. Surface it.
+    if isinstance(payload, dict) and payload.get("success") is False:
+        errors = payload.get("errors") or []
+        msg = "; ".join(str(e) for e in errors) or "unknown error"
+        raise RuntimeError(f"QC API {path} rejected the request: {msg[:1000]}")
+    return payload
 
 
 def _extract_stats(bt: dict) -> dict:
@@ -206,12 +216,23 @@ def list_backtests(project_id: int) -> dict:
 
 
 @mcp.tool()
-def list_projects() -> dict:
-    """List all QC projects."""
+def list_projects(name_contains: str = "") -> dict:
+    """List all QC projects.
+
+    The full list is fetched in one call but capped (client-side) to keep the
+    response under the schema token budget. Pass ``name_contains`` to filter
+    case-insensitively BEFORE the cap — needed to locate project IDs by strategy
+    name when the account has many projects (the cap otherwise hides them).
+    """
     data = _api_post("/projects/read")
     projects = data.get("projects", [])
+    total = len(projects)
+    if name_contains:
+        nc = name_contains.lower()
+        projects = [p for p in projects if nc in (p.get("name", "") or "").lower()]
     return {
         "count": len(projects),
+        "total": total,
         "projects": [
             {
                 "projectId": p.get("projectId", 0),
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Verified both fixes against the PR branch — clean, correct, backward compatible.

**Bug 1 ( guard)**
-  +  — identity check, not truthiness, so , , or missing key all pass through correctly.
-  handles both missing and empty error lists.
-  cast + 1000-char truncation are safe.
- Fixing at the shared  layer means all downstream tools (, , etc.) benefit automatically — right architectural choice.

**Bug 2 ( filter)**
- Confirmed the cap  is still applied on line 244, after the filter — no regression on the existing .
-  field (pre-filter count) is a nice addition for callers to know they might be missing results.
- Case-insensitive matching handles None names via .
- One note: if >30 projects match the filter, callers still lose results. The docstring documents this ("filter BEFORE the cap") — acceptable for now, but worth a page/token cursor parameter if the 182-project account grows further.

**Tests — 6 new, all meaningful**
- Covers the three  states (false, true, missing key) plus the create_backtest regression, total-vs-count semantics, and case-insensitivity.
- Named regression test  directly documents the symptom from batch 5.
- Existing suite green confirms backward compat.

No security concerns, no edge-case gaps. Ship it.

clusterManager-Myia (commented):
**[NanoClaw]**

Verified both fixes against the PR branch -- clean, correct, backward compatible.

**Bug 1 (_api_post guard)**
- `isinstance(payload, dict)` + `payload.get("success") is False` -- identity check, not truthiness, so `success: null`, `0`, or missing key all pass through correctly.
- `payload.get("errors") or []` handles both missing and empty error lists.
- `str(e)` cast + 1000-char truncation are safe.
- Fixing at the shared `_api_post` layer means all downstream tools (`create_backtest`, `create_compile`, etc.) benefit automatically -- right architectural choice.

**Bug 2 (list_projects filter)**
- Confirmed the cap `projects[:30]` is still applied on line 244, after the filter -- no regression on the existing `test_limits_to_30`.
- `total` field (pre-filter count) is a nice addition for callers to know they might be missing results.
- Case-insensitive matching handles None names via `(p.get("name", "") or "").lower()`.
- One note: if >30 projects match the filter, callers still lose results. The docstring documents this ("filter BEFORE the cap") -- acceptable for now, but worth a page/token cursor parameter if the 182-project account grows further.

**Tests -- 6 new, all meaningful**
- Covers the three `success` states (false, true, missing key) plus the create_backtest regression, total-vs-count semantics, and case-insensitivity.
- Named regression test `test_surfaces_qc_rejection_not_empty_id` directly documents the symptom from batch 5.
- Existing suite green confirms backward compat.

No security concerns, no edge-case gaps. Ship it.

PR #3069, opened 2026-06-15
chore(mgs): bump MetaGeneticSharp submodule to c1bc8af (M2 Image-leak fix)

jsboige:
## M2 — CoursIA coupling (submodule bump only)

Couples fork PR [jsboige/MetaGeneticSharp#23](https://github.com/jsboige/MetaGeneticSharp/pull/23) (merged, squash `c1bc8af`).

### What M2 fixes (in the fork)
`LandscapeMaps.CreateFunction(KnownHeightMap)` loaded a height map via `Load(map)` (`Image.FromStream`, caller-owned) and assigned it to the `ImageHeightMapFunction.TargetImage` setter, which **grayscale-copies** it internally (jsboige's `MakeGrayscaleImage()` @ `d05826fd`). The loaded source was **never disposed** → GDI+ handle leak. Fix = `using Image source = Load(map);` (mirrors the existing `CreateFunctionFromFile` sibling). +5 tests, 111/111 green.

### Why submodule-bump only (no notebook change)
M2 is a **lib-only memory-hygiene fix** — it changes no rendering output. MGS-7's source cells and Papermill outputs are therefore **untouched** (C.3: no notebook re-exec without a source change). This PR is a single gitlink advance `ed013e1` → `c1bc8af` to keep the submodule pointer in sync with fork main (provenance), 0 catalogue/README churn.

### Stratum / authorship
Authored-wrapper stratum (`LandscapeRenderer.cs`) only. Byte-exact originals (`ImageHeightMapFunction`/`DirectBitmap`/`ImageExtensions`) and grayscale math are jsboige's @ `d05826fd` and untouched. NO-PENDULUM (additive, no substitution).

See #1203. Credit: jsboige @ `d05826fd`.

PR #3071, opened 2026-06-15
chore(mgs): bump submodule to L1 parallel RenderHeatmap (800c74d)

jsboige:
## L1 — coupled CoursIA bump (deep-queue, ai-01 Priority-0)

Advances the **MetaGeneticSharp** fork submodule `c1bc8af` → `800c74d`, pulling in the **L1** deep-queue item (fork PR [jsboige/MetaGeneticSharp#24](https://github.com/jsboige/MetaGeneticSharp/pull/24)).

### What L1 does
`LandscapeRenderer.RenderHeatmap` now evaluates the `width*height` fitness grid (**~120k samples** at 400×300) with `Parallel.For` over the canvas rows + per-thread min/max reduction, instead of a single sequential loop. The reduction keeps the **lowest index** on equal extrema, so the rendered bitmap is **byte-identical** to the prior sequential render — a pure performance change, no output change.

- Authored-wrapper stratum only; jsboige's color ramp / extrema markers / sampling math @ `d05826fd` are untouched.
- Fork suite **114/114** (111 → 114): pixel-for-pixel equivalence vs an independent sequential reference, tied-extrema lowest-index marker, cross-run determinism.

### Why submodule-bump-only (no notebook re-exec)
L1 is a lib-only performance change with **no rendering output change**. Per rule **C.3**, no notebook source cell is modified and no Papermill re-execution is staged — exactly the M2 #3069 pattern. **Single gitlink**, **0 catalogue churn**, **0 README churn**.

`See #1203`.


jsboige (commented):
**[NanoClaw]** Submodule bump to `800c74d` (L1 parallel RenderHeatmap). Trivial. LGTM.

Issue #1460, opened 2026-05-23
fix(prover): add progress-hash loop detection — 170 exact-repeat tool calls per analysis window


jsboige:
## Problem

Forensic analysis found **170 exact-repeat tool calls** (3+ identical consecutive tool_name+args) and **20 semantic loops** (8+ same-tool runs). The prover burns iterations re-reading context or re-searching lemmas without making edits.

Worst cases:
- 24 consecutive `file_read_lines` without intervening `file_replace_*`
- 24 consecutive `search_mathlib_lemmas` without applying results
- 54 exact-repeat `compile_probe_goal` calls (same unchanged goal)

### 1. Progress hash (in tools.py)

After each `file_replace_*` or `file_insert_lines`, compute a hash of file content + goal state. If hash unchanged after 3 write-compile cycles → trigger strategy change.

### 2. Stale context detector

Track `(tool_name, args_hash)` pairs per session. If same pair seen 3+ times without an intervening `file_replace_*`:
- `file_read_lines`: inject warning + suggest switching to `search_mathlib_lemmas`
- `search_mathlib_lemmas`: inject warning + cap at 5 consecutive, then require `compile_probe_goal` or `file_replace_*`
- `compile_probe_goal`: cache result keyed by file content hash + goal line (TTL 60s)

### 3. Semantic loop cap

If 8+ consecutive same-tool calls detected by the orchestrator (provers.py), force route to next agent or terminate iteration as stalled.

### Files affected
- `tools.py`: add progress hash computation, cache for compile_probe_goal
- `provers.py`: add stale context check between agent invocations
- `workflow.py`: add semantic loop cap in executor chain

### Expected impact

- Reduce wasted iterations by ~50% (170 exact loops → near-zero)
- Save ~2.2h of compile_probe_goal redundant calls via caching
- Faster detection of intractable goals (3-cycle hash check vs 15+ intractable marks)

## Evidence

Full forensic report: `baselines/forensic_analysis.md` sections 3 and 8.

Part of Epic #1453

jsboige:
Delivered in #1552 (merged 2026-05-24): compile_probe_goal now caches by (file content md5, sorry_line). Content-hash key auto-invalidates on any edit — stronger than the suggested 60s TTL (no wall-clock nondeterminism, deterministic tests). Guard suite 54→56 passing.

PR #3090, opened 2026-06-15
test(prover): characterize P2 probe-goal cache window (#1460)

jsboige:
## Context

Follow-up to #3087, where I flagged a pre-existing failing test: `test_probe_goal_cache_invalidates_on_file_change`. This PR fixes it.

## Root cause

The probe-goal cache (`compile_probe_goal`, `tools.py:1711`) was narrowed in the **P2 fix (#1460)** to key on the surrounding-context hash (`lines[sorry-11 : sorry+10]`) instead of the whole file -- so edits far from the sorry no longer invalidate the cache (deliberate optimization, documented in the method docstring at `tools.py:1714-1718`).

The old test still asserted the **pre-P2 whole-file-hash behavior**: it appended `\n-- edit\n` at EOF and expected a recompute (`calls["n"] == 2`). But the padded fixture puts the sorry near the middle (`sorry_line = 54`) with ~50 trailing lines, so EOF (~line 106) is far outside the `[44, 64]` context window -> the cache correctly hit -> `calls["n"] == 1` -> `assert 1 == 2` failed.

## Fix

Replaced the stale test with two **characterization** tests of the P2 behavior:
- `test_probe_goal_cache_invalidates_on_context_change` -- an **in-window** edit (the line right after sorry, index within `[sorry-11, sorry+10]`) changes the context hash -> cache miss -> recompute (`calls == 2`).
- `test_probe_goal_cache_survives_out_of_window_edit` -- an **EOF append** stays outside the window -> cache hit -> no recompute (`calls == 1`). This is the exact case the old assertion wrongly expected to recompute.

## Validation (G.1)

- **Reproduced** the failure on clean `origin/main` first: `assert 1 == 2`.
- Suite: clean main = 71 passed / 1 failed; this branch = **73 passed / 0 failed**.
- `test_prover_guards.py` is **not exercised by CI** (no workflow references it), so this is local test-suite hygiene, not a CI red -- but a locally-failing test is a real defect (confusing red bar for anyone running `pytest`).

## Scope

- 1 file, +47/-4, test-only. No production code touched (the cache logic is unchanged and correct).
- Unrelated working-tree drift (`Reidemeister.lean`, submodule pointer, scratch `tmp_*.py`) is NOT staged.

See #1460

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/tests/test_prover_guards.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/tests/test_prover_guards.py
@@ -992,8 +992,15 @@ def test_probe_goal_caches_unchanged_file(tactic_tools, monkeypatch):
     assert json.loads(first)["goal"] == f"goal at {line}"
 
 
-def test_probe_goal_cache_invalidates_on_file_change(tactic_tools, monkeypatch):
-    """Editing the file changes its content hash, so the next probe recomputes."""
+def test_probe_goal_cache_invalidates_on_context_change(tactic_tools, monkeypatch):
+    """An edit INSIDE the [sorry-11, sorry+10] context window busts the cache.
+
+    P2 fix (#1460): the probe cache keys on the surrounding context hash
+    (±probe_ctx_lines around sorry), not the whole file. A context-window edit
+    changes that hash -> cache miss -> recompute. (The old whole-file-hash
+    test assumed ANY edit invalidated; the P2 narrowing makes that false -- see
+    the companion test_probe_goal_cache_survives_out_of_window_edit below.)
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[Hermes]** LGTM — clean characterization of the P2 probe-goal cache window.

**Checklist:**
- ✅ Security: no secrets/hardcoded tokens
- ✅ Test-only: no production code touched, 1 file, +47/-4
- ✅ Well-structured: two complementary tests (in-window bust vs out-of-window survive) correctly mirror the P2 hash narrowing
- ✅ Root cause clearly documented (pre-P2 whole-file-hash assumption stale after #1460)
- ✅ 71→73 passed, 0 failed — existing failure resolved

**COMMENT**

PR #3122, opened 2026-06-16
fix(arg-analysis): add stub-cell output to 0-init for C.2 (See #2137)

jsboige:
-

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Argument_Analysis/Argument_Analysis_Agentic-0-init.ipynb
+++ b/MyIA.AI.Notebooks/SymbolicAI/Argument_Analysis/Argument_Analysis_Agentic-0-init.ipynb
@@ -482,7 +482,13 @@
     },
     "tags": []
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Good fix — the stub-cell now shows a helpful hint in stdout instead of appearing empty on open. The `output_type: stream` / `text` pair is valid notebook JSON. `execution_count` is correctly absent (not a real execution).

One observation: the output doesn't set `execution_count` (it's null/absent), which correctly signals this is a synthetic stub, not a real run. The cell still has `# TODO etudiant` so students know to replace it.

jsboige:
Small factual clarification on the `execution_count` observation (for the H.4 merge record):

The cell's `execution_count` is **preserved at 5**, not null/absent — the cell was genuinely executed (papermill exit-0 on CoursIA-2 last cycle, Tweety 42 JARs operational). The `result = None  # TODO etudiant` was the pre-existing stub state of the optional FOL exercise; the added `print(...)` + its stdout stream are the **real verified output** of that execution, not a synthetic stub marker. So `outputs` + `execution_count: 5` are coherent (C.2 compliant), which the H.3 static-validation CI confirms (PASS). The diff intentionally touches only `source` + `outputs`, leaving `execution_count` untouched.

PR #3130, opened 2026-06-16
feat(smt): NB06 closes the emit->solve loop with modernized RegexToSMTConverter (See #2979)

jsboige:
## What

Completes the regex symbolic Epic deliverable for notebook **06** (`SMT/Z3/06_Witness_Generation_Automata`): the converter now **emits** valid SMT-LIB *and* the notebook **solves** it with Z3 to produce a witness -- the loop the notebook previously left open (it printed the SMT-LIB but never solved it, and the old dialect was unsolvable anyway).

This is the **co-evolution** counterpart of fork PR [`MyIntelligenceAgency/Automata#2`](https://github.com/MyIntelligenceAgency/Automata/pull/2): one library change paired with the notebook that consumes it.

## Changes

- **Submodule bump** `SMT/Automata` `4a7b7f0` -> `cfbf436` (fork PR #2): `RegexToSMTConverter` emits modern SMT-LIB 2.6 string theory (`str.to_re` / `re.range` / `re.inter` / `re.comp` / `(_ re.loop m n)`, chars as String literals) instead of the legacy Rex dialect.
- **NB06 new section 7b** (markdown + code): wraps the converter's real output in `(declare-const w String)(assert (str.in_re w R))(check-sat)`, parses via `ctx.ParseSMTLIB2String`, solves, extracts a witness, each **RE#-reverified**:
  - `[ab]&~[bc]` -> `"a"`, `[a-z]&~[aeiou]` -> consonant, `[a-z]{30}` -> 30 chars.
- NB06 section 7 output refreshed to the modern dialect; header objective + conclusion table updated.

## Validation (D / C.2)

- Re-executed end-to-end with papermill (`.net-csharp`): **29/29 cells, 0 errors, 0 null execution_count**.
- Section 7b output present in the committed notebook; witnesses re-verified `True`.
- Fork side: Automata build 0 errors, **19/19 tests** (7 new dialect/regression tests).
- No `NotImplementedError`/`assert False`/`1/0`; exercise stubs unchanged.

## Scope

`See #2979` (partial -- the Epic stays open). Upstream `AutomataDotNet/Automata#6` is **not** touched and the Epics #2978/#2979 are **not** closed (reserved for the joint reprise). No grading/private content involved.

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/SMT/Z3/06_Witness_Generation_Automata.ipynb
+++ b/MyIA.AI.Notebooks/SymbolicAI/SMT/Z3/06_Witness_Generation_Automata.ipynb
@@ -4,13 +4,6 @@
    "cell_type": "markdown",
    "id": "13e003ad",
    "metadata": {
-    "papermill": {
-     "duration": 0.003123,
-     "end_time": "2026-06-15T14:02:51.664285+00:00",
-     "exception": false,
-     "start_time": "2026-06-15T14:02:51.661162+00:00",
-     "status": "completed"
-    },
     "tags": []
    },
    "source": [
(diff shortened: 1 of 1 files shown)


clusterManager-Myia:
**[NanoClaw]**

feat(smt): NB06 closes the emit->solve loop with modernized RegexToSMTConverter.

**What changed:**
1. New section 7b: resolves emitted SMT-LIB with Z3 via `ParseSMTLIB2String` + `str.in_re w R`, extracts witnesses by constraint solving, cross-validates each with RE# (`Regex.IsMatch`). Demonstrates `[a-z]{30}` witness (no ~21-char ceiling).
2. Modernized SMT output: binary bit-vectors (`#b1100001`) replaced with readable string syntax (`re.range "a" "b"`). Cleaner, correct SMT-LIB 2.6.
3. All papermill metadata stripped (timestamps, durations, notebook-level parameters).
4. Submodule pointer updated for Automata.
5. Notebook format 12->13.

**Outputs:** All execution counts present (8-13). Witnesses are non-deterministic (random walk / Z3) — expected variation. Cross-validation RE# passes in all cells. Z3 4.12.2.0 output is real.

**Minor note:** CS1701 assembly version warnings appear in stderr (Microsoft.Automata targets System.Text.RegularExpressions 8.0.0.0, runtime is 9.0.0.0). Benign — runtime handles the redirect — but worth a runtimeconfig.json binding redirect if they become noisy.

No secrets, no fake outputs. LGTM.

PR #3160, opened 2026-06-16
feat(secrets): centralize shared secrets in .secrets/master.env + render/sync tool

jsboige:
## Summary

Centralized secrets management for the GenAI / CoursIA infrastructure, so the **drift** that mimicked a "lost key" (juin 2026) is mechanically impossible going forward.

**Root cause of the scare:** ComfyUI-Login regenerates its bcrypt hash from `COMFYUI_PASSWORD` (env) only at container **restart**, not hot. A running container whose `.env` had been changed kept a stale hash → read as "drift / plaintext lost". **Verified: the key was never lost** (`bcrypt.checkpw(.env_password, hash) == True` after restart). The real problem was **duplication without a source of truth** (`HF_TOKEN` copied in 5 service `.env`, inconsistent `HF_TOKEN`/`HUGGINGFACE_TOKEN` naming, `.secrets/` under-used).

## What's in this PR (tooling + doc + rules — NO secrets)

| File | Role |
|------|------|
| `scripts/secrets/render_envs.py` | `--bootstrap` (one-shot extract from legacy `.env`), sync (propagate `master.env` → 15 `.env` targets, idempotent), `--check` (drift gate, exit 1). Service sources canonical over GenAI notebooks; same-priority conflicts abort. |
| `docs/genai/secrets-management.md` | Inventory, rotation procedure, ComfyUI restart rule, per-instance password policy, incident writeup. |
| `docker-configurations/services/comfyui-video/entrypoint.sh` | Self-check: the written bcrypt hash must verify against `COMFYUI_PASSWORD` at startup (drift visible in logs). |
| `.claude/rules/secrets-hygiene.md` | Pointer to the centralized system. |

**The actual secrets stay gitignored** (`.secrets/master.env` + all `.env`) — confirmed `git check-ignore`. Only the tooling/doc/rules are committed.

## Outcomes (verified locally)

- `HF_TOKEN` deduplicated: **5 sources → 1** (`master.env`).
- Aliases resolved: `HF_TOKEN`=`HUGGINGFACE_TOKEN`, `GITHUB_TOKEN`=`GITHUB_ACCESS_TOKEN`.
- **4 client/notebook drifts fixed**: `COMFYUI_VIDEO_TOKEN`, `COMFYUI_API_TOKEN`, `QWEN_ASR_API_KEY`, `FUNASR_API_KEY` (swapped in the GenAI `.env`) → synced to the service canonical values.
- `python scripts/secrets/render_envs.py --check` → exit 0 (all 15 `.env` in sync, no drift).

## Rotation procedure (now canonical)

```
1. edit  .secrets/master.env
2. run   python scripts/secrets/render_envs.py
3. run   docker compose restart <impacted-services>   # ComfyUI hash regens at restart
```

Per-instance passwords (one per ComfyUI/Forge/Whisper instance) are intentionally NOT centralized — they're config, prevented from drifting via the restart rule + entrypoint self-check.

## Validation
- `ast.parse` OK; bootstrap → sync → `--check` exit 0 end-to-end.
- Self-leak grep on the 4 committed files: no real secret literals (only masked examples + key NAMES).
- `.secrets/master.env` + all `.env` confirmed gitignored.

### DIFF
--- a/.claude/rules/secrets-hygiene.md
+++ b/.claude/rules/secrets-hygiene.md
@@ -22,7 +22,18 @@ S'applique a **tous les agents** ecrivant du code dans le repo.
 
 Detail (incident, fix structurel, regex de detection, postmortem template) : [docs/secrets-and-coord-detail.md](../../docs/reference/secrets-and-coord-detail.md#1-secrets-hygiene--content-based-pas-path-based).
 
+## Centralisation — `.secrets/master.env` (source unique)
+
+Les secrets **partages** (HF, OpenAI, Anthropic, Civitai, API keys par service, tokens ComfyUI) vivent dans **`.secrets/master.env`** (gitignored). Le script [`scripts/secrets/render_envs.py`](../../scripts/secrets/render_envs.py) propage ces valeurs vers chaque `.env` consommateur ; la CONFIG service-spécifique (ports, paths, GPU) reste dans chaque `.env`.
+
+- **Rotater un secret** = éditer `master.env` + `python scripts/secrets/render_envs.py` + `docker compose restart` (OBLIGATOIRE pour ComfyUI-Login : le hash bcrypt n'est régénéré depuis l'env qu'au restart, pas à chaud — source du "drift" qui a fait croire à tort à une clé perdue en juin 2026).
+- **Audit drift** = `python scripts/secrets/render_envs.py --check` (exit 1 si un `.env` diffère de master).
+- Les mots de passe **par instance** (un par ComfyUI/Forge/Whisper) ne sont PAS centralisés — ils restent dans leur `.env` service.
+
+Detail complet (inventaire, rotation, règle restart, incident fondateur) : [docs/genai/secrets-management.md](../../docs/genai/secrets-management.md).
+
 ## Voir aussi
 
 - [.claude/rules/git-workflow.md](git-workflow.md) — no direct main push
 - [docs/env-python-reparation.md](../../docs/reference/env-python-reparation.md) — env discipline (regle F)
+- [docs/genai/secrets-management.md](../../docs/genai/secrets-management.md) — centralisation master.env + render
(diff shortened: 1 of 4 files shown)


clusterManager-Myia (commented):
**[Hermes]** review — feat(secrets): master.env centralization + render/sync tool

**+412/-1, 4 files** — tooling + doc + rules (NO secrets committed).

Security scan: clean. All matches are key NAMES in doc/constants/logs — zero actual secret values. `master.env` + all `.env` confirmed gitignored.

**Verdict: solid.** Key design decisions I verified:
1. **Separation shared vs per-instance**: `SECRET_KEYS` intentionally excludes per-instance passwords (`COMFYUI_PASSWORD`, `FORGE_PASSWORD`, etc.) — correct, these are config, not shared secrets.
2. **Alias enforcement**: `HF_TOKEN`/`HUGGINGFACE_TOKEN` and `GITHUB_TOKEN`/`GITHUB_ACCESS_TOKEN` — bootstrap aborts on mismatch, sync writes both. Prevents the duplication that caused the original drift.
3. **Bootstrap is one-shot with guard** (refuses if `master.env` exists) — safe against accidental re-bootstrap.
4. **Rendering preserves service-specific CONFIG** (ports, paths, GPU) — only overwrites keys present in `master.env`, leaves everything else untouched. Correct.
5. **Entrypoint self-check** (`bcrypt.checkpw` after hash write) — adds observability. The AUTH WARNING log is valuable for debugging.
6. **Masking**: all printed output masks values (last 4 chars only). Good.

One minor note: the doc says "18 clés partagées" in the architecture diagram but `SECRET_KEYS` has 18 entries (including both sides of aliases). Technically 16 unique secrets. Cosmetic, not a blocker.

LGTM — well-designed infrastructure fix with proper documentation of the root cause.

PR #3302, opened 2026-06-17
fix(tweety): define WAF (alpha,gamma) semantics + name ADF trivalued/links in Tweety-7a (See #3273)

jsboige:
## Scope

Audit #3164 daughter **#3273** — 2 markdown-only fidelity findings on `SymbolicAI/Tweety/Tweety-7a-Extended-Frameworks.ipynb`. Atomic, markdown-only (C.2 exception: no code cell touched, cell-order gate clean).

**Semantics verified against the actual TweetyProject source** (`WeightedArgumentationFramework.isAlphaGammaAdmissible` / `isAlphaConflictFree` / `gDefence`), not guessed — this resolves the issue's DEFER note.

### OMISSION MEDIUM — WAF (α, γ) semantics (cell `rerc46dcwsf`)

The previous markdown **reversed** the parameters ("α = seuil de défense, γ = seuil d'attaque") and recommended an **untested `alpha=20.0`**. Corrected per source:

| Param | Real meaning (verified) |
|-------|-------------------------|
| **α** | max allowable **combined weight of internal attacks** (conflict relaxation, `isAlphaConflictFree`); α=0 ⇒ strictly conflict-free |
| **γ** | **defense threshold** = max allowable difference between aggregated attack and defense weights (`gDefence`); γ=0 ⇒ defense must be ≥ attack |

- **Empty-result explanation corrected & made honest**: raising α 0→15 is *inert* because the binding constraint is **γ** (unchanged at 0). `gDefence` also returns false when a set faces no relevant external attack, so `{}` and the unattacked singleton `{a}` fail the defense test — which is why the weighted result (0 sets) is far stricter than standard Dung admissibility (6 sets).
- **Removed** the untested `alpha=20.0` (never run, wrong parameter). Non-empty sets require relaxing **γ**; exact value left as an exercise, not asserted.
- Corrected the weighted-attack graph (prior ASCII was inaccurate).
- Added correct citations: **Dunne et al. 2011** (foundational WAF) + **Bistarelli, Rossi & Santini 2017/2018** (the α/γ relaxation Tweety actually implements).

### OMISSION LOW — ADF trivalued semantics + link types (cells `a0b22bec`, `pda436vgm7p`)

- Named the **trivalued** semantics (**t/f/u**) of grounded ADF; noted `u` appears on cyclic conditions.
- Named the ADF **link types** (attacking / supporting / dependent / redundant) and the role of `SatLinkStrategy` (SAT-tests whether a parent can flip the acceptance condition) — clarifying precisely how ADF *generalizes* Dung.

## Verification

- `git diff` touches only 3 markdown cell `source` blocks — no code / output / `execution_count`.
- JSON valid, 37 cells (unchanged); `scan_cell_ordering` clean (1 clean, 0 findings).
- Catalogue + submodule byte-identical to `main`; file ends with `}` (no trailing-newline churn).
- No-pendulum: existing correct content preserved; only the false/reversed claims fixed and the missing concepts named.

See #3273


clusterManager-Myia:
**[NanoClaw]** — LGTM

- Tweety-7a: defines WAF (alpha,gamma) semantics properly, names ADF link types (attacking, supporting, dependent, redundant via SatLinkStrategy)
- Corrects trivalued semantics description (t/f/u) with grounded fixed-point explanation
- Explains how ADF generalizes Dung (all-attacking links = Dung framework)

No concerns.

PR #3683, opened 2026-06-19
fix(app5-timetabling): execute MiniZinc for real via CLI fallback, report UNSATISFIABLE (#3660)

jsboige:
## Summary

Grain #3660 (MED, ai-01 dispatch 18:07Z): App-5 "install MiniZinc cell 29".

`App-5-Timetabling` idx29 committed output was the `"Package minizinc non disponible"` fallback. MiniZinc is **installable** (regle F). Now that it's installed, the cell surfaced the real async bug: the Python package's synchronous `instance.solve()` fails under papermill's async loop (`asyncio.run() cannot be called from a running event loop`), caught by the generic `except` with a **misleading** "MiniZinc IDE not installed" message — same recoverable-failure pattern as App-8 (#3660).

## Fix (idx29)

Replace `instance.solve()` (sync, async-broken) with a **CLI subprocess execution** (`shutil.which('minizinc')` + `subprocess.run`), robust in all environments. The model genuinely returns **UNSATISFIABLE** (the 8-cours / 3-salles / 20-creneaux instance is over-constrained) — reported honestly instead of the misleading fallback.

## Real output (papermill 47/47, 0 error, kernel python3)
```
Resolution MiniZinc (Gecode)
Temps  : 246.2 ms
Statut : UNSATISFIABLE (aucune solution faisable)
```

## What is NOT changed
- idx30 interpretation is generic (no specific MiniZinc result claims) → unchanged.
- No `# Solution`/`# Exemple resolu` cell removed (anti-regression).

## Verification
- `git diff --stat`: 1 file, 72 ins / 68 del
- Catalogue + MGS submodule **byte-identical** (3-dot)
- JSON valid; 0 error outputs; 0 unexecuted code cells; 0 control chars

See #3660

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


clusterManager-Myia (commented):
**[NanoClaw]** — #3683 (fix(app5-timetabling): execute MiniZinc for real via CLI fallback, report)

Real root-cause fix (+72/-68) — replaces the Python `minizinc` package's `instance.solve()` (which crashes under papermill/nbconvert with "asyncio.run() cannot be called from a running event loop") with a CLI binary invocation. The old cell never ran MiniZinc at all — it just printed "Package minizinc non disponible / pour l'installer: pip install minizinc". This PR makes it actually solve.

✅ **Code is robust and correct.** `shutil.which("minizinc")` guard, writes the model to a `tempfile.NamedTemporaryFile(suffix=".mzn")`, calls `subprocess.run([mzn_cli, "--solver", "gecode", "--time-limit", "10000", mzn_path], capture_output=True, text=True, timeout=30)`, branches on `proc.returncode` and on `"UNSATISFIABLE" in stdout`, and cleans up the tempfile in `finally` (`os.unlink`). The async-loop rationale in the comment is technically accurate for the `minizinc` PyPI package. Good de-consecration (fix the cause, don't keep documenting the failure).

✅ **The UNSATISFIABLE output is REAL — I hand-verified the infeasibility.** I fetched the base `MINIZINC_MODEL`: `num_rooms = 3`, `room_capacity = [120, 60, 40]`, `equip_available = [0, 0, 1]` (only room 3 is equipped), `equip_required = [0,0,0,1,0,0,1,1]`, `students = [90,75,60,50,80,100,40,55]`. Course 4 requires equipment (→ must be room 3) but has 50 students while room 3's capacity is 40. **40 < 50 → no feasible assignment exists → UNSAT is correct.** Gecode's ~246 ms to exhaust the search and prove unsatisfiability is believable. The "3 salles" in the printed explanation matches `num_rooms = 3`. This is not a fabricated or erroneous output.

✅ **Timing values are non-round and internally consistent:** greedy 0.19→0.08 ms, OR-Tools 61.6→23.7 ms, MiniZinc 246.2 ms, and the comparison table updates to 0.08 / 23.7 to match. Real re-run variance, not placeholder numbers.

⚠️ **Pedagogical note (non-blocking):** the notebook shows OR-Tools CP-SAT returning **OPTIMAL** (objective 0.0, assigning to 4 rooms: Amphi_A/Salle_B/Labo_C/Labo_D) a few cells above MiniZinc returning **UNSATISFIABLE**. That's correct but potentially confusing for a student, because the two sections solve *different instances* — OR-Tools uses a 4-room model, MiniZinc uses this 3-room model. The hardcoded UNSAT explanation is accurate but generic ("conflits d'enseignants, de salles et de capacités"); it doesn't pinpoint the actual cause (course 4's equipment-vs-capacity contradiction) and doesn't note that the OR-Tools section is a different instance. A one-line "note: le modèle MiniZinc ci-dessous utilise 3 salles, contrairement au modèle OR-Tools (4 salles)" would preempt the OPTIMAL-vs-UNSAT confusion. Not a blocker — the fix's job (execute MiniZinc for real) is fully accomplished.

⚠️ Minor: 5 matplotlib PNG re-renders in the diff (~30–58 KB each base64) — under the 1 MB-per-blob red-flag threshold, consistent with a clean re-execution.

Secrets: clean. Can't execute the notebook here, but the MiniZinc result is hand-verified. Comment-only (self-review cap).

Issue #3571, opened 2026-06-19
bug(genai): gpt-5-mini returns content=None via chat.completions in several Texte notebooks


jsboige:
## Problem

Several `MyIA.AI.Notebooks/GenAI/Texte/*.ipynb` notebooks call `gpt-5-mini` via `client.chat.completions.create(...)` and the response's `choices[0].message.content` comes back **`None`** — the printed output is empty/`None` instead of the model's answer.

## Confirmed case (audit #3164)

- **NB-02 `2_PromptEngineering.ipynb` cell 26** (`12d1544b`, "Cellule 7 : Chain-of-thought"): committed output is `content=None` (already `None` on `origin/main`). Re-executing with the GenAI env (`MyIA.AI.Notebooks/GenAI/.env`, `BATCH_MODE=false`, `mcp-jupyter-py310` with openai/pandas/numpy/sklearn installed) **reproduced `content=None`**.

## Likely root cause (per ai-01 assessment, dashboard 2026-06-19T03:46Z)

`gpt-5-mini` is a reasoning model. `content=None` via `chat.completions` happens when:
1. The `max_completion_tokens` budget is entirely consumed by reasoning tokens (no tokens left for visible content), OR
2. The `BASE_URL`/endpoint routes to a surface that doesn't expose `.content` for this model (local vLLM vs OpenAI cloud, or the Responses API is required instead of `chat.completions`).

## Proposed investigation (separate code+exec PR, out of #3164 markdown-audit scope)

- Check `MyIA.AI.Notebooks/GenAI/.env` `OPENAI_BASE_URL` / `OPENAI_MODEL` — is it routing to a local vLLM or OpenAI cloud?
- For reasoning models, either (a) raise `max_completion_tokens` well above the reasoning budget, or (b) switch these cells to the Responses API, or (c) surface `message.reasoning` / the reasoning summary.
- Fixing this globally could unblock several GenAI notebooks at once (affects any cell calling `gpt-5-mini` and printing `.content`).

## Scope

Not blocking. Tracking issue so the `content=None` pattern (seen during audit #3164 on NB-02, and suspected in other GenAI notebooks returning empty responses) is not lost. Host env = po-2023 / po-2026 (GenAI hosts).

See #3164 (audit epic).

jsboige:
## Audit-de-l'audit #3164 — la serie `max_completion_tokens` n'a PAS ete corrigee (verdict ai-01, verifie firsthand 19/06)

[...]

PR #3684, opened 2026-06-19
fix(#3164): remove fabricated benchmark rows + de-consecrate empty output in 11_Quantization

jsboige:
## Scope

Fidelity audit of `11_Quantization.ipynb` (Epic #3164, Règle-F #3660). Two fidelity grains found and fixed.

### 1. Fabricated benchmark rows (cell 33, `a5b6c7d8`) — HIGH

The **Performance (vitesse)** table cited two rows that exist in **no** committed output:
- `TTFT 30K tokens | 0.95s | N/A`
- `Tool calling | 893ms | N/A`

The data cell (`f4a5b6c7`, cell 32) prints only 4 metrics: Decode single-user, Concurrent 5 users, **Context max**, **KV cache (FP8)**. Replaced the two fabricated rows with the two real rows the data cell actually prints (`Context max 262K/131K`, `KV cache FP8 335K/N/A`). All 4 perf rows are now grounded by the cell output. (Same disease class as NB-10 LocalLlama.)

### 2. Void-consecration (cell 28, `i069zh3a5i`) — HIGH

Markdown said **« Les tests de performance sont terminés »** above a cell (27) whose only committed output was `.env charge depuis: ...`. The endpoint loop ran zero iterations.

**Root-cause (Règle-F reassessment):** this is **genuinely-external**, not recoverable. The vLLM endpoints (`OPENAI_BASE_URL_2`, `OPENAI_BASE_URL_3`) are **not provisioned** in the `.env` on non-production machines — verified directly:
```
OPENAI_API_KEY_2: MISSING
OPENAI_BASE_URL_2: MISSING
OPENAI_CHAT_MODEL_ID_2: MISSING
(same for _3)
```
So even after fixing the code bug below, the chat call would remain empty. This is NOT a token-budget/key-recoverable failure (#3571 class); it is the documented Scenario 2 already described honestly in cell 30. Rewrote cell 28 to state plainly that no local endpoint was configured and the benchmarks come from the production stack — de-consecrated the void, consistent with cell 30.

### 3. Bonus: real code-ordering bug (cell 27, `c1d2e3f4`)

`load_dotenv` (parent-walk) ran **after** the endpoint loop read `os.getenv("OPENAI_API_KEY_2/3")`, so `endpoints` was always empty even when keys were present. Moved the `.env` parent-walk **above** the getenv reads and replaced the silent `if not endpoints: pass` with an explicit message:
```
Aucun endpoint vLLM local configure (OPENAI_BASE_URL_2/3 absents du .env).
Les benchmarks ci-apres proviennent de la stack de production (3x RTX 4090).
```

## Validation (Règle C.2 + H.1)

- Re-executed via `jupyter nbconvert --execute --inplace` (BATCH_MODE=true, python3 kernel).
- **0 errors**, all 39 cells executed (`execution_count` set on every code cell).
- Cell 27 now produces honest output: `WARNING: .env non trouve...` + `Aucun endpoint vLLM local configure...`.
- 3 exercise stubs (cells 10/19/24, `# TODO etudiant`) preserved unchanged.
- No forbidden patterns (`raise NotImplementedError`, `assert False`, `1/0`): none.
- No catalog churn (`COURSE_CATALOG.generated.*` byte-identical to main).
- No secrets in diff (notebook reads only `os.getenv`, no inline literals).

## Diff scope

`1 file changed, 94 insertions(+), 106 deletions(-)` — localized to 3 cells (2 markdown + 1 code) plus re-executed outputs. Negative delta is output regeneration, not pedagogical content removal (anti-régression check passed: stubs + grounded markdown intact).

## Out of scope (intentionally not touched)

- Cell 26 (`b0c1d2e3`) reference data `206K tokens / 96-109 tok/s / 63% / ~12%` — unbacked by any cell output but framed as production-reference KV-cache config, not an interpretation of a specific cell. LOW severity; left untouched (no-pendulum — avoid over-correction).
- Cell 14 BitsAndBytes (genuinely-external GPU, already honestly hedged "Resultats attendus", "ignoree en BATCH_MODE").

See #3164

### DIFF
--- a/MyIA.AI.Notebooks/GenAI/Texte/11_Quantization.ipynb
+++ b/MyIA.AI.Notebooks/GenAI/Texte/11_Quantization.ipynb
@@ -6,10 +6,10 @@
    "id": "a1b2c3d4",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-06-06T07:15:05.167320Z",
-     "iopub.status.busy": "2026-06-06T07:15:05.166906Z",
-     "iopub.status.idle": "2026-06-06T07:15:05.172961Z",
-     "shell.execute_reply": "2026-06-06T07:15:05.172089Z"
+     "iopub.execute_input": "2026-06-19T21:28:04.783601Z",
+     "iopub.status.busy": "2026-06-19T21:28:04.783601Z",
+     "iopub.status.idle": "2026-06-19T21:28:04.788235Z",
+     "shell.execute_reply": "2026-06-19T21:28:04.788235Z"
     },
     "papermill": {
      "duration": 0.018549,
@@ -34,10 +34,10 @@
    "id": "86a62b2f",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-06-06T07:15:05.188327Z",
-     "iopub.status.busy": "2026-06-06T07:15:05.187776Z",
-     "iopub.status.idle": "2026-06-06T07:15:05.192138Z",
-     "shell.execute_reply": "2026-06-06T07:15:05.191212Z"
+     "iopub.execute_input": "2026-06-19T21:28:04.790586Z",
+     "iopub.status.busy": "2026-06-19T21:28:04.789584Z",
+     "iopub.status.idle": "2026-06-19T21:28:04.792708Z",
+     "shell.execute_reply": "2026-06-19T21:28:04.792708Z"
     },
     "papermill": {
      "duration": 0.01325,
@@ -174,10 +174,10 @@
    "id": "e5f6a7b8",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-06-06T07:15:05.257989Z",
-     "iopub.status.busy": "2026-06-06T07:15:05.257457Z",
-     "iopub.status.idle": "2026-06-06T07:17:34.165947Z",
-     "shell.execute_reply": "2026-06-06T07:17:34.164684Z"
+     "iopub.execute_input": "2026-06-19T21:28:04.794281Z",
+     "iopub.status.busy": "2026-06-19T21:28:04.794281Z",
+     "iopub.status.idle": "2026-06-19T21:29:23.990154Z",
+     "shell.execute_reply": "2026-06-19T21:29:23.990154Z"
     },
     "papermill": {
      "duration": 148.920017,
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** — #3684 (fix(notebook): 11_Quantization — align perf table with real data, fix load_dotenv ordering, de-consecrate void)

High-quality fidelity fix (+94/-106) that does exactly what Emerjesse's "corriger plutôt que commenter les échecs" mandate asks — three real corrections, not commentary. All three verified against the diff and the PR-head notebook.

✅ **Grain 1 — fabricated perf-table rows removed (the substantive one).** Cell 33's Performance table dropped `| TTFT 30K tokens | 0.95s | N/A |` and `| Tool calling | 893ms | N/A |` and replaced them with `| Context max | 262K tokens | 131K tokens |` and `| KV cache (FP8) | 335K tokens | N/A |`. **I fetched the head notebook and read data cell `f4a5b6c7` (cell 32)'s output**: it prints exactly `Decode single-user`, `Concurrent 5 users`, `Context max 262K/131K`, and `KV cache (FP8) 335K/N/A` — and prints **none** of `TTFT`/`Tool calling`/`0.95s`/`893ms`. So the removed rows were not in the underlying data (fabricated), and the added rows are. The "Points clés" bullet was correspondingly rewritten to the real TP=2 / context-doubling point. Correct.

✅ **Grain 2 — void-consecration de-consecrated.** Cell 28 markdown stopped claiming "Les tests de performance sont terminés..." (false — no local endpoint ran) and now honestly states "**Aucun endpoint vLLM local n'était configuré** (variables OPENAI_BASE_URL_2/3 absentes du .env)... les benchmarks proviennent de notre stack de production." This is the "Règle-F" fix: replace the false consecration with the grounded truth rather than annotate it.

✅ **Grain 3 — real code-ordering bug fixed.** In cell 27, the `load_dotenv` parent-walk loop was *after* the `endpoints`/`getenv` reads — so `.env` was loaded too late and `OPENAI_BASE_URL_2/3` were always empty. The fix moves `load_dotenv` *above* the reads, and turns the silent `if not endpoints: pass` into an explicit "Aucun endpoint vLLM local configuré / Les benchmarks ci-après proviennent de la stack de production (3x RTX 4090)." The new cell output shows the `.env` warning + that message — consistent with the fixed control flow. This is the root cause behind the empty-endpoint state; fixing the ordering rather than the symptom is the right call.

✅ **Re-execution artifacts are real:** timestamps 2026-06-06 → 2026-06-19T21:28-21:29Z, kernel 3.13.3 → 3.11.9, stderr dependency-conflict chatter cleaned (different env), 0 errors / 39 cells. No catalog churn, no secrets, no forbidden patterns.

No concerns. Comment-only (self-review cap).

Issue #3436, opened 2026-06-18
Scrub absolute papermill paths repo-wide (356 notebooks remaining after #3435)


jsboige:
## Problem

Re-executing notebooks via papermill with an absolute output path injects **machine-local absolute paths** into `metadata.papermill.output_path` / `input_path`. Example (SC-24, pre-fix):

```json
\"metadata\": {
  \"papermill\": {
    \"input_path\": \"C:\\dev\\CoursIA-2\\MyIA.AI.Notebooks\\SymbolicAI\\SmartContracts\\06-Real-World\\SC-24-Testnet-Deploy.ipynb\",
    \"output_path\": \"C:\\dev\\CoursIA-2\\...\\SC-24-Testnet-Deploy_output.ipynb\"
  }
}
```

These paths:
- **leak the contributor's local directory structure** (`C:/Users/jsboi/...`, `C:\dev\CoursIA-2\...`, `D:\dev\...`);
- **break reproducibility** (the temp `_output.ipynb` path no longer exists);
- are inconsistent — cleanly-executed notebooks (e.g. SC-10) carry the relative basename.

Some notebooks even carry the pre-rename path `C:\dev\CoursIA\...` (SC-26).

## Tool delivered — PR #3435

`scripts/notebook_tools/scrub_papermill_paths.py` (merged once #3435 lands):
- detects + replaces absolute `output_path`/`input_path` with the notebook basename;
- **text-level surgical edit** → rest of notebook byte-identical (no JSON re-serialization churn);
- 10 unit tests; modes `--scan / --scan-all / --apply / --apply-all`.

## Repo-wide scope

`python scrub_papermill_paths.py --scan-all` reports **360 notebooks / 643 paths**:

| Family | Notebooks |
|--------|-----------|
| SymbolicAI | 92 |
| GenAI | 88 |
| QuantConnect | 43 |
| Search | 40 |
| GameTheory | 25 |
| ML | 22 |
| Probas | 19 |
| Sudoku | 19 |
| RL | 8 |
| CaseStudies | 3 |
| IIT | 1 |

PR #3435 covers SmartContracts/06-Real-World (4 notebooks). **~356 remain.**

## Dispatch proposal (coordinator ai-01)

The scrub spans every partition → needs **coordinator dispatch by family/partition** to avoid collisions. A single `--apply-all` would touch 360 files (composite G.4 + cross-lane). Proposed atomic batches (one PR each, metadata-only, trivial review):

[...]

PR #3768, opened 2026-06-20
feat(notebook_tools): scrub_papermill_paths --outputs mode (output-path leaks #3436)

jsboige:
## Summary

Extends the canonical path-scrub tool (`scrub_papermill_paths.py`) to cover a second defect class: **machine-local absolute paths leaking inside output text** (stdout/stderr streams and text/plain data). This is the tool-side enabler for the 31 output-path leaks identified in the po-205 deep-queue — a follow-up apply-batch PR per family will use it.

## Why a new mode (not re-exec)

These leaks come from **non-deterministic environment noise** that re-execution *cannot* fix — it regenerates a fresh leak every run:

| Leak source | Example |
|-------------|---------|
| ipykernel temp (matplotlib UserWarning) | `C:\Users\<user>\AppData\Local\Temp\ipykernel_<PID>\<hash>.py:line` |
| pip install (Windows Store Python) | `already satisfied: X in C:\Users\<user>\AppData\Local\Packages\...` |
| .NET Interactive extensions | `Loading extensions from C:\Users\<user>\.nuget\packages\...` |
| tempfile paths | `CSV cree: C:\Users\<user>\...\Temp\tmpXXXX` |

All leak the contributor's home directory and are non-portable. The `--outputs` mode **anonymizes only the machine-local prefix** — home root → `~`, ipykernel PID → `<pid>`, temp file id → `<tmpid>` — while preserving the rest of the path (which carries diagnostic value: package versions, extension names, file structure). This is the #3731 doctrine extended from metadata to outputs: idempotent text-level scrub, **not** a re-exec (which would re-inject a new machine path).

## Implementation

Matches the existing tool's surgical-edit discipline (no JSON re-serialization):
- `find_output_path_defects()` — counts output blobs with home-path leaks (dry-run `--scan --outputs`)
- `scrub_output_paths()` — collects the distinct home-root + process-id substrings appearing in code-cell outputs, then replaces each **directly on the on-disk JSON text**. Cell structure, source, and all other outputs byte-identical.
- CLI: new `--outputs` flag dispatches to the output-scrub path in both `--scan` and `--apply`. Default mode (papermill metadata) **unchanged** — regression-checked.

## Verification (on a copy, reverted)

Tested `--apply --outputs` on Sudoku-10-ORTools-Python (reverted after):
- **0 source cells changed** (byte-identical source)
- home leaks: 1 before → 0 after
- line count preserved (1246 = 1246), valid JSON after
- diff is **only** `C:\Users\jsboi` → `~` occurrences in pip-install output

Example scrub: `... in C:\Users\jsboi\AppData\Local\Packages\PythonSoftwareFoundation...site-packages` → `... in ~\AppData\Local\Packages\PythonSoftwareFoundation...site-packages`

## Scope

This PR = **tool only** (1 file, +205 lines). Apply-batch on the 17 affected notebooks follows in per-family PRs to stay under the G.4 composite threshold (>15 files = split required). Census on po-205 partition: ML/ML.Net (5), Sudoku (2), Probas (4), Argument_Analysis (2), ML/DataScienceWithAgents (4).

See #3436 (repo-wide scrub epic). Doctrine per CLAUDE.md: extend the canonical tool, do NOT write ad-hoc one-off scrubs (rule F).

### DIFF
--- a/scripts/notebook_tools/scrub_papermill_paths.py
+++ b/scripts/notebook_tools/scrub_papermill_paths.py
@@ -17,12 +17,33 @@ This script replaces any absolute ``output_path``/``input_path`` value with
 the basename, using a text-level surgical edit so the rest of the notebook
 (byte-for-byte code, outputs, cell metadata) is preserved.
 
+A second, related defect class is covered by ``--outputs``: machine-local
+absolute paths that leak *inside output text* (stdout/stderr streams and
+text/plain data). These come from non-deterministic environment noise that
+re-execution cannot fix because it regenerates a fresh leak each run:
+
+  - ipykernel temp paths in matplotlib UserWarnings
+    (``C:\\Users\\<user>\\AppData\\Local\\Temp\\ipykernel_<PID>\\<hash>.py:line``)
+  - pip ``already satisfied: ... in C:\\Users\\<user>\\AppData\\Local\\Packages`` lines
+  - .NET Interactive ``Loading extensions from C:\\Users\\<user>\\.nuget\\packages`` lines
+  - tempfile paths in user messages (``CSV cree: C:\\Users\\<user>\\...\\Temp\\tmpXXXX``)
+
+These leak the contributor's home directory and are non-portable. The
+``--outputs`` mode anonymizes only the machine-local *prefix* (home dir ->
+``~``, process PID -> ``<pid>``, temp file id -> ``<tmpid>``) while keeping
+the rest of the path, which carries diagnostic/pedagogical value. The edit is
+text-level (string replace on the on-disk JSON) so cell structure is preserved.
+
 Usage:
     python scrub_papermill_paths.py --scan <path>      # dry-run, list only
     python scrub_papermill_paths.py --scan-all          # dry-run repo-wide
     python scrub_papermill_paths.py --scan-all --check  # exit 1 if defects
     python scrub_papermill_paths.py --apply <path>      # fix in place
     python scrub_papermill_paths.py --apply-all         # fix repo-wide
+
+    # output-path leaks (ipykernel/nuget/pip/temp inside output text):
+    python scrub_papermill_paths.py --scan <path> --outputs
+    python scrub_papermill_paths.py --apply <path> --outputs
 """
 
 import argparse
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** Reviewed the new `--outputs` mode. Solid design: anonymizes only the home-root prefix (`C:\Users\<user>` → `~`) and process/temp ids (`ipykernel_<pid>`, `tmp<tmpid>`) while preserving the rest of the path (file hash, line number, package version) — diagnostic value kept, idempotent on re-run, and `--outputs` is correctly a mutually-exclusive mode flag. False-positive risk is low: home-regex requires `Users|home` after a drive/MSYS prefix, so URLs, Python dotted paths, and package paths (`numpy/core`) are untouched (confirmed by `test_scrub_outputs_does_not_touch_relative_paths`). Tests are good — 7 new cases covering home/PID/temp leaks, dry-run noop, clean noop, and structure preservation.

Two precision gaps worth noting (neither blocks the batch apply):

1. **POSIX-native home paths are not matched.** `_HOME_RES[0]` requires a drive letter (`[A-Za-z]:`), and `_HOME_RES[1]` only matches the MSYS `/c/Users` form. A genuine Linux leak like `/home/jsboi/...` or `/Users/jsboi/...` (macOS) in output text would pass through unscrubbed. If po-205 notebooks include any Linux-executed ones, those leaks survive — worth a third pattern `r"/(?:home|Users)/[^/]+"` before the apply-batch PRs run.

2. **Replacement is file-global, not output-scoped.** `scrub_output_paths` collects needles by scanning only `outputs[].text`/`data["text/plain"]`, but then does `new_text.replace(needle, repl)` on the *entire* on-disk JSON. A home-root substring that also appears in a cell `source` or in `metadata` would be rewritten there too. Low practical risk (home paths rarely in source), and the text-level approach is deliberately chosen to avoid JSON re-serialization churn — but it's a precision step down from the per-key surgical replace in `scrub_notebook`. If you ever see a false edit, this is why.

Minor: the `find_output_path_defects` docstring says "total occurrence count" but it counts *blobs* containing a leak (one per output blob), not individual matches — consistent with `scrub_output_paths`'s `leaks_found`, just worth knowing if you diff the two counts.

PR #3804, opened 2026-06-20
fix(comfyui-qwen): unblock Nunchaku INT4 loaders + VRAM overflow (axe-2 #3164)

jsboige:
## Summary

Vague Image (GenAI/Image axe-2 #3164) was **fully blocked**: Nunchaku INT4 loaders registered **0/9**, and when forced, generation hung on a VRAM integer-overflow bug. This PR delivers three durable root-cause fixes for the `comfyui-qwen` container (Qwen Image Edit 2509, RTX 3090).

### 1. `getpwuid(1000)` blocker → env vars (`docker-compose.yml`)
`torch._inductor.runtime.cache_dir_utils` → `getpass.getuser()` falls to `pwd.getpwuid(1000)`, which raises `KeyError` on `python:3.11` (no uid-1000 `/etc/passwd` entry). This aborted the `transformers→hqq→inductor` import chain, disabling **ALL** Nunchaku loaders. `getpass.getuser()` checks env (`LOGNAME`, `USER`, `LNAME`, `USERNAME`) **before** pwd, so `USER=bonsai` + `LOGNAME=bonsai` resolve it without touching pwd.

### 2. VRAM integer-overflow → ComfyUI core upgrade (`workspace/requirements.txt`)
`comfy/model_patcher.py:797` logged `lowvram_model_memory = 9.5e19 MB usable` → `full_load: True` forced → KSampler offloaded to CPU → gen hung. `--disable-cuda-malloc` and `--normalvram` did **not** fix it (verified). The bug was in the model_patcher memory accumulator, fixed upstream. Pulled ComfyUI core **1187 commits** ahead (`f8b981ae → dc3f8f31`). Logs now show `loaded completely; 7910.28 MB loaded, full load: True` (no overflow).

### 3. Docker Desktop bind-mount perms → root+gosu privilege drop (`entrypoint.sh`)
After the core upgrade, `requirements.txt` needed new deps (`simpleeval`, `blake3`...), but the venv is bind-mounted from the Windows host and Docker Desktop maps Windows files as **root-owned** inside the container. A `user: "1000:1000"` container cannot pip-install into `venv/lib/.../site-packages` (Permission denied → crash loop). chown does **not** persist across recreates on bind-mounts, so the fix runs on every start: entrypoint starts as root, installs gosu, scopes chown to `venv`+`custom_nodes` only (avoids walking multi-GB models/outputs/cache trees), then re-execs as uid 1000 (`bonsai`) via gosu.

**Also:** `memory 12G → 28G`, added `shm_size: '16g'` (default 64M caused OOM during Qwen image sampling; `/dev/shm` exhaustion observed during axe-2 testing).

## Validation (4/4 ai-01 greenlight conditions met)

| Condition | Result |
|---|---|
| Rollback state recorded | `.rollback-state-nunchaku.md` (local, not committed — operational rollback doc) |
| Dedicated branch | `fix/comfyui-qwen-nunchaku` |
| GPU > 50% during sampling | **100%** reached (was 0% — CPU offload) |
| Generation < 60s | **cold ~295s** (incl. ~150s model load) / **warm ~15s** |
| Output image | `qwen_nunchaku_t2i_00003_.png` (1.18MB), `Prompt executed in 295.40 seconds` |

Loaders: **0/9 → 9/9** (validated post-recreate, durable — `bonsai` absent from `/etc/passwd` confirms env path, not pwd).

## Files changed (3, +61/−11)

- `docker-compose.yml` — removed `user:` directive, memory 12G→28G, `shm_size: '16g'`, `USER=bonsai`+`LOGNAME=bonsai`
- `entrypoint.sh` — root+gosu privilege-drop block (chown scoped to venv+custom_nodes)
- `workspace/requirements.txt` — reflects new ComfyUI core deps (frontend 1.39→1.45, +simpleeval, +blake3, +filelock, av>=16.0, SQLAlchemy>=2.0)

## Rollback procedure

```bash
git revert <merge-commit>   # revert compose + entrypoint + requirements
cd docker-configurations/services/comfyui-qwen/workspace/ComfyUI
git checkout f8b981ae        # restore pre-upgrade ComfyUI core
docker compose down && docker compose up -d   # recreate
```

## Scope (partial delivery — `See #3164`, not `Closes`)

Infrastructure fix only. The Vague Image notebooks themselves (`02-1-Qwen-Image-Edit-2509.ipynb` et al.) still need **re-execution on the unblocked stack** as a separate follow-up. Epic #3164 stays open.

### DIFF
--- a/docker-configurations/services/comfyui-qwen/docker-compose.yml
+++ b/docker-configurations/services/comfyui-qwen/docker-compose.yml
@@ -3,18 +3,26 @@ services:
     image: python:3.11
     container_name: comfyui-qwen
     hostname: comfyui-qwen
-    user: "1000:1000"
-    
+    # No `user:` directive: the entrypoint starts as root to chown the
+    # bind-mounted workspace (Docker Desktop Windows maps host files as
+    # root-owned), installs gosu, then drops to uid 1000 (bonsai) before
+    # launching the server. See entrypoint.sh privilege-drop block. (axe-2 #3164)
+
     deploy:
       resources:
         limits:
           cpus: '4'
-          memory: 12G
+          memory: 28G
         reservations:
           devices:
             - driver: nvidia
               device_ids: ['${GPU_DEVICE_ID:-0}']
               capabilities: [gpu]
+
+    # Shared memory for CUDA/PyTorch IPC. Default 64M causes OOM during Qwen
+    # image sampling (tensor allocator exhausts /dev/shm). 16g matches the
+    # working set observed during axe-2 #3164 generation testing.
+    shm_size: '16g'
     
     ports:
       - "127.0.0.1:${COMFYUI_PORT:-8188}:8188"
@@ -45,6 +53,14 @@ services:
       - NVIDIA_VISIBLE_DEVICES=${NVIDIA_VISIBLE_DEVICES:-0}
       - PYTHONUNBUFFERED=1
       - PYTHONDONTWRITEBYTECODE=1
+      # The server runs as uid 1000 (bonsai) after the entrypoint privilege
+      # drop. python:3.11 has no uid-1000 /etc/passwd entry, so
+      # torch._inductor's getpass.getuser() -> pwd.getpwuid(1000) would raise
+      # KeyError and abort the transformers->hqq->inductor import chain,
+      # disabling ALL Nunchaku loaders. Setting USER/LOGNAME makes
+      # getpass.getuser() return early without touching pwd. (axe-2 #3164)
+      - USER=bonsai
+      - LOGNAME=bonsai
       - TZ=${TZ:-Europe/Paris}
       - COMFYUI_PORT=8188
       - COMFYUI_LISTEN=0.0.0.0
--- a/docker-configurations/services/comfyui-qwen/entrypoint.sh
+++ b/docker-configurations/services/comfyui-qwen/entrypoint.sh
@@ -3,6 +3,38 @@ set -e
 
 echo "🚀 Démarrage de l'entrypoint ComfyUI..."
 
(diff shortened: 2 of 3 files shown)


jsboige (commented):
## Review: #3804 (Hermes, po-2026)

**Verdict**: COMMENT (self-review cap) — fix is sound, no blockers.

Read the full diff (docker-compose.yml + entrypoint.sh + requirements.txt). The root-cause analysis is correct and the fix is the standard secure pattern:

**Privilege drop** — removing `user: "1000:1000"` and instead running the entrypoint as root → chown the bind-mounted venv/custom_nodes → `exec gosu bonsai` is the textbook fix for the Docker-Desktop-maps-Windows-files-as-root problem. Server never runs as root. ✅
- **Scoped chown** (venv + custom_nodes only, not models/outputs/cache) is the right optimization — avoids walking multi-GB trees on every start.
- **USER/LOGNAME=bonsai** correctly short-circuits `getpass.getuser()` so torch._inductor doesn't hit a `KeyError` from the missing uid-1000 passwd entry. Good catch on a non-obvious failure mode.

**Resources** — memory 12G→28G and `shm_size: 16g` are justified by the documented OOM during Qwen sampling (/dev/shm exhaustion is a real PyTorch/CUDA failure). Reasonable.

**requirements.txt** — version bumps + tighter pinning (`>=`→`==` on comfy-kitchen/aimdo) + dropped PyOpenGL-accelerate. Standard dep refresh, nothing risky.

**Minor nits (non-blocking)**: `apt-get install gosu` is unpinned (acceptable for a dev container; could `--no-install-recommends` for a slightly smaller layer). No secrets — the only `secret`-matching line is a comment noting `.secrets is ro`.

No changes requested.

PR #3806, opened 2026-06-20
fix(planners-8): repair degenerate drone model — real 1.80x parallelism (Prong B #3164/#3801)

jsboige:
## Summary

SOTA **Prong B** fix (mandat user 2026-06-21, EPIC #3801, rule `sota-not-workaround.md`). The drone delivery example in Planners-8-Temporal was a **degenerate case** where the CP-SAT engine's distinctive multi-resource scheduling capability was never exercised — the committed output showed **Gain 1.00x** (makespan 27 = sequential sum), parallelism completely unused despite 2 drones. The notebook documented this honestly but left the engine unvalued.

### Root cause

The drone constraint was modeled with a buggy pairwise bool reification (`same_drone` + `i_before_j` + `AddBoolOr`) that **over-constrained** the model and prevented parallelization. The true optimum is makespan **15** (D3 on one drone 0-12, D1+D2 chained on the other 0-9-15), **Gain 1.80x**.

### Changes (cell 36 source+output, cell 37 markdown; 32 ins / 35 del)

| Cell | Change |
|------|--------|
| 36 (`cb081225`) source | Replace buggy pairwise logic with the idiomatic OR-Tools CP-SAT pattern: optional intervals per drone (`AddExactlyOne` + `NewOptionalIntervalVar`) + `AddNoOverlap` per drone. Cleaner AND correct. |
| 36 output | Re-executed: `Makespan 15 min`, `Gain 1.80x` (was 27 min / 1.00x). |
| 37 (`d829c3fa`) markdown | Rewrite the \"Pas de parallelisme\" bullet and the callout to reflect the real 1.80x gain. Robust to drone-assignment variance (multiple optima exist). |

### Verification

- **H.1**: 16/16 cells executed, 0 errors (re-executed cleanly).
- **C.3**: only cell 36 (source+output) and cell 37 (markdown) changed, 0 churn on other cells.
- **G.1**: confirmed the bug by reproducing both models standalone — old pairwise model gives makespan 27, clean `AddNoOverlap` model gives 15. The 1.80x gain is the engine's real capability now visible.
- The factory example (idx 29-32) retains its 1.83x gain as a second regime of parallelism.

This is the textbook Prong B fix: the engine (CP-SAT) now demonstrates its distinctive power (multi-resource scheduling with real parallelism) instead of degenerating to a sequential baseline.

See #3164, #3801

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3812, opened 2026-06-21
fix(sudoku-4): repair SA energy bug + degenerate instance (Prong B #3164/#3801)

jsboige:
## Summary

SOTA **Prong B** fix (mandat user 2026-06-21, EPIC #3801, rule `sota-not-workaround.md`). The Sudoku-4-Python Simulated Annealing notebook was **doubly degenerate**: a **buggy energy function** (issue #3377) masked a **degenerate instance** where the committed output showed SA doing zero work (`Energy 0.00` / `Accept 100%` / `Improve 0%` across the whole cooling trace). Classic Prong B failure: the SOTA engine (SA) was indistinguishable from a trivial baseline.

### Root cause (G.1 self-verified)

The energy function summed `counts[1:] - 1` over **all** values 1-9, including absent ones (count=0 → contributes −1). On the complete-grid invariant maintained by row-permutation init, each duplicate (+1) was exactly cancelled by a missing value (−1), so **energy = 0 for any complete grid, valid or not**. SA could never find signal to optimize. The C# twin (`Sudoku-4-Csharp`) already had the correct clamped energy and a rich demo — the Python version was the laggard.

### Changes (211 ins / 261 del, 1 file)

| Cell | Change |
|------|--------|
| 9 (`69b53f0e`) source | `np.sum(counts[1:] - 1)` → `int(np.sum(np.maximum(counts[1:] - 1, 0)))` (clamp absent values to 0). Energy now correctly counts column/block duplicates. |
| 9 + 16 + 20 + 25 + 29 + 32 outputs | Re-executed from the corrected energy. |
| 29 (`4dd820e2`) source | Benchmark `num_puzzles=5` → `3` (rich demo, bounded runtime — Prong B \"discriminant mais borne\"). |
| 10 / 21 / 30 / Conclusion (md) | Removed stale #3377 caveats; documented the real SA dynamics (29→0 trajectory) + measured benchmark results + honest SA-vs-backtracking comparison. |
| 8 / 14 / 28 (exercise TODO cells) | Added `print(\"Exercice a completer\")` to satisfy output-systematique convention (their source lacked a print; main had stale outputs that didn't match source). |

### What the output now shows (SA capability made visible)

```
Temperature   Energy   Accept   Improve
   1.00000     29.00                      <- init by permutation: columns/blocks violated
   0.93325      7.00    23.60%    9.20%   <- high T: real exploration
   0.87096     13.00     9.60%    4.00%   <- uphill moves accepted
   ...
   0.00100      0.00     0.00%    0.00%   <- low T: converged to valid solution
Solution valide: True  (now GENUINELY valid)
```

Benchmark: **3/3 solved** (was 5/5 trivial at 0.00s) — `~0.1s / ~11s / ~220s`. Honest comparison vs backtracking (~1-9ms): SA loses on speed but the comparison is now **meaningful** (before, SA \"won\" at 0ms because it did nothing).

### Verification

- **H.1**: 18/18 cells executed, 0 errors, exec_count sequential 1..18.
- **C.2**: `check_c2_compliance.py` = 1/1 compliant (was failing on 3 exercise cells whose stale outputs didn't match source).
- **C.3**: only energy cell (source+output), benchmark cell (source+output), 4 interpretation markdown cells, and 3 exercise-cell prints changed. Catalog byte-identical to main.
- **G.1**: reproduced both energy functions standalone — buggy returns 0 on the init grid, corrected returns 29. Confirmed the degeneracy independently of the audit.

This is the textbook Prong B fix: the engine (SA) now demonstrates its distinctive power (landscape exploration via probabilistic uphill moves, escaping local minima) instead of degenerating to a baseline that solves nothing.

See #3164, #3801

### DIFF
--- a/MyIA.AI.Notebooks/Sudoku/Sudoku-4-SimulatedAnnealing-Python.ipynb
+++ b/MyIA.AI.Notebooks/Sudoku/Sudoku-4-SimulatedAnnealing-Python.ipynb
@@ -332,7 +332,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 4,
    "metadata": {},
    "outputs": [
     {
@@ -349,12 +349,13 @@
     "    # TODO: Compter les conflits (doublons) uniquement dans les\n",
     "    # lignes/colonnes/blocs qui contiennent des valeurs non nulles\n",
     "    result = 0  # TODO etudiant\n",
-    "    return result\n"
+    "    return result\n",
+    "print(\"Exercice a completer\")\n"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 5,
    "id": "69b53f0e",
    "metadata": {
     "execution": {
@@ -378,7 +379,7 @@
      "output_type": "stream",
      "text": [
       "Fonction compute_energy definie.\n",
-      "Energie de la grille de test: -72\n"
+      "Energie de la grille de test: 0\n"
      ]
     }
    ],
@@ -391,14 +392,14 @@
     "    for col in range(9):\n",
     "        values = grid[:, col]\n",
     "        counts = np.bincount(values[values > 0], minlength=10)\n",
-    "        energy += np.sum(counts[1:] - 1)  # -1 car la valeur presente compte 1\n",
+    "        energy += int(np.sum(np.maximum(counts[1:] - 1, 0)))  # doublons uniquement (clamp: valeur absente = 0)\n",
     "    \n",
     "    # Doublons dans les blocs 3x3\n",
     "    for box_row in range(3):\n",
     "        for box_col in range(3):\n",
     "            block = grid[box_row*3:(box_row+1)*3, box_col*3:(box_col+1)*3].flatten()\n",
     "            counts = np.bincount(block[block > 0], minlength=10)\n",
-    "            energy += np.sum(counts[1:] - 1)\n",
+    "            energy += int(np.sum(np.maximum(counts[1:] - 1, 0)))  # doublons uniquement (clamp: valeur absente = 0)\n",
     "    \n",
     "    return energy\n",
     "\n",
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** — logic review · CoursIA #3812 (Sudoku-4 SA energy + degenerate-instance fix)

**Verdict:** LGTM (0 blocking, 2 minor non-blocking)

**Scope:** 1 notebook (`Sudoku-4-SimulatedAnnealing-Python.ipynb`), +211/-261. Only ~6 source LOC actually changed (1 energy fix × 2 sites, 3 exercise prints, 1 benchmark arg); the −261 is re-execution of the SA trace + honest markdown rewrites. Same Prong B family as #3806.

**Energy bug:** **Real bug, correct fix.** I reproduced it independently.
- OLD (cell `69b53f0e`): `energy += np.sum(counts[1:] - 1)` summed `count−1` over all values 1–9, **including absent ones** (count=0 → contributes −1). On any complete unit, every `+1` from a duplicate is exactly cancelled by a `−1` from a missing value, so the function returns 0 for any complete unit, valid or not. I verified numerically: a unit `[1,1,2,3,4,5,6,7,8]` (one dup of 1, missing 9) gives OLD=0, NEW=1. A full invalid grid I constructed (rows = permutations, but boxes violated) gives OLD=0, NEW=18. The OLD test-grid output in main was `Energie de la grille de test: -72` (negative — smoking gun), the NEW is `0`.
- NEW: `energy += int(np.sum(np.maximum(counts[1:] - 1, 0)))` clamps absent values to 0, so only duplicates contribute. This is the correct standard Sudoku SA energy. Counts all 18 units that can bear violations (9 cols + 9 boxes); rows are valid by construction (permutation init), so omitting them is correct, not a missing-constraint bug. Matches the C# twin.

**Degenerate instance:** **Genuine repair of the *demo*, not a swap of the puzzle.** Important nuance: the PR title says "degenerate instance" but the **puzzle instance itself is byte-identical to main** — same `SudokuGrid.from_string` load, same clues. What was degenerate was the *committed output trace*: main's cell 10 showed `Energy 0.00 / Accept 100% / Improve 0%` for the entire cooling schedule — i.e. SA did literally zero work (a direct consequence of the energy bug: with E≡0, every move is "non-worsening" and accepted, but nothing is ever optimized). The fix re-executes the same puzzle under the corrected energy and the trace now shows real SA dynamics. This is the right kind of fix: the *problem* isn't dressed up, the *engine* is unmasked. (A purist could argue the title slightly overclaims — there's no new puzzle — but the body is precise about root cause and the artefact is correct.)

**Convergence:** **Final grid genuinely valid.** I parsed the cell 10 final solution and verified independently: all 9 rows, 9 cols, and 9 boxes contain {1..9} with no repeats (all 27 units valid). It also respects every non-zero clue of the input puzzle (36 fixed cells, all match). NEW energy of the final grid = 0. The trace shows textbook SA: `E_init=29` (the permutation-completed grid violates cols/boxes), then `29→7→13→9→7→4→2→0→2→0→...→0` with `Accept≈23%` / `Improve≈9%` at high T collapsing to `0%`/`0%` at low T. There is a real **uphill move** at T=0.87096 (E went 7→13) — the probabilistic-acceptance signature that distinguishes SA from greedy descent. Cooling ratio between consecutive logged points is `0.93325 = 0.999^69`, consistent with the declared `alpha=0.999, iterations_per_temp=100`. Not hand-pasted.

**Output integrity:**
- 18 code cells, `execution_count` sequential `1..18` (H.1 pass).
- 0 error outputs.
- SA trace numbers are non-round and internally consistent (11.39s solver vs 0.15s manual vs 232.13s benchmark total — all coherent).
- Benchmark `num_puzzles=5→3` is the only behavioural knob change; honest Prong B trade-off (rich demo, bounded runtime — the 3rd puzzle takes ~221s). Old "5/5 at 0.00s" was the degenerate give-away; new "3/3 at 0.14s/11s/221s" is real.
- Markdown rewrites (cells 10/21/30/40) are honest: they document the correct energy semantics, the 29→0 trajectory, and explicitly note SA loses to backtracking on speed (~1–9 ms vs up to ~220 s). No overclaiming.

**Verified:**
- Energy bug reproduced standalone: OLD returns 0 for an invalid complete unit, NEW returns the correct duplicate count. Fix is the standard clamped-energy formulation.
- Final solution grid parsed and checked: all 27 Sudoku units valid; all 36 puzzle clues respected; NEW energy = 0.
- SA cooling schedule matches `alpha=0.999`; uphill move observed at T=0.87096 (genuine exploration).
- `execution_count` 1..18 sequential, 0 errors.
- Secret/token scan: 0 hits.
- Code delta is minimal and surgical (6 LOC of real code); the −261 is mostly the replaced SA trace + rewritten markdown, all reflows of the same corrected execution.
- Stale #3377 caveat notes correctly removed from cells 10/21/30/40 (the bug they warned about is now fixed).

**Concerns:**
1. *(Non-blocking, cosmetic)* PR title says "degenerate instance" but the puzzle instance is unchanged — what was degenerate was the *output trace* caused by the energy bug. The body is precise about this ("doubly degenerate: a buggy energy function masked a degenerate instance"), but a reader scanning the title may expect a puzzle swap that didn't happen. Consider "repair SA energy bug + re-execute degenerate trace" — but this is documentation polish, not a code issue.
2. *(Non-blocking)* Benchmark `num_puzzles=3` is hardcoded at the call site while the function signature still defaults to `num_puzzles: int = 5`. Minor — a student running `benchmark_sa(easy_puzzles)` without args would wait ~15 min for the 4th/5th puzzles. Not wrong, just a slight friction point for re-execution. The inline comment does flag "bounded runtime".

— NanoClaw

PR #3913, opened 2026-06-22
security(genai): fix root cause — NotebookMaker prints API key material (not just redact outputs) (#3911)

jsboige:
## Pourquoi (mandat user 2026-06-22)

Le scrub #3125 (PR #3903, commit `88465901`) a **redacté les fragments de clé dans les outputs committés** mais a laissé **intactes les cellules source qui impriment la clé** — fix-symptôme : prochaine ré-exécution → re-leak.

> Verbatim user : « Si des scripts produisent des snippets de clés, les "redacter" n'est pas la solution. Ce sont les scripts qui doivent être corrigés. »

Cette PR corrige **la cause racine**.

## Cause racine (G.1 firsthand, evidence-cited)

Famille `GenAI/SemanticKernel/NotebookMaker` (3 notebooks) :

| Source | Pattern fautif | Fix |
|--------|----------------|-----|
| `print(f"... = {new_api_key}")` | imprime la **clé entière** (widget) | masquée → `configuree` / `(non configuree)` |
| `print(f"... = {openai_api_key[0:10] + '...'}")` | imprime les **10 premiers chars** (= `sk-F6ZJntH`, fragment redacté par #3903) | masquée |
| NB10 `.env` write `{new_api_key[0:10]+'...'}` | écrit une clé **tronquée** (bug séparé) | écrit la clé entière |

**Pattern de masquage déjà présent dans le repo** (jamais appliqué ici) : `01-SemanticKernel-Intro` (`'Oui'/'Non'`), `QC-Py-26` (`'***'+key[-4:]`).

## G.1 — leak supplémentaire trouvé et corrigé

Les outputs committés de **10a/10b** portaient un **2e préfixe de clé réel `sk-proj-Ap...`** que #3125/#3903 n'avait **jamais attrapé** (ce scan ne cherchait que `sk-F6ZJn`). Masquer la source supprime **tous** ces leaks structurellement — la démonstration exacte du point user.

## Vérifications

- 3 JSON valides
- **0** fragment `sk-` (source OU outputs) post-fix
- Outputs committés alignés (C.2 ; notebooks widget non ré-exécutables headless → alignement chirurgical source+output)
- Diff chirurgical **12/12** lignes, 3 notebooks ; catalogue + submodules byte-identiques

## Process (durcissement bots — voir #3911)

Une PR qui REDACTE un secret en output DOIT corriger/tracker le script source, sinon CHANGES_REQUESTED.

Closes #3911. See #3125.

[...]

PR #3917, opened 2026-06-22
scrub(archive-tweety): anonymize checkout-root paths in outputs (#3436)

jsboige:
## Summary

Scrub the `C:\dev\CoursIA\` checkout-root path leaks in `archive/Tweety.ipynb` (my po-204 .NET residual grain, gated on #3895 which merged as `2aee662f`).

## Root-cause lens (per user mandate 2026-06-22 — "fix the scripts, don't just redact")

I verified firsthand (G.1) where the leak comes from: the `C:\dev\CoursIA\` paths appear **ONLY in stdout stream outputs (cells 6/7/10), NEVER in source**. The notebook prints its runtime location (`Chemin du dossier libs : …`, `Outils externes dans '…/ext_tools'`) where the dir is **dynamically resolved** (not hardcoded). This is **environmental noise** — the same class as ipykernel temp paths, .NET nuget-load lines, pip "already satisfied" — that re-execution regenerates afresh each run.

Per the user's mandate, scrubbing is the **correct root-cause fix here**, NOT a redaction-symptom. This is distinct from the key-leak case (NotebookMaker #3913 / #3911) where the source *deliberately prints secret material* — that requires a source fix. Here there is no secret-printing source to fix; the path reveals only the machine's directory structure, resolved at runtime.

## Verification (corrected tool, post-#3895 lookbehind fix)

- Re-scan: **0 residual**
- 0 mangle (`<X<repo>` = 0; the lookbehind fix preserves any URL scheme — none here anyway, plain paths)
- JSON valid
- Diff: **7/7 path-only**, 0 source / cell_type / execution_count change
- Catalogue + submodule byte-identical (only `Tweety.ipynb` touched)

\`See #3436\`

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/archive/Tweety.ipynb
+++ b/MyIA.AI.Notebooks/SymbolicAI/archive/Tweety.ipynb
@@ -309,7 +309,7 @@
       "  JARs Modules:\n",
       "    - Nouveaux téléchargés : 0\n",
       "    - Total présents       : 32 / 32\n",
-      "  Chemin du dossier libs : C:\\dev\\CoursIA\\libs\n",
+      "  Chemin du dossier libs : <repo>libs\n",
       "\n",
       "✔️ Vérification des JARs Tweety terminée. Le démarrage de la JVM peut continuer.\n"
      ]
@@ -471,7 +471,7 @@
      "text": [
       "\n",
       "--- 1.4bis Téléchargement des Fichiers de Données ---\n",
-      "ℹ️ Les fichiers de données seront téléchargés dans: C:\\dev\\CoursIA\\resources\n",
+      "ℹ️ Les fichiers de données seront téléchargés dans: <repo>resources\n",
       "\n",
       "Vérification/Téléchargement de 29 fichiers de données...\n"
      ]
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Verified the scrub end-to-end on the raw post-scrub notebook (350 KB, fetched via `raw.githubusercontent.com`).

**Scope is correct and tight**
- Single file, +7/−7, output-stream text only (cells 6, 7, 10). Diff shows every hunk is a stdout line, never source.
- Structural integrity confirmed by parsing both sides: 72 cells both sides, 0 cell_type changes, 0 source changes, 0 execution_count changes, metadata + nbformat_minor byte-identical. JSON valid both sides.

**Scrub is complete and consistent**
- Base file has exactly 7 occurrences of `C:\dev\CoursIA\` (JSON-doubled form), all in `outputs[].text`, none in `source`. Post file has exactly 7 `<repo>` tokens in the same positions, 0 residual leak instances.
- Replacement token (`<repo>`) is uniform — no `<X<repo>` mangle artifacts from an over-greedy lookbehind.
- No other checkout-root variants slipped through: no `D:\dev\...`, no `users\...` home paths, no `jsboige` username anywhere in the post-scrub file.
- The only remaining Windows drive-path is `C:\Program Files\Java\jdk-17` in a markdown instruction telling students where to install Java — that's correct pedagogical content, not a leak, and was rightly left untouched.

**Root-cause framing is sound**
The PR description's distinction between this case (path dynamically resolved at runtime, printed to stdout — environmental noise) and the NotebookMaker key-leak class (source deliberately prints secret material) holds up: there is no secret-printing source here to fix, so redacting the output is the correct fix, not a symptom patch.

Pedagogical intent fully preserved — outputs still show the runtime structure, just with the machine-local root anonymized. Verdict: **LGTM**.

PR #3997, opened 2026-06-23
feat(lean): prove PR1 witness excluded under Reidemeister1Connected (#2874, Stage 1/C)

jsboige:
## Summary

Stage 1 of ai-01's **mandate (C)** for #2874 (cable `Reidemeister1Connected` into `ReidemeisterStep`/`ReidemeisterEquiv`). This PR is a **build-green additive non-regression gate**: it certifies that the disjoint-kink counter-example witness — the pair that refutes `tricolorable_invariant` (Invariant.lean L116) under the free-ρ `Reidemeister1` model (`tricolorable_invariant_fails_under_pr1_model`, L211-284) — is **also unreachable under the connected R1 move (option C)**, not only under the ρ-determined `Reidemeister1'` model (proven L317).

### New lemma — `Knots/Invariant.lean` §3c-bis

```lean
theorem pr1_counterexample_excluded_under_connected :
    ¬ Reidemeister1Connected
        { crossings := [⟨1, 2, 1, 2⟩], numEdges := 2, hwell := by trivial }
        { crossings := [⟨1, 2, 1, 2⟩, ⟨3, 4, 3, 4⟩], numEdges := 4, hwell := by trivial }
```

**Why option C excludes it.** `Reidemeister1Connected` (Reidemeister.lean L224) requires the appended kink to be `⟨a, n+1, n+2, n+2⟩` with `1 ≤ a ≤ d₁.numEdges` a REAL arc of `d₁`. For the witness (`d₁.numEdges = 2`), surgery forces `d₂`'s last crossing `⟨3,4,3,4⟩` to equal `⟨a,3,4,4⟩`; the **e2 field projection** (`congrArg PDCrossing.e2`) gives `4 = 3` — a direct structural contradiction, independent of `a`'s value. The witness pair is unreachable under any connected R1 model: the twist must splice a real arc, and `d₁`'s sole crossing has no arc labelled `3`. **This is what makes option C the honest SOTA fix rather than the (X) reframe**: the refuting witness vanishes under the correct equivalence by construction.

## Lean PR discipline (§B)

| # | Element | Evidence |
|---|---------|----------|
| 1 | `grep -c sorry` avant/après | Invariant.lean: **6 → 6** (5 real + 1 prose; my lemma adds **0 sorry**). CI gate `lean-knot.yml` baseline **27 unchanged**. |
[...]

PR #3999, opened 2026-06-23
feat(lean): wire Reidemeister1Connected into ReidemeisterStep (Knot #2874 Stage 2)

jsboige:
## Summary

**Stage 2 of mandate (C)** — wire `Reidemeister1Connected` into `ReidemeisterStep` / `ReidemeisterEquiv`, so the target invariant `tricolorable_invariant` stands on the geometrically-connected move instead of the unsound free-ρ `Reidemeister1`.

### The change

`ReidemeisterStep.r1` now requires the **symmetric closure** of the connected move:

```lean
| r1 {d'} :
    (Reidemeister1Connected d d' ∨ Reidemeister1Connected d' d) →
    ReidemeisterStep d d'
```

(was `Reidemeister1 d d'`). The `∨` gives symmetry for free via `Or.symm`, so `reidemeister_equiv_symm`'s `r1` branch keeps `h.symm` byte-identical (it now resolves to `Or.symm` over the disjunction rather than `Reidemeister1.symm`) — no `Reidemeister1Connected.symm` lemma and no separate untwist arm needed.

**Why this closes the soundness hole.** The free-ρ `Reidemeister1` model carries the new crossing `c` and the edge-relabeling `ρ` as two independent existentials, leaving `ρ` decoupled from `c`'s labels — so a disjoint kink can create tricolorability out of nothing. `Reidemeister1Connected` splices into an EXISTING arc `a` of `d₁`, tying the fresh edges to `color a` via Fox, which is what makes tricolorability transfer along the move. The equivalence (reflexive-transitive closure of `ReidemeisterStep`) now propagates only sound moves.

### Reconciliation of the certified counter-example (anti-regression care)

The free-ρ counter-example `tricolorable_invariant_fails_under_pr1_model` (§3b) used to be stated as `ReidemeisterEquiv d₁ d₂ ∧ ¬ IsTricolorable d₁ ∧ IsTricolorable d₂`. After the rewire that statement would be **FALSE** (the witness pair is no longer `ReidemeisterEquiv`-reachable). Rather than delete it, it is **restated on the raw move** `Reidemeister1 d₁ d₂` (option β, decoupled from `ReidemeisterStep`):

- the witness proof is unchanged in content (only the `ReidemeisterEquiv.step (ReidemeisterStep.r1 ?_)` wrapper is dropped — the same `⟨⟨3,4,3,4⟩, ⟨ρ, surgery⟩⟩` witness now provides `Reidemeister1` directly);
- the theorem now documents **why the free-ρ model needed the rewire** — its witness pair is provably NOT a `Reidemeister1Connected` move (the exclusion lemma `pr1_counterexample_excluded_under_connected`, §3c-bis, **PR #3997** pending merge), so it no longer refutes the equivalence-based invariant.

Consequently `tricolorable_invariant` (§2) is **no longer REFUTED** by this witness — it now stands on the sound connected equivalence. It remains `sorry` (an open proof obligation, the forward fox/col transfer — §3e / #3003), but it is no longer known-false.

## Build-gate proofs (garde-fou anti-reg D)

- **`lake build` SUCCESS — 321 jobs, 0 errors** (native Windows `lake.exe`, Lean 4.31.0-rc1). `reidemeister_equiv_symm` compiles with the rewired `r1` branch (`h.symm` → `Or.symm`, verified).
- **Sorry baseline 27 preserved** (CI `prose-header` filter, `lean-knot.yml`). Verified by running the workflow's exact count on the working tree: **27** (= origin/main, 0 regression). 0 tactic sorries added or removed — the only `sorry` token in the diff is the pre-existing `exact sorry` on `tricolorable_invariant`.
- **No proven content deleted** — the counter-example witness and its 3 sub-proofs (wf on both diagrams, ρ: Fin 2 ↪ Fin 4, twist surgery; `¬ IsTricolorable d₁`; `IsTricolorable d₂`) are byte-identical; only the statement wrapper changed.
- Blast radius mapped before editing: 3 code sites (`ReidemeisterStep.r1` def, `reidemeister_equiv_symm` r1 branch, the §3b theorem statement) + prose sites audited (the `STILL REFUTED` claim in the `tricolorable_invariant` sorry-block is now corrected to the honest "no longer refuted / open forward transfer" status).

## Related

- Sibling PR **#3997** (Stage 1/C) — the exclusion lemma `pr1_counterexample_excluded_under_connected` that justifies this rewire's soundness. Independent and additive; this PR cites it as justification but does not depend on its code.
- See #2874 (epic), mandate (C). Not a close — the forward transfer (#3003) and Conway P5 remain open.

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/knot_lean/Knots/Invariant.lean
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/knot_lean/Knots/Invariant.lean
@@ -114,26 +114,23 @@ theorem tricolorable_invariant :
       ReidemeisterEquiv d₁ d₂ →
       IsTricolorable d₁ ↔ IsTricolorable d₂ := by
   exact sorry
-  -- BLOCKED (Phase 5 — model STILL too coarse after PR1). PR1 (#2929)
-  -- strengthened the move model with `KnotDiagram.wf` on both diagrams and an
-  -- edge-renaming `ρ`. That REMOVED the Phase-3 obstruction (malformed witness
-  -- `⟨7,8,9,10⟩`, excluded by `wf`), but it did NOT make this statement true.
-  -- A SECOND certified counter-example — `tricolorable_invariant_fails_under_pr1_model`
-  -- below (§3b) — proves the biconditional is STILL REFUTED under the PR1 model.
+  -- BLOCKED (forward transfer, Phase 5 PR2). `ReidemeisterStep.r1` was rewired
+  -- (Stage 2, #2874) to the GEOMETRICALLY CONNECTED move `Reidemeister1Connected`,
+  -- so the free-ρ counter-example of §3b is no longer `ReidemeisterEquiv`-reachable
+  -- (it is provably NOT a connected move, §3c-bis / PR #3997). The invariant is
+  -- therefore NO LONGER REFUTED by that witness — it now stands on the sound
+  -- connected equivalence. It is still OPEN: the FORWARD direction is unproven,
+  -- i.e. a tricoloring of `d₁` must EXTEND across a connected R1 curl
+  -- (`Reidemeister1Connected`), so the two fresh edges inherit `color a`.
   --
-  -- Root cause (the real blocker): `wf`'s "every label appears exactly twice"
-  -- condition forces an R1-twist's new crossing `c` to use ONLY the two fresh
-  -- edges `{n+1, n+2}` (labels `1..n` already appear twice in `d₁`, so `c`
-  -- cannot reuse any without breaking the parity), and `ρ` is a FREE injection
-  -- not tied to `c`'s labels. The new crossing's Fox condition is therefore
-  -- DECOUPLED from `d₁`'s coloring — a twist can CREATE tricolorability from
-  -- nothing (or hide the ≥2-colours entirely in the fresh edges while `d₁` is
-  -- forced monochrome). The fix is NOT a transfer lemma but a STRONGER
-  -- constructor model in which `ρ` DETERMINES `c`'s labels: a genuine R1 curl
-  -- on arc `a` is `⟨a, a, n+1, n+2⟩` (one strand is the EXISTING arc `a`),
-  -- whose Fox condition ties the new edges to `color a`. See §3b for the
-  -- witness and the Phase-5 prerequisite. Reference: Fox (1962); Adams, "The
-  -- Knot Book".
+  -- Historical diagnosis (why the OLD free-ρ `Reidemeister1` model failed):
+  -- `wf`'s "every label appears exactly twice" condition forced an R1-twist's new
+  -- crossing `c` to use ONLY the two fresh edges `{n+1, n+2}` (labels `1..n`
+  -- already appear twice in `d₁`), and `ρ` was a FREE injection not tied to `c`'s
+  -- labels. The new crossing's Fox condition was therefore DECOUPLED from `d₁`'s
+  -- coloring — a twist could CREATE tricolorability from nothing. The connected
+  -- move fixes this by splicing into an EXISTING arc `a`, tying the fresh edges
+  -- to `color a` via Fox. Reference: Fox (1962); Adams, "The Knot Book".
 
 /-! ## 3. The trefoil is tricolorable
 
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Reviewed the full diff (Invariant.lean, Reidemeister.lean). Honest, scoped review — I cannot run Lean here, so I verified structurally by reading the referenced definitions in both files at HEAD.

## Blocking: `reidemeister_equiv_symm` no longer type-checks

The `r1` constructor was changed from
```
| r1 {d'} : Reidemeister1 d d' → ReidemeisterStep d d'
```
to
```
| r1 {d'} :
    (Reidemeister1Connected d d' ∨ Reidemeister1Connected d' d) →
    ReidemeisterStep d d'
```
i.e. the argument is now an `Or` (disjunction), not a single move predicate.

But `Reidemeister.lean` still contains (unchanged by this PR):
```
| r1 h => exact ReidemeisterStep.r1 h.symm
```
in `reidemeister_equiv_symm` (line 511). After this PR:
- `h : Reidemeister1Connected d d' ∨ Reidemeister1Connected d' d` — an `Or`, which has **no `.symm` field**. `h.symm` is invalid.
- Worse, there is no `Reidemeister1Connected.symm` lemma anywhere in the file (only `Reidemeister1.symm`, `Reidemeister2.symm`, `Reidemeister3.symm` exist). So the symmetry of the connected move is not proven, and the `r1` branch of the symm proof has nothing to call.

This breaks `reidemeister_equiv_symm` and consequently `reidemeister_equiv_equivalence` (which is `symm := reidemeister_equiv_symm`). Since `ReidemeisterEquiv` is the central equivalence the whole invariant file rests on, this is a compile failure on a headline theorem, not a latent issue.

To land this, this PR must either:
1. add `theorem Reidemeister1Connected.symm` proving the connected move is symmetric in its diagram arguments, and rewrite the `r1` branch as `| r1 h => exact ReidemeisterStep.r1 (Or.inr h.symm) <or> (Or.inl h.symm)` (case-splitting the `Or`), or
2. restructure so `ReidemeisterStep.r1` carries the connected move in a single direction with a separate reverse-step constructor, avoiding the disjunction.

## Note (non-blocking): design of the disjunction in `r1`

Encoding direction as `Reidemeister1Connected d d' ∨ Reidemeister1Connected d' d` inside the constructor is slightly unusual — it makes every consumer of `r1` case-split on `Or`, and it duplicates the directionality that `ReidemeisterEquiv.symm` is supposed to handle. A constructor `| r1c {d'} : Reidemeister1Connected d d' → ReidemeisterStep d d'` (single direction) plus the symm proof is the more idiomatic Lean shape and would sidestep the breakage above. Not a blocker if the disjunction is intentional, but flagging it.

## What I verified and could not verify

- **Verified:** `Reidemeister1Connected` exists (Reidemeister.lean:224), the new `r1` references it with the right arity, and the Invariant.lean changes to `tricolorable_invariant_fails_under_pr1_model` are consistent — the statement was correctly downgraded from `ReidemeisterEquiv d₁ d₂` to `Reidemeister1 d₁ d₂`, and the witness proof's `(a)` branch produces exactly the `Reidemeister1` structure (`⟨?_, ?_, ⟨⟨3,4,3,4⟩, ⟨?_, ?_⟩⟩⟩`). The reworded comments are accurate about which move the counter-example now refutes.
- **Not verified:** the actual mathematical claim that the connected splice preserves tricolorability (the forward-transfer lemma the comments say is still OPEN, #3003). That is out of scope for this wiring PR and correctly marked as such.
- **The `exact sorry` at the `tricolorable_invariant` theorem is pre-existing** (the BLOCKED placeholder), not introduced here — fine, and the updated comment correctly reflects the new state (no longer refuted by the §3b witness, forward direction still open).
- **Security:** no secrets/tokens/keys in the diff. No new `sorry`/`admit`/debug code added.

## Verdict

The Invariant.lean rewording + statement downgrade is correct and internally consistent. But the Reidemeister.lean constructor change breaks an existing headline theorem (`reidemeister_equiv_symm`) by leaving a dangling `h.symm` on an `Or`, with no `Reidemeister1Connected.symm` lemma to back it. This needs the symm lemma + branch rewrite before it can compile. COMMENT only — not requesting changes, author's call on the fix shape.

PR #4002, opened 2026-06-23
feat(search): MGS-7-TSP Config 3 — geometric crossover on permutations (Moraglio) #3964

jsboige:
## Summary

Third configuration for **MGS-7-TSP**, extending the geometric framework (DE/WOA/EO/FBI on continuous `double` genes, MGS-5) to **permutation** genes — `#3964` (B1).

**Config 3** wires `GeometricCrossover<int>` with `OrderedEmbedding<int>` (`IsOrdered=true`) + `ReverseSequenceMutation`. The offspring walks by swaps (`FlipGene`) along the **swap-distance metric segment** between parents — Moraglio's geodesic crossover — and is always a valid permutation.

### The pedagogical point (prong-B non-triviality)

The default centroid operator acts on **city-index labels**, not coordinates — so `(v₁ + v₂)/2` has no geographic meaning. The geometric crossover therefore **underperforms** OX1:

| Config | Tour final | vs baseline |
|---|---|---|
| Baseline (panmictic, OX1) | **440.8** | — |
| Islands (4×10, Static) | 513.9 | −16.6 % |
| **Geometric (Moraglio)** | **674.1** | **−52.9 %** |

This is exactly the limit that **geometric DL theory** predicts and which motivates permutation-equivariant embeddings (Bronstein *et al.*, *Geometric Deep Learning*). The naive centroid-on-labels is the **baseline that measures the predicted limit** — not a workaround, a deliberate motivating demonstration of where the SOTA abstraction already reaches, and where it needs a learned/positional embedding.

### Changes (atomic — MGS-7-TSP only)
- **+2 cells**: theory intro (Moraglio & Poli GECCO 2004 / thesis 2007; swap distance; `OrderedEmbedding` swap-walking; geometric-DL pivot) + geometric run.
- Comparison table + Lecture nuance + Conclusion extension + References — all theory-neutral, numbers from one coherent run (no pendulum).
- `#2161` met: 3 exercises preserved (cells 16/18/20); Config 3 is a guided example, not an exercise.

### Validation (H.1 — real execution)
- Full **.net-csharp papermill re-exec**: **10/10 code cells, 0 errors**.
- No `raise NotImplementedError` / `assert False` / `1/0` (C.1).
- Outputs committed (C.2); only `metadata.papermill` input/output_path normalized to basename (#3962). Zero machine paths / secrets in any cell output (scanned).
- No submodule bump (`GeometricCrossover` already in the pinned fork @ `f7ec022c`), no catalog churn (byte-identical to main).

### Geometric crossover theory
The fork abstraction + acceptance tests (`GeometricCrossoverTests.cs`: `OrderedEmbedding_Ordered_PreservesPermutationMultiset`, `FlipGene_SwapsTwoGenes`) already exist (PR#87 port). This PR is the **demonstration** that the same `GeometricCrossover` serving the continuous compounds applies to permutations via the embedding — the geometry lives in the **metric/embedding**, not the operator.

`See #3964` (partial contribution to the MGS extension epic; fork abstraction land separately).

### DIFF
--- a/MyIA.AI.Notebooks/Search/Part4-Metaheuristics/MGS-7-TSP.ipynb
+++ b/MyIA.AI.Notebooks/Search/Part4-Metaheuristics/MGS-7-TSP.ipynb
@@ -5,10 +5,10 @@
    "id": "bafc5427",
    "metadata": {
     "papermill": {
-     "duration": 0.003449,
-     "end_time": "2026-06-14T18:22:23.443937+00:00",
+     "duration": 0.006353,
+     "end_time": "2026-06-23T03:15:15.773789+00:00",
      "exception": false,
-     "start_time": "2026-06-14T18:22:23.440488+00:00",
+     "start_time": "2026-06-23T03:15:15.767436+00:00",
      "status": "completed"
     },
     "tags": []
@@ -32,16 +32,16 @@
      "language": "csharp"
     },
     "execution": {
-     "iopub.execute_input": "2026-06-14T18:22:23.465364Z",
-     "iopub.status.busy": "2026-06-14T18:22:23.455875Z",
-     "iopub.status.idle": "2026-06-14T18:22:25.292293Z",
-     "shell.execute_reply": "2026-06-14T18:22:25.287356Z"
+     "iopub.execute_input": "2026-06-23T03:15:15.793601Z",
+     "iopub.status.busy": "2026-06-23T03:15:15.785667Z",
+     "iopub.status.idle": "2026-06-23T03:15:17.748093Z",
+     "shell.execute_reply": "2026-06-23T03:15:17.743447Z"
     },
     "papermill": {
-     "duration": 1.843692,
-     "end_time": "2026-06-14T18:22:25.293014+00:00",
+     "duration": 1.97063,
+     "end_time": "2026-06-23T03:15:17.748211+00:00",
      "exception": false,
-     "start_time": "2026-06-14T18:22:23.449322+00:00",
+     "start_time": "2026-06-23T03:15:15.777581+00:00",
      "status": "completed"
     },
     "tags": []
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** — DEEP review (algorithm logic), PR #4002, MGS-7-TSP Config 3.

Reviewed: the single touched notebook (`MyIA.AI.Notebooks/Search/Part4-Metaheuristics/MGS-7-TSP.ipynb`, +210/−117) AND the fork source it depends on — `GeometricCrossover.cs`, `OrderedEmbedding.cs`, `GeometricExtensions.cs` (`FlipGene`), and `GeometricCrossoverTests.cs` at pin `f7ec022c` — because the pedagogical claims rest on the embedding's behavior, not just the notebook text. 10/10 code cells re-executed (papermill, no fake outputs, no `NotImplementedException`/`assert False`/`1/0`). Security scan clean (no AKIA/`ghp_`/`sk-`/private-key/password hits); no machine paths leaked.

## Algorithmic verdict — the core claim is correct

**"Always a valid permutation" — TRUE.** Traced `MapFromGeometryOrdered` end-to-end. The offspring starts as `parents.First().Clone()` (a permutation) and is mutated *only* by `FlipGene(i, indexVal)`, which is a clean two-position swap (`GetGene`/`ReplaceGene` both ends). A swap preserves the multiset invariant, so by induction the offspring is a permutation at every step. Confirmed across 5 trace scenarios including the degenerate all-same-target case and a real integer-centroid of two permutations. The acceptance test `OrderedEmbedding_Ordered_PreservesPermutationMultiset` asserts exactly this (multiset equivalence, not target-equality) and passes.

**Edge cases — fine for course scale.** Identical parents → no swaps fire (`replacement == values[i]` for every index), offspring == clone. Single element → trivially valid. `n=20` (the notebook's TSP) → O(n) swaps per crossover, well within budget. No pathological complexity.

## The one substantive finding — wording overclaims "reaches the centroid"

The notebook markdown states the offspring "marche par swaps **vers la cible**" (the computed metric centroid) and is "un point du segment swap-distance entre les parents." The first half is **loosely worded**: `offspringIndexes` is a snapshot built once and **never refreshed after each swap**. Once a swap moves a value, subsequent lookups for that value return the *stale* position, so the walk does not reliably converge to the precomputed centroid target.

Concrete trace, parent `[0,1,2,3]`, target `[2,0,3,1]` (the test case itself):
- Walk produces offspring `[1,0,3,2]`, **not** the target `[2,0,3,1]`.

This is **not a bug for the pedagogical purpose** — the offspring is still a valid permutation and still lies within a swap-distance ball of `parentA` (so "on a metric segment between the parents" holds in the loose sense). But "vers la cible" / "le centroïde métrique" overstates the precision. Suggest a one-line footnote acknowledging the walk approximates the centroid direction rather than landing on it, so students tracing the code don't conclude the implementation is broken when offspring ≠ target. (For reference: this is faithful to the upstream PR#87 behavior — the stale map is inherited, not introduced here.)

## Minor nits (non-blocking)

- **Dead defensive code** in `OrderedEmbedding.cs`: `if (indexVal < 0) throw ...` cannot trigger — `offspringIndexes[values[i]]` throws `KeyNotFoundException` for an absent value before any `< 0` comparison, and all present values map to indices `≥ 0`. Harmless but misleading; a `TryGetValue` would express the intent.
- **`output_path` normalized** from the wrong `MGS-6-TSP.output.ipynb` to `MGS-7-TSP_output.ipynb` — this is a fix, good.
- The −52.9 % result for Geometric vs baseline is **honest and theoretically motivated** (centroid on arbitrary city-index labels has no geographic meaning). The PR frames this correctly as the predicted limit, not a bug — good pedagogical framing.

## Conclusion

Sound for merge. The keystone correctness property (valid permutations) is proven by the acceptance tests and re-verified by trace. The only real issue is a wording precision nit about centroid-reach that a one-line footnote would resolve; it does not affect validity of any produced tour. Comment-only (cluster self-review policy).

PR #4065, opened 2026-06-23
fix(prover): #1453 ban consecutive compile() without edit (forensic P1+P2)

jsboige:
## Summary
Forensic ping-pong on the Lean prover harness (Epic #1453, 2026-06-23) found its dominant pathology: the TacticAgent treats `compile()` as a status poller. On a verified Bondareva trace it called `compile()` 6x consecutively with **zero** file edits in between (116s = 32.6% of the run wasted). Across **24 post-baseline runs there were ZERO real sorry reductions**.

## Changes
- **P1 (`instructions.py`)** - HARD directive in `TACTIC_AGENT_INSTRUCTIONS` forbidding back-to-back `compile()` without an intervening file edit. `compile()` is an edit-verifier, not a status poller; on `STAGNATION` the agent must edit before re-compiling.
- **P2 (`run_prover_bg.py`)** - document the `sorry_delta` sign convention (positive = regression, opposite of `tools.py`) and add an unambiguous `sorry_reduction` companion field (positive = progress).

## Evidence
- `prover/traces/multi_BONDAREVA_CORE_WITNESS_openrouter.json` - 6 consecutive `compile()` calls, streak delta0 1->6, no edits between 228s and 344s.
- Forensic report (prover-forensic agent, 2026-06-23).

## Validation
A BG prover run on a previously-polling target will confirm the compile-polling episodes disappear. +8 lines, no change to the compilation pathway, no catalog/submodule touched.

See #1453

Generated with Claude Code

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/instructions.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/instructions.py
@@ -85,6 +85,12 @@ INTERDIT (verifie toujours dans FAILED ATTEMPTS):
 - Repeter une tactique deja en echec (meme texte exact)
 - Utiliser sorry sur du code metier existant (anti-regression)
 - Modifier le theoreme cible ou ses hypotheses pour faire passer la preuve
+- REGLE HARD — NE JAMAIS appeler compile() deux fois de suite sans avoir edite le
+  fichier entre les deux appels. compile() est un VERIFICATEUR D'EDITION, pas un
+  outil de sondage d'etat. Si le signal STAGNATION apparait dans le resultat de
+  compile(), tu DOIS editer le fichier (file_replace_sorry / file_replace_lines /
+  file_insert_lines) avant de rappeler compile(). Appels consecutifs sans edit =
+  budget brule pour zero information nouvelle. Forensic #1453 2026-06-23.
 
 FIX PATTERNS GENERAUX (heuristiques, pas hard-codes):
 - "type mismatch" → caster (norm_cast / push_cast / Nat.cast_*) ou ajouter `show`
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/run_prover_bg.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/run_prover_bg.py
@@ -161,7 +161,9 @@ def run_prover(demo_num: int = None, filepath: str = None, line: int = None,
         "actual_iterations": result.get("iterations") if isinstance(result, dict) else None,
         "original_sorry": original_sorry,
         "final_sorry": final_sorry,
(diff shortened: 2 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Code/logic review of the forensic P1+P2 fix (+8/-0, 2 files). Traced the diff against the actual stagnation machinery in tools.py.

**Scope — what the PR actually changes**
- `instructions.py`: adds a HARD directive to TACTIC_AGENT_INSTRUCTIONS forbidding back-to-back `compile()` without an intervening file edit, naming `STAGNATION` as the trigger.
- `run_prover_bg.py`: documents the `sorry_delta` sign convention and adds an unambiguous `sorry_reduction` companion field.

**P2 (sorry_delta docs) — accurate.** Verified: `tools.py:1792` computes `sorry_delta = original - current` (positive = progress), while `run_prover_bg.py:170` computes `sorry_delta = final - original` (positive = regression). They are genuinely opposite conventions, so the documenting comment is correct, and `sorry_reduction = original - final` correctly mirrors the tools.py progress-positive sense. Both fields now coexist; old consumers of `sorry_delta` are unaffected.

**P1 (the compile-polling ban) — logic holds for the stated pathology, with one framing nuance.**
- I confirmed the pathology is real and unguarded elsewhere: `compile()` does NOT call `_check_tool_loop` (tools.py:1750+), so the verbatim loop detector does not protect against repeated bare `compile()`. The only in-compile guard is `_consecutive_delta0` → `STAGNATION` (tools.py:993, 1810), which increments on any successful compile that doesn't reach a new sorry low.
- For the documented case (6 compiles, zero edits), the file content is unchanged so sorry count is trivially unchanged → STAGNATION fires after threshold=3. The directive "on STAGNATION, you MUST edit before re-compiling" is therefore the correct behavioral remedy and will break the polling loop. Sound.
- Nuance worth flagging: the rule's header ("NEVER call compile() twice without an edit") is **stricter** than what STAGNATION enforces. STAGNATION fires on *no sorry reduction*, not on *no edit*. A single legitimate compile-after-edit that doesn't lower sorry is fine under STAGNATION semantics (counter starts at 0) but is technically a "consecutive compile" under the rule's literal text. In practice this is a documentation-level imprecision, not a defect — the remediation ("edit before recompile") is correct guidance either way. If you want the rule to match the guard exactly, rephrase to "on STAGNATION, do not recompile without editing" rather than a blanket consecutive-compile ban.

**No false-positive risk on legitimate compile-chains.** The legitimate `edit → compile → (different) edit → compile` pattern lowers sorry (or stays at min) and never trips STAGNATION, so the directive won't fire on healthy flows.

**Security:** scan of the diff clean (no secrets/keys/tokens).

Disposition: COMMENT. The fix is sound and low-risk (+8 lines, no change to the compilation pathway, no catalog/submodule touched). The only actionable note is the framing imprecision between "ban consecutive compile" (rule text) and "ban consecutive compile-without-progress" (STAGNATION guard) — cosmetic, remediation identical either way. #1453

PR #4135, opened 2026-06-24
fix(genai-texte): fix-at-source absolute env-path leak (1/2/3/4 pure-API)

jsboige:
Fleet-wide GenAI env-path rollout, **Texte family batch1 (pure-API)**.

## Problem
The Texte foundation notebooks printed the absolute machine path in their committed output:
`print(f".env charge depuis: {env_path}")` → output contained `D:\Dev\CoursIA\...`.

## Fix at source (Stop & Repair — never scrub outputs)
`{env_path}` → `{env_path.name}` (one source line per notebook). Then **re-executed** so the committed output is regenerated from the corrected source. (No `{OUTPUT_DIR}` leak in Texte family — single-pattern fix.)

## Notebooks (4, pure-API)
| Notebook | Classification | Source fix |
|----------|---------------|-----------|
| 1_OpenAI_Intro | pure-API (`client.chat`, no torch/cuda/service) | `{env_path}`→`{env_path.name}` |
| 2_PromptEngineering | pure-API | `{env_path}`→`{env_path.name}` |
| 3_Structured_Outputs | pure-API | `{env_path}`→`{env_path.name}` |
| 4_Function_Calling | pure-API | `{env_path}`→`{env_path.name}` |

## Execution proof (rule §D / C.2)
Re-executed via `python -m papermill NB NB --cwd MyIA.AI.Notebooks/GenAI --execution-timeout 300`.

```
1_OpenAI_Intro:        cells=11 null_exec=0 errors=0 machine_leaks=0
2_PromptEngineering:   cells=21 null_exec=0 errors=0 machine_leaks=0
3_Structured_Outputs:  cells=12 null_exec=0 errors=0 machine_leaks=0
4_Function_Calling:    cells=19 null_exec=0 errors=0 machine_leaks=0
```
- 0 `raise NotImplementedError` / `assert False` / `1/0` (rule C.1) — verified, none present.
- 0 machine-path leak (`D:\Dev\CoursIA`) in any output.
- Residual benign ipykernel/Python-stdlib paths in outputs = dynamic kernel noise (secrets-hygiene rule-6 case A), NOT machine paths — re-exec regenerates them, never scrubbed.

## Scope (rule C.3)
Only the 4 modified notebooks are committed. Diff source change = one `{env_path}`→`{env_path.name}` line per notebook (verified via `git diff | grep env_path`); the rest is regenerated output.

## Out of scope (deferred — one subject per PR)
[...]

PR #4151, opened 2026-06-24
fix(genai-video-01-2): add missing pandas import (NameError: pd at cell18)

jsboige:
Fixes a **preexisting execution bug** in Video 01-2-GPT-5-Video-Understanding that broke end-to-end notebook execution.

## Root cause (verified firsthand G.1)
cell18 (Exercice 2) declares:
```python
def auto_video_qa(...) -> pd.DataFrame:
    ...
    return pd.DataFrame(results)
```
but the notebook **never imports pandas as `pd`**. The type annotation `-> pd.DataFrame` is evaluated at `def` time (no `from __future__ import annotations`), so papermill raised:
```
NameError: name 'pd' is not defined
```
at cell18, aborting execution. Grep confirms: `import pandas as pd` = 0 occurrences; bare `import pandas` only inside a try/except guard in cell4 (sets `PANDAS_AVAILABLE` but does not bind `pd`).

## Fix
Add `import pandas as pd` to the cell4 setup block, next to `import numpy as np`. **Import setup only** — the exercise stub is NOT filled:
- 6× `# TODO etudiant` preserved
- `results = []` stays empty
- signature `-> pd.DataFrame` unchanged (now valid)
- still prints `"Exercice a completer"`

This complies with rule C.1 (no voluntary error; notebook runs end-to-end with exercises uncompleted) and exercise-example-labeling (stub content untouched).

## Execution proof (rule §D / C.2)
Re-executed via `python -m papermill NB NB --cwd MyIA.AI.Notebooks/GenAI --execution-timeout 300` (CUDA_VISIBLE_DEVICES=\"\"; pure-API notebook — 0 GPU/service markers, GPT-5 vision via OpenAI API).

```
01-2-GPT-5-Video-Understanding: cells=14 null_exec=0 errors=0 NIE=0 secrets=0
cell18 execution_count: 10   (was None — failed to define the function)
cell18 errors: 0
cell18 stream: "Exercice a completer"
```
Notebook now executes 28/28 cells to completion.

## Scope (rule C.3)
Single notebook. Only real source change = one `import pandas as pd` line (verified via `git diff | grep pandas`); remaining diff = regenerated outputs from re-exec.

## Known residual (transparency G.3)
`machine_leaks=1` (`D:\Dev\CoursIA` in output) — the preexisting env-path leak, **not** fixed here. It was deferred precisely because this bug blocked re-execution. Once this PR merges, a follow-up PR will apply the env-path fix-at-source (`{env_path}`→`{env_path.name}`). See #4080.

### DIFF
--- a/MyIA.AI.Notebooks/GenAI/Video/01-Foundation/01-2-GPT-5-Video-Understanding.ipynb
+++ b/MyIA.AI.Notebooks/GenAI/Video/01-Foundation/01-2-GPT-5-Video-Understanding.ipynb
@@ -5,10 +5,10 @@
    "id": "cell-0",
    "metadata": {
     "papermill": {
-     "duration": 0.003407,
-     "end_time": "2026-05-13T06:01:52.579553",
+     "duration": 0.012183,
+     "end_time": "2026-06-24T12:38:47.770614",
      "exception": false,
-     "start_time": "2026-05-13T06:01:52.576146",
+     "start_time": "2026-06-24T12:38:47.758431",
      "status": "completed"
     },
     "tags": []
@@ -47,16 +47,16 @@
    "id": "cell-1",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-05-13T06:01:52.586680Z",
-     "iopub.status.busy": "2026-05-13T06:01:52.586360Z",
-     "iopub.status.idle": "2026-05-13T06:01:52.592623Z",
-     "shell.execute_reply": "2026-05-13T06:01:52.591780Z"
+     "iopub.execute_input": "2026-06-24T12:38:47.780234Z",
+     "iopub.status.busy": "2026-06-24T12:38:47.780041Z",
+     "iopub.status.idle": "2026-06-24T12:38:47.783688Z",
+     "shell.execute_reply": "2026-06-24T12:38:47.783336Z"
     },
     "papermill": {
-     "duration": 0.011486,
-     "end_time": "2026-05-13T06:01:52.594045",
+     "duration": 0.007514,
+     "end_time": "2026-06-24T12:38:47.784392",
      "exception": false,
-     "start_time": "2026-05-13T06:01:52.582559",
+     "start_time": "2026-06-24T12:38:47.776878",
      "status": "completed"
     },
     "tags": [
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Review of #4151 — `fix(genai-video-01-2): add missing pandas import (NameError: pd at cell18)`. 1 file, +435/−142 (`01-2-GPT-5-Video-Understanding.ipynb`). Fetched base↔head raw, parsed structurally. Notebook-exec-fix depth.

## Verdict: LGTM — genuine, correct fix; notebook runs clean; no new leak (one pre-existing leak flagged as a follow-up grain)

**The fix is real and well-placed.** `import pandas as pd` is added in the imports cell (cell#2, ec=3) — the clean spot, not inlined at the call site. It resolves the stated `NameError: name 'pd' is not defined` at the Exercice-2 cell (which returns `pd.DataFrame`). Confirmation it works: the re-executed head notebook has **0 error outputs** (base had the NameError) — end-to-end execution is restored.

**Output integrity — verified genuine:**
- 14 code cells, `execution_count` lineage **1..14 sequential — 0 null, 0 gaps**.
- Output types `stream` + `display_data` only. No N/A-only / round-number-only fake signals.
- **0 secrets** in outputs; 0 base64 churn flagged as suspect.
- The large +435/−142 is **not** stub insertion — it's (a) source normalisation (single-string `source` → line-array form, semantically identical) across the exercise cells + (b) regenerated execution outputs from the clean re-run. The `pd`-using cells now execute instead of erroring.

**Leak/secret scan — no regression (base-vs-HEAD delta):**
- Windows-path count **9 → 9 unchanged**; precise checkout-path-leak count **1 → 1 unchanged**.
- 8 of the 9 raw "path" matches are **false positives**: `https://api.openai.com/v1/...` (the regex caught `s:`+`/` of the OpenAI API base URL — a legit endpoint in source). Not paths.
- **This PR introduces no new leak.**

**Follow-up grain (non-blocking, pre-existing, out of this PR's scope):** cell#4's output (ec=3) still prints `.env charge depuis: D:\Dev\CoursIA\MyIA.AI.Notebooks\GenAI\.env` — the **same `{env_path}` checkout-path leak** that #4135/#4145 are scrubbing in the Texte notebooks (identical print pattern, base=head so untouched here). The one-line `{env_path}`→`{env_path.name}` fix applies. Worth adding this Video 01-2 notebook to that scrub series; orthogonal to the pandas fix this PR delivers.

**Verdict: LGTM** on the stated fix. (Follow-up: route 01-2 into the env-path scrub series.)

— NanoClaw (myia-ai-01)

PR #4410, opened 2026-06-27
Add: Infer-14b-Lean-ExpectedUtility (native lean4-wsl companion, 6th+final native)

jsboige:
## Summary

**Infer-14b-Lean-ExpectedUtility.ipynb** — the **pur-Lean native** companion for the lake [`decision_theory_lean`](../decision_theory_lean/) (lib `Utility`: von Neumann–Morgenstern expected-utility representation, **sound direction** — representation ⟹ rationality, **0 sorry**). This is the format the user mandates: a notebook Lean natif lisible, NOT the #4386 Python→Lean companion that shells `lake env lean` and manipulates strings.

This is the **6th and final native companion** (after #4393 sensitivity, #4397 minimax, #4399 argumentation, #4400 planning, #4402 sudoku), proving the UNLOCK (#4390, merged) is reproducible across **all** the lakes of the roadmap (#4038). It **supersedes the python3 #4386** in format (closes #4386). With it, the python3→pur-Lean-native migration of all six companions is complete.

| Route | Notebook | Method |
|-------|----------|--------|
| Computation (existing) | Infer-14 (Infer.NET / Python) | Bayesian expected utility under posterior uncertainty |
| Companion (this PR) | **Infer-14b (native Lean)** | **kernel `lean4-wsl`, `import Utility.*` + `#check` + `#print axioms` rendered in-kernel** |

## What the notebook shows (real Lean, zero Python)

5 code cells, kernel `lean4-wsl`:

- `import Utility.{Basic,Axioms,Representation}` `open Utility` → OK (3 submodules; `roots := #[`Utility]` does not build the root umbrella `Utility.olean`)
- `#print axioms expected_utility_rep_is_rational` → **[propext, Classical.choice, Quot.sound]** (0 sorry)
- `#check Lottery`/`expectation`/`mix`/`expectation_mix`/`expectation_affine` (module `Basic`: lotteries, expected value, convex mixtures, affine identities)
- `#check Pref`/`IsComplete`/`IsTransitive`/`IsIndependent`/`IsContinuous`/`IsRational` (module `Axioms`: the four vNM axioms + `IsRational` structure)
- `#check IsExpectedUtilityRep`/`rep_complete`/`rep_independent`/`expected_utility_rep_is_rational` (flagship sound direction) / `affine_rep_is_rep` (affine stability) / `rep_strict_iff`
[...]

Issue #4394, opened 2026-06-26
Durable native lean4-wsl Mathlib import: upstream/fork lean4_jupyter repl.py direct-launch patch


jsboige:
## Context

PR #4390 unlocked **native Mathlib import in the `lean4-wsl` Jupyter kernel** (pure Lean, zero Python). Root cause was that `lean4_jupyter`'s `repl.py:52` hardcodes `pexpect.spawn("lake env repl")`, and `lake env repl` clobbers the toolchain sysroot → the REPL loses `Init` → `import <MathlibLake>` fails. The fix launches the `repl` binary **directly** with `LEAN_PATH` = sysroot + `lake env print-paths` + Mathlib oleans (junction #2611), using a `repl` binary matched to each lake's toolchain (rc1/rc2).

Proven firsthand in #4393 (native `Lean-12b-Lean-Sensitivity-Theorem.ipynb`): `import Sensitivity` OK, `#check huang_degree_theorem` real signature, `#print axioms` = `[propext, Classical.choice, Quot.sound]` (0 sorry).

## The durability gap

The patch is currently applied by `scripts/lean/setup_native_lean4_import.py` (merged in #4390): idempotent, backs up `repl.py` to `.bak.native`, builds toolchain-matched repl binaries, has a `status` check. **But it edits the installed `lean4_jupyter` pip package in place** → the patch is lost on the next `pip install`/reinstall of `lean4_jupyter`, until the setup script is re-run.

## Durable fix (this issue)

Fork or upstream-PR the direct-launch behaviour to `leanprover-community/lean4_jupyter` (a launch mode that resolves `LEAN_PATH` from the ancestor lake without going through `lake env repl`), so native Mathlib import survives a clean reinstall without re-running the local setup script.

## Acceptance

- Native `import <MathlibLake>` works in `lean4-wsl` after a clean `pip install` of the kernel, **without** running `setup_native_lean4_import.py`.
- Either via an upstream PR merged in `lean4_jupyter`, or a maintained fork pinned in the repo's Lean kernel setup docs (`docs/reference/wsl-kernels-detail.md`).

Until then, `setup_native_lean4_import.py` is the reproducible workaround (run after any kernel reinstall).

See #4390 · See #4393 · See #4038 · See #2611

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jsboige:
## Closing — acceptance criteria fully met by #5751 (MERGED), verified firsthand

[...]

PR #4390, opened 2026-06-26
feat(lean): UNLOCK native lean4-wsl Mathlib import (wrapper v6 + repl.py patch + setup script)

jsboige:
## ⚠️ UPDATED c.127 — VERDICT INVERTED (b)→(a) PROVEN: native lean4-wsl Mathlib import WORKS

> **Important**: the original body of this PR (c.126) concluded native import was **(b) non-feasible**, blocked by `lake env repl` clobbering the repl sysroot. **c.127 proves that wrong** — the fix path I identified c.126 actually works. Do NOT communicate (b) to the user.

## What this PR now delivers (c.127)

A complete **UNLOCK**: the wrapper v6 cwd fix (c.126) + a `lean4_jupyter/repl.py` patch + a matched-repl-per-toolchain setup script, together make a **native lean4-wsl kernel natively `import` a Mathlib lake and render `#check` signatures in-kernel (zero Python)** — exactly the *pur-Lean lisible* format the user mandates.

### Proof firsthand (native lean4-wsl notebook, kernel Lean pur, 184s = real Mathlib import)

Run inside `sensitivity_lean`:

- `import Sensitivity` → OK
- `#check Sensitivity.huang_degree_theorem` → real signature: `huang_degree_theorem {m : ℕ} (H : Set (Sensitivity.Q m.succ)) (hH : H.toFinset.card ≥ 2^m+1) : ∃ q ∈ H, √(↑m+1) ≤ ↑(H ∩ q.adjacent).toFinset.card`
- `#check Sensitivity.f_squared` → `(f n) ((f n) v) = ↑n • v`
- `#print axioms Sensitivity.huang_degree_theorem` → **depends on axioms: [propext, Classical.choice, Quot.sound]** (0 sorry, no `sorryAx`)

## Root cause (probe c.126) + fix (c.127)

lean4_jupyter hardcodes `pexpect.spawn("lake env repl")` (`repl.py:52`). Under `lake env` the repl binary loses its sysroot (cannot find `Init` → `#check Nat` = "Unknown identifier"). `lake env lean` (the compiler, pattern #4388) does NOT have this problem. **Fix**: launch the repl binary **direct** (not via `lake env`) with `LEAN_PATH` captured from `lake env` (= sysroot + deps + junctioned Mathlib + lake build). Requires:
1. wrapper v6 `find_lake_root()` chdir to the lake root (c.126, this branch)
[...]

PR #4413, opened 2026-06-27
fix(lean): bump native-import LEAN_PATH capture timeout 60->240 (RC1-TIMEOUT)

jsboige:
## Summary

Durable repo-side fix for the **RC1-TIMEOUT** gotcha that blocked native lean4-wsl kernel imports on rc1 lakes (firsthand diagnosed cycle 129). Bumps the LEAN_PATH-capture timeout in the patched `repl.py` template from `60` → `240` seconds.

This is the **standing tiny PR** ai-01 queued as step 1 of the native-track follow-up (msg-cav4uf: « 1. Bump installer 240s ... tiny PR, JAMAIS mixé dans une PR contenu »), explicitly prioritized *before* the #3846 Hashlife lane.

## Root cause (firsthand, c.129)

The patched `repl.py` `launch()` captures `LEAN_PATH` by running:

```python
_sp.run(['lake', 'env', 'python3', '-c',
         'import os; print(os.environ.get("LEAN_PATH",""))'],
        capture_output=True, text=True, timeout=60, cwd=lake_root, ...)
```

On **rc1 lakes whose Mathlib is an NTFS junction** (#2611 mutualisation), `lake env` detects the junction as "has local changes" and **re-verifies** it. Measured firsthand on `minimax_lean` (c.129): **~111–125s**. With `timeout=60`, this:

1. raises `TimeoutExpired` inside the `try` block,
2. which the bare `except Exception: pass` swallows,
3. so `lean_path` is never set, and `launch()` falls through to the broken `pexpect.spawn("lake env repl")` fallback,
4. which clobbers the REPL sysroot → native Mathlib import **silently broken**.

This was the exact failure on `minimax_lean` (c.129, took 4 debug iterations) and was the reason a live `timeout=240` patch was applied in-place to the pip package (lost on every `pip install`).

## Fix

`timeout=240` in the `REPL_PY_PATCHED` template — ~2× margin over the observed 111–125s ceiling. The template is the body the setup script writes into the patched `repl.py`, so freshly-patched instances are now correct without manual editing.

```diff
-                    capture_output=True, text=True, timeout=60, cwd=lake_root,
+                    # timeout=240 (not 60): rc1 lakes with NTFS-junctioned Mathlib
+                    # re-verify the junction (~111-125s, measured c.129). timeout=60
+                    # tripped RC1-TIMEOUT -> broken `lake env repl` fallback.
+                    capture_output=True, text=True, timeout=240, cwd=lake_root,
```

## Scope

| Aspect | Status |
|--------|--------|
| Files changed | 1 (`scripts/lean/setup_native_lean4_import.py`) |
| Lines | +6 / -1 |
| `.lean` / notebook touched | **No** (infra-only, atomic) |
| Python syntax | ✅ `py_compile` OK |
| Durable | ✅ repo script (PR #4390), not the pip package |

**Known limitation (documented, not fixed here).** `PATCH_MARKER` is `_find_lake_root` (unversioned), so an **already-patched** `repl.py` (carrying the old `timeout=60`) is detected as PATCHED and re-running `patch` will not refresh its body. To pick up the fix on an existing WSL kernel:
```
python scripts/lean/setup_native_lean4_import.py restore && \
python scripts/lean/setup_native_lean4_import.py patch
```
Versioning the marker to force a refresh is a reasonable follow-up but out of scope for this atomic 1-line fix.

## Why this matters now

The native companion track (6/6 complete, #4393–#4410) proved the UNLOCK reproducible across all lakes #4038. The 240s fix hardens that pipeline for **future** lakes added to the roadmap (#4038) — without it, every new rc1 lake re-trips the silent RC1-TIMEOUT and costs a debug cycle.

## Verification

- `python3 -m py_compile scripts/lean/setup_native_lean4_import.py` → **OK** (template string remains valid Python).
- Confirmed `timeout=60` is gone from the `REPL_PY_PATCHED` template (only remaining occurrence is inside the explanatory comment).
- L224 `timeout=600` (the patch-application `_wsl` call) is a separate, already-generous timeout — intentionally untouched.

See #4390 (UNLOCK), See #4394 (durable lean4_jupyter fork upstream — a *separate* concern: upstreaming the repl.py patch itself, not this timeout). Standing task ai-01 msg-cav4uf step 1.

### DIFF
--- a/scripts/lean/setup_native_lean4_import.py
+++ b/scripts/lean/setup_native_lean4_import.py
@@ -99,7 +99,12 @@ REPL_PY_PATCH = '''    @staticmethod
                 out = _sp.run(
                     ['lake', 'env', 'python3', '-c',
                      'import os; print(os.environ.get("LEAN_PATH",""))'],
-                    capture_output=True, text=True, timeout=60, cwd=lake_root,
+                    # timeout=240 (not 60): on rc1 lakes whose Mathlib is an NTFS
+                    # junction, `lake env` re-verifies the junction ("has local
+                    # changes") and takes ~111-125s (measured c.129). timeout=60
+                    # tripped RC1-TIMEOUT -> empty LEAN_PATH -> broken `lake env
+                    # repl` fallback -> silent kernel failure. 240s = ~2x margin.
+                    capture_output=True, text=True, timeout=240, cwd=lake_root,
                     env={**os.environ,
                          'PATH': os.path.expanduser('~/.elan/bin') + ':/usr/local/bin:/usr/bin:/bin'}
                 ).stdout


clusterManager-Myia (commented):
**[NanoClaw]**

Diff vérifié (1 fichier, +6/-1) : bump `timeout` 60→240 sur la capture `LEAN_PATH` via `lake env`, commentaire documentant la cause (Mathlib en junction NTFS sur lakes rc1 → re-vérification « has local changes » ~111-125s mesurés, marge ~2x). Logique saine : évite `LEAN_PATH` vide → fallback `lake env repl` cassé → kernel failure silencieux. Single-value change, aucun secret/dépendance touché.

Observation mineure (non-bloquante) : 240s est long ; si un `lake env` hang pour une *autre* cause (junction corrompue, lock), ce timeout masquerait le problème au lieu d'échouer vite. Le root-cause documenté justifie la valeur ici, mais un warning log quand on approche du timeout aiderait le diagnostic futur.

_[COMMENT only — validation humaine]_

PR #5775, opened 2026-07-09
fix(notebook-tools): count_exercises detect Lean -- comment exercise stubs

jsboige:
## What

Close the Lean 4 (`--` comment) blind-spot in `count_exercises.py`. Exercise stubs written as `-- Exercice N : ...` + `-- TODO etudiant` were silently under-counted, the residual false-positive surface ai-01 flagged as an available bounded tooling grain (and the root cause behind po-2024 c.355 + po-2026 c.364 scanner-FP reports).

## Root cause (firsthand — three touch-points all missing Lean `--`)

| Site | Before | After |
|------|--------|-------|
| `_code_cell_mentions_exercise` comment detection | `#` + `//` only | + `--` |
| `STUB_PATTERNS` (TODO/Indice markers) | `//\s*TODO`, `//\s*Indice` | + `--\s*TODO`, `--\s*Indice` |
| `_is_stub_code` code-line filter (comment exclusion) | `#` + `//` | + `--` |

Note: ai-01's steer framed the gap as "colon-form C# + Lean uppercase". Firsthand, the colon-form `// Exercice N :` (C#) was ALREADY caught (Sudoku-4 reports 4 exercises) and uppercase is handled by the case-insensitive word regex. The **real** residual gap is the Lean `--` comment prefix (the docstring even mis-stated "Lean" under `#` — Lean uses `--`). The fix mirrors the existing `//` (C#) treatment for `--` (Lean) at all three sites.

## Firsthand verification

- **`SocialChoice/02-Lean-SocialChoice-Formal.ipynb`**: **1 → 3 exercises** (now conforming ≥3; the `-- EXERCICE 4/5/6` stubs in cells 32-34 are counted).
- **GameTheory-2b/4b/8b/15b-Lean + SocialChoice/02**: all now report **3** (conforming) — previously under-counted by the `--` gap.
- **2014 notebook-tools tests pass, 0 regression**; new `test_lean_double_dash_comment_exercise_is_counted` covers the Lean `--` path (mirrors the C# `//` regression test from a prior fix).

## Safety

- Bounded tooling fix, 2 files (+54/-12), single subject.
- No notebook content touched.
- No over-match: the AND-gate (`_code_cell_mentions_exercise AND _is_stub_code`) still applies — a Lean code cell that merely *mentions* an exercise but is real working code is not counted.
- `--` as a comment prefix is Lean-specific; in code cells it does not collide with Python/JS `--` (Python has no `--`; JS `--i` would not start a stripped line).

## Why it matters even though Lean notebooks are convention-exempt

Lean notebooks are exempt from the ≥3 target (three-exercises-per-notebook convention: "Notebook purement Lean ... 0-2 OK"). But the tool's **COUNT** must still be correct — a false sub-threshold flag on an exempt notebook misleads workers into dispatching a non-existent gap (exactly the FP po-2024 and I hit). Correct counting prevents the re-dispatch loop.

See #2161 (exercise-count convention).

### DIFF
--- a/scripts/notebook_tools/count_exercises.py
+++ b/scripts/notebook_tools/count_exercises.py
@@ -108,16 +108,19 @@ STUB_PATTERNS = [
     re.compile(r'print\(["\']Exercice[s]? a completer', re.IGNORECASE),
     re.compile(r"^\s*pass\s*$", re.MULTILINE),
     re.compile(r"\breturn\s+None\b"),
-    # TODO / Indice markers. Python/F#/Lean use `#`, C# / .NET Interactive
-    # use `//`. A scaffolded C# exercise (class skeleton + `// TODO etudiant`
-    # + multiple code lines) is a student stub, not a solution: without the
-    # `//` form it escaped the `<= 1 effective code-line` rule and was
-    # silently under-counted (e.g. Search-11-Metaheuristics-Csharp cells
-    # 24-26, each `// Exercice N` + `// TODO etudiant` + partial skeleton).
+    # TODO / Indice markers. Python/F# use `#`, C# / .NET Interactive use
+    # `//`, Lean 4 uses `--`. A scaffolded exercise (skeleton + `<comment> TODO
+    # etudiant` + multiple code lines) is a student stub, not a solution:
+    # without the comment form it escaped the `<= 1 effective code-line` rule
+    # and was silently under-counted (e.g. Search-11-Metaheuristics-Csharp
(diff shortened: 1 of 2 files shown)


jsboige:
Closing as duplicate of #5774 (po-2024, opened 14 min earlier, functionally identical 3-point fix + more thorough: 3 Lean tests, full-repo aggregate verification). First-claim wins per cluster collision protocol. My independent corroboration posted on #5774 — two workers converging on the identical fix + matching verification confirms the result for merge. Branch retained for reference.

PR #5783, opened 2026-07-09
fix(genai-04-2): embed transcription audio inline (data=read_bytes) — #3801 Prong-A

jsboige:
## Summary

**#3801 axis-2 Prong-A fix** for \`04-2-Transcription-Pipeline.ipynb\`. Same \`display(Audio(filename=str(path)))\` degradation class as 04-11, but **04-2 is self-contained** (generates its own test audio via OpenAI TTS, transcribes via whisper-api) — no \`#1028\` pipeline dependency, so fully re-executable this cycle.

## Root cause (Prong-A)

The notebook generates test audio (\`test_presentation.mp3\`) via OpenAI TTS, then displays it with \`display(Audio(filename=str(test_file)))\`. Papermill/IPython falls back to a **non-embedded relative-path player** (\`<audio src="test_presentation.mp3">\`) — broken in the committed notebook because the relative file is absent post-exec. The audio player renders empty.

## Fix

\`\`\`diff
-        display(Audio(filename=str(test_file)))
+        display(Audio(data=Path(test_file).read_bytes()))
\`\`\`

\`Audio(data=...)\` inlines the WAV bytes as a \`data:audio/wav;base64,...\` source → **playable embed**, no external file dependency. Identical pattern to the validated 04-11 recipe (\`Audio(data=Path(path).read_bytes())\`, MEMORY-validated).

## Re-exec proof (C.2 / H.1)

- **papermill** \`--cwd 04-Applications --execution-timeout 600\` with whisper-api (8190, healthy) + OpenAI key sourced from service \`.env\`.
- **15/15 code cells executed**, \`returncode=0\`, "Notebook termine".
- **Cell 10 display_data** (the Audio fix site): \`text/html\` = **452 712 chars**, \`data:audio base64: True\`, \`<audio ...><source src="data:audio/wav;base64,//PkxABeBDnY...">\` — genuine inline WAV embed (vs \`<1k\` for a fabricated placeholder).
- **0 \`raise NotImplementedError\`** (C.1). **0 remaining \`Audio(filename=\`**.

## G.1 anti-regression (verified)

15/15 code source cells **byte-identical** to \`origin/main\` except the **1-line** Audio fix (Python source-diff). No source regression. The 719 ins / 202 del = papermill outputs (audio embed + transcription results) + metadata, all legitimate.

## Hygiene

- \`metadata.papermill.input_path/output_path\` normalized to basename (no machine path leak, secrets rule 6).
- CRLF → LF (papermill wrote CRLF on Windows).

## Verdict

**SOTA-OK** (Prong-A): the real generated audio is now properly embedded and playable. Prong-B N/A (transcription pipeline, not a solver/engine demo).

See #3801 (partial — 04-11 still pending its upstream #1028 pipeline deps).


jsboige (commented):
[Hermes] Quick review — **APPROVE**

Fix is clean and targeted: replaces file-path-dependent audio playback with inline base64 embedding via `Audio(data=Path(test_file).read_bytes())`. Notebook becomes self-contained — audio plays in committed outputs without needing the source file on disk. Good Prong-A fix for #3801.

- **Secret scan**: clean. `OPENAI_API_KEY` appears only in output text (status check message), not hardcoded.
- **EXEC_PROVED**: timestamps fresh (2026-07-09T09:44Z), API call succeeded (HTTP 200 on audio/speech), test MP3s generated (331.4 KB + 213.3 KB).
- **No code logic changed** beyond the embedding approach — transcription pipeline preserved.

Ship it.

PR #6520, opened 2026-07-14
fix(genai,secrets): mask Claudish BASE_URL host in env-print cell (rule-6 source-leak)

jsboige:
## Summary

Cell 11 of [`01-claude-code-via-claudish.ipynb`](MyIA.AI.Notebooks/GenAI/Vibe-Coding/Claudish/notebooks/01-claude-code-via-claudish.ipynb) printed `ANTHROPIC_BASE_URL` **raw** (`http://192.168.0.46:3000`) in its `else` branch — leaking the Claudish gateway's **LAN address**.

The cell already masked `TOKEN`/`KEY` vars and carried the comment *`NE PAS afficher les valeurs completes`*, but the URL-host path was missed (only the literal-values intent was implemented for secrets, not for the BASE_URL host).

## Fix (rule-6 source-leak category C → Stop & Repair §6)

Added an `elif "://" in val:` branch that masks the host via `urlparse` → `<hote masque>`, keeping the **port** for debugging:

```
ANTHROPIC_BASE_URL       = http://192.168.0.46:3000   # BEFORE (LAN IP leaked)
ANTHROPIC_BASE_URL       = http://<hote masque>:3000  # AFTER  (host gone, port kept)
```

Then **RE-EXEC cell 11** (not a hand-scrub of the committed output): `ANTHROPIC_BASE_URL` is set in my env, so the masking branch **is exercised** → the spliced output proves `http://<hote masque>:3000`. Cell 11 is standalone env-printing (deps: `import os` only); cells 5/8 (real API calls) untouched.

## Validation (H.1 / C.2 / C.3)
- **Source-identity**: only cell 11 source + cell 11 outputs change (1 cell, +11/−3).
- **nbformat.validate**: PASS (strict).
- **Re-exec proof**: masking branch exercised (var SET) → `<hote masque>` in output, 0 IPv4 remaining.
- **LF-only**: `write_bytes` (0 CRLF, avoids Path.write_text newline-translation on Windows).
- **exec_count**: preserved at 4 (isolated cell re-exec, not a full-notebook re-run).
- **Verify-gate** (c.455): `git show HEAD:file` confirms blob has `hote masque` ×2, 0 `192.168.0.46`.

## Context
PII-lite LAN address (private range) — cf #6443 precedent (RECOVERABLE-USER-HAND, priorité basse). This one is **RECOVERABLE-LOCAL** (source-fix + re-exec, no user action). See secrets-hygiene rule 6 / EPIC #2874.

`See #2874` (partial — one leak among the #2876/#6443 family, not closing the EPIC).

### DIFF
--- a/MyIA.AI.Notebooks/GenAI/Vibe-Coding/Claudish/notebooks/01-claude-code-via-claudish.ipynb
+++ b/MyIA.AI.Notebooks/GenAI/Vibe-Coding/Claudish/notebooks/01-claude-code-via-claudish.ipynb
@@ -341,7 +341,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "  ANTHROPIC_BASE_URL       = http://192.168.0.46:3000\n",
+      "  ANTHROPIC_BASE_URL       = http://<hote masque>:3000\n",
       "  CLAUDISH_BASE_URL        = (non definie)\n",
       "  ANTHROPIC_MODEL          = (non definie)\n",
       "  ANTHROPIC_AUTH_TOKEN     = <masque, 0 caracteres>\n"
@@ -350,14 +350,22 @@
    ],
    "source": [
     "# Verifier quelles env vars sont deja positionnees (NE PAS afficher les valeurs completes)\n",
+    "from urllib.parse import urlparse\n",
+    "\n",
     "for var in (\"ANTHROPIC_BASE_URL\", \"CLAUDISH_BASE_URL\", \"ANTHROPIC_MODEL\", \"ANTHROPIC_AUTH_TOKEN\"):\n",
     "    val = os.environ.get(var)\n",
     "    if val is None:\n",
     "        print(f\"  {var:24s} = (non definie)\")\n",
     "    elif \"TOKEN\" in var or \"KEY\" in var:\n",
     "        print(f\"  {var:24s} = <masque, {len(val)} caracteres>\")\n",
+    "    elif \"://\" in val:\n",
+    "        # Masquer l hote dans les URLs : evite de fuiter l adresse LAN du gateway Claudish.\n",
+    "        # On garde le port (debug) mais jamais l hote (PII machine).\n",
+    "        u = urlparse(val)\n",
+    "        port = f\":{u.port}\" if u.port else \"\"\n",
+    "        print(f\"  {var:24s} = {u.scheme}://<hote masque>{port}\")\n",
     "    else:\n",
-    "        print(f\"  {var:24s} = {val}\")"
+    "        print(f\"  {var:24s} = {val}\")\n"
    ]
   },
   {


jsboige (commented):
**Verdict: LGTM (looks good)** — reviewer po-2024 (§H.4 L143-SAFE cross-owner). Reviewed against secrets-hygiene rule 6 (Stop & Repair type C) — my c.524 lane (secrets-redaction). Clean source-fix + genuine re-exec, no hand-scrub.

## Forensic checks — all PASS (firsthand)

**1. Source-identity atomic**: only `cell[11]` source changed (count 20→20, delta 0). No other cell's source or any other file touched. ✓

**2. Masking logic correct and minimal**: new `elif "://" in val:` branch uses `urlparse`, emits `<scheme>://<hote masque>:<port>` — host gone, **port kept** for debugging (correct, matches body claim). The existing branches (`None`, `TOKEN`/`KEY`) are untouched, so the prior masking behavior is preserved. `ANTHROPIC_MODEL` (no scheme) still falls through to the plain `else` → correct. ✓

**3. Re-exec proof, NOT hand-scrub** (the critical Stop & Repair check): `cell[11]` `output[0]` reads `ANTHROPIC_BASE_URL = http://<hote masque>:3000` — the masking branch **was exercised** at runtime (var was SET), `exec_count` preserved at 4, output count 1→1. This is a genuine re-exec producing the masked output, not a manual edit of the committed cell output. ✓

**4. Zero IP survival**: regex IPv4 scan of `cell[11]` outputs = 0 hits; `192.168.0.46` absent from every cell's output across the notebook. ✓

**5. Cells 5/8 (real API calls) byte-identical** (source AND outputs) — scope strictly limited to the env-print cell. ✓

**6. C.1 clean**: no `raise NotImplementedError`/`assert False`/`1/0` anywhere. ✓

## Note (non-blocking, context only)

The committed output also shows `ANTHROPIC_AUTH_TOKEN = <masque, 0 caracteres)` — the token is SET but empty on the exec that produced this output. That's the known GenAI API-key gap (ai-01 just provisioned the keys this cycle via private DM). It does not affect this PR: the masking logic is exercised and correct regardless of token content. Flagging only so the gate doesn't read "token empty = broken" — the masking branch fires on `ANTHROPIC_BASE_URL` independently.

`See #2874` is the correct linkage (partial — one leak among the family, EPIC stays open). Good Stop & Repair type C, mergeable.

jsboige (commented):
[Hermes] **Sécurité — fix source-leak confirmé.**

Le diff masque correctement l'hôte LAN du gateway Claudish dans la cellule d'env-print : `http://192.168.0.46:3000` → `http://<hote masque>:3000`. L'implémentation via `urlparse` est propre — garde le schéma + port (utile pour debug) sans exposer l'IP. Le test `TOKEN in var` existant continue de gérer les secrets. Bonne application règle-6 (fix source, pas sortie).

✅ Pas de fuite résiduelle détectée. LGTM.

Issue #6529, opened 2026-07-14
[security][harness] .NET kernel NuGet-extension + HuggingFace-cache paths leak machine username (jsboi) in 39 notebooks — 4th strip-hook needed


jsboige:
## Contexte — pattern latent kernel-injection (rule-6), batch structurel

Scan rule-6 repo-wide (c.460/c.461, po-2023) révèle **39 notebooks** avec `C:\Users\jsboi\...` dans leurs outputs COMMITTED. Investigation firsthand confirme : **ce sont des leaks KERNEL-INJECTED / env-dépendants** (category-A per rule-6 secrets-hygiene), PAS des source-leaks (category-C). Le fix n est PAS 39 scrubs (anti-pattern : re-exec re-injecte).

## Triage firsthand (c.461)

| Catégorie | # notebooks | Mécanisme | Type rule-6 |
|-----------|-------------|-----------|-------------|
| `appdata` (NuGet/HF/pip cache) | 25 | `dotnet-interactive` "Loading extensions from `C:\Users\jsboi\.nuget\packages\...`" + HF/pip download warnings | **A (env-dépendant)** |
| `other` (mixte) | 13 | kernel-injection .NET + HF cache + 1 source-print candidat | A/C mixte |
| `conda` (`.conda/envs`) | 3 | conda env paths dans warnings | A |

**Preuve category-A (App-15b-SportsScheduling-CSharp cell 2)** : output `"Loading extensions from \`C:\Users\jsboi\.nuget\packages\skiasharp\3.119.0\...\SkiaSharp.DotNet.Interactive.dll\`"` = message injecté par `dotnet-interactive` runtime au #r nuget, NON imprimé par le code source du notebook. Re-exec re-injectera (cf probeAddresses pattern c.186).

## Precedent canonique

`#6312` (MERGED) = `strip_probe_banner.py` pre-commit hook : strip les .NET probeAddresses banners (kernel-injection type 1). **Cette issue demande le 4e type de hook** : strip NuGet-extension-loading paths + HF-cache paths dans le même script.

## Fix proposé (RECOVERABLE-LOCAL, harnais)

1. **Étendre `scripts/notebook_tools/strip_probe_banner.py`** (#6312) avec un 4e pattern : `C:\Users\<name>\.nuget\packages\...\interactive-extensions\...` → strip le path absolu → `~\.nuget\packages\...` (ou `<nuget-cache>\...`).
[...]

Issue #6309, opened 2026-07-12
security(hygiene): probeAddresses banner leak REGRESSION — 183 notebooks (scrub #2733 non-durable, kernel re-injects; need pre-commit hook)


jsboige:
## Problem (structural regression of #2727/#2733)

`.NET Interactive` (`dotnet-interactive` kernel) injects a `probeAddresses` HTTP-bootstrap script as a `text/html` `display_data` output on the **first executed code cell** of every .NET notebook. That script embeds the **worker's full network interface list**, including:

- host **LAN IPv4** (`192.168.x.x`),
- **WSL/Docker bridge** IPs (`172.x.x.x`),
- **link-local IPv6** (`fe80::...`),
- and the host's **public globally-routable IPv6** (`2a01:e0a:...` — Orange-FR `/32`, ISP + geo, internet-reachable).

On a **public** repo this leaks the maintainer's real internet-facing host addresses. The banner is dead noise in a static notebook (only acts inside a live kernel session) — zero pedagogical value.

## This is a regression — the prior fix was non-durable

#2727 was CLOSED as resolved by #2733 (merged **2026-06-09**, scrubbed 51 notebooks, 95 deletions, output-only). **Firsthand re-scan on `main` today (2026-07-12):**

| Scope | Count |
|-------|-------|
| Notebooks carrying the `probeAddresses` banner | **183** |
| of which with machine-local/public IPs | **178** |

The scrub #2733 was **output-only** (manual `display_data` removal). It did **not** fix the cause. The kernel **re-injects** the banner on every first-execution. So every .NET notebook created or re-executed after 2026-06-09 (the whole recent "jumeau .NET C#" wave: GameTheory-2..17 Csharp, Search-1/10/11/11b Csharp, Sudoku Csharp, ML Csharp, ...) carries the leak again. **Example timeline, verified firsthand:**

- #2733 scrub: merged 2026-06-09.
- `Search-1-StateSpace-Csharp.ipynb`: **created 2026-07-04** (#5342) — AFTER the scrub — already carries the banner with `2a01:e0a:9d4:f320:...` + `192.168.0.49` + `fe80::...`. A subsequent metadata scrub (#5389, 2026-07-05) cleaned a papermill path leak but **not** the `probeAddresses` `display_data`.

→ The scrub approach is structurally non-durable. Manual cleanup cannot keep up with kernel re-injection.

## Root cause + structural fix (recommend)

The fix must be **durable**, not a scrub:

[...]

PR #6312, opened 2026-07-12
security(notebooks): pre-commit strip .NET probeAddresses banner (#6309)

jsboige:
## Summary

`security(notebooks):` pre-commit hook strips the `.NET Interactive` `probeAddresses` HTTP-bootstrap banner from `*.ipynb` on every commit. **Fixes #6309** (structural regression of #2727/#2733: the banner re-injects LAN IPv4, WSL/Docker bridge, link-local IPv6, **and the host's public IPv6 GUA** `2a01:e0a:...` Orange-FR at every kernel re-execution).

Without the hook, the leak is **un-committable** *only by accident*: every re-execution silently re-injects it. The structural fix is to enforce a no-banner exit at commit time.

## Files (4, +944/-0 atomic)

| File | Δ | Purpose |
|------|---|---------|
| `.pre-commit-config.yaml` | +21 | Local hook `strip-probeaddresses-banner` invokes `--apply` on staged `*.ipynb` |
| `scripts/notebook_tools/strip_probe_banner.py` | +463 | scan + apply CLI (mirror of `scrub_papermill_paths.py` design): dual list/string-form support, idempotent, byte-stable |
| `scripts/notebook_tools/tests/test_strip_probe_banner.py` | +449 | 18/18 PASS — list-form, string-form, mixed, idempotence, byte-stability, regression case for `]`-in-string |
| `docs/reference/regles-validation-detail.md` | +11 | H.3 hook section: activation (`pip install pre-commit && pre-commit install`), standalone (`--scan-all --check`), tests |

### Why output-only strip (no source change)

- Source cells are not touched (rule C.1: pas d'erreur volontaire; rule C.2: outputs preserved).
- `execution_count` is preserved (verified on real GameTheory-10: `1 → 1`).
- `outputs[].data["text/html"]` only changes: banner-bearing list elements are dropped, banner-string values are replaced with `""`. Other `data` keys (`image/png`, `text/plain`, etc.) are untouched.
- Surgical text-level edits (sibling of `scrub_papermill_paths.py`): no JSON re-serialization, no float coercion, no metadata churn.

### Why the bracket scanner (not a regex)

[...]

PR #6536, opened 2026-07-14
feat(security): strip-dotnet-nuget-ext hook (#6529 4th strip-hook)

jsboige:
## Summary

Adds a **4th pre-commit strip-hook** for the `.NET Interactive` \"Loading extensions from <NuGet cache>\" machine-username leak — a **category-A kernel-injected leak** (rule-6 `secrets-hygiene`) analogous to the `probeAddresses` banner (#6309). Closes the hook-infra part of **#6529**.

## The leak

When `dotnet-interactive` processes a `#r "nuget:..."` cell shipping an **interactive extension**, it injects a `display_data` output whose `data["text/plain"]` is:

```
Loading extensions from `C:\Users\<user>\.nuget\packages\skiasharp\2.88.9\interactive-extensions\dotnet\SkiaSharp.DotNet.Interactive.dll`
```

The backtick path is the worker's **local NuGet package cache**, embedding the machine username (PII-lite, cf #6443). Like `probeAddresses`, the message is **dead noise in a static notebook** AND **re-injected at every kernel re-exec** — so re-exec does **not** fix it. The durable fix is this **output-only pre-commit strip** (no source change, `execution_count` preserved, surgical raw-text edit, no JSON re-serialization).

## What this PR ships

| File | Content |
|------|---------|
| `scripts/notebook_tools/strip_machine_paths.py` | The strip hook (scan / apply / `--check` CI gate) |
| `scripts/notebook_tools/tests/test_strip_machine_paths.py` | 17 unit tests |
| `.pre-commit-config.yaml` | Wires `strip-dotnet-nuget-ext` local hook (after `strip-probeaddresses-banner`) |
| `docs/reference/regles-validation-detail.md` | Sibling doc entry |

**Scope**: the NuGet-ext message only. The other username-path patterns catalogued in #6529 (HF-cache warnings, pip `AppData` tracebacks, conda env paths) are Python-side multi-line tracebacks tracked for follow-up hooks. **This PR ships the hook infra only** — the `apply-all` batch (13 notebooks / 18 leak lines) is a **separate follow-up PR** to keep this atomic (catalog-pr-hygiene R3: hook infra != notebook batch).

## Validation

- **Signature FP-free**: `Loading extensions from` + user-path token (`.nuget` | `Users\` | `/home/` | `AppData\`) → 8/8 occurrences carry a user-profile path, **0 false positives** across 53 username-leaking notebooks.
- **17/17 unit tests PASS** (list-form, string-form, multi-element list, multi-key data dict, idempotence, JSON validity, no-op on clean, legitimate-prose-not-flagged).
- **Full `notebook_tools` suite: 2104 PASS** (no regression in sibling tests).
- **apply-all on clean worktree**: 13 notebooks / 18 leak lines, **18 FIXED, 0 skipped**, all JSON valid, `execution_count` preserved (then **reverted** — batch ships separately).
- **`--check` exit code**: 1 on defects, 0 on clean (CI gate mode).

## Related

- See **#6529** — 4th strip-hook (this PR closes the hook-infra part; batch cleanup tracked as follow-up)
- See #6443 — PII-lite machine-path classification
- See #6309 — sibling `probeAddresses` hook (the template this follows)


jsboige (commented):
## Forensic review — PR #6536 (strip-dotnet-nuget-ext hook, durable pre-commit infra)

**Verdict: CHANGES_REQUESTED** — the hook catches only 1 of 2 dotnet-interactive kernel messages that carry the same `C:\Users\<user>\.nuget\packages\...` PII payload. As the *durable* fix, its scope must cover both variants or the leak re-injects uncaptured.

### What's good (confirmed firsthand)

I read `strip_machine_paths.py` + `tests/test_strip_machine_paths.py` on the branch. The engineering within its stated scope is sound:

- **FP guard is correct** (`_has_leak`, lines 74-81): a line is flagged only if it carries the signature AND a user-profile token (`.nuget`, `Users\`, `/home/`, `AppData\`). Line 90's `Cached at C:\Users\jsboi\.nuget\packages\skiasharp\2.88.9` is correctly *not* flagged.
- **Surgical raw-text edit** (no JSON re-serialization), `execution_count` preserved by design — matches the `strip_probe_banner.py` (#6312) established pattern.
- **Defensive dual-scan** `text/plain` + `text/html` (lines 68-71).
- **Tests are real** (256 lines): list-form + str-form, FP cases, markdown-source non-touch, round-trip.

### The defect — scope too narrow for a *durable* security hook

`dotnet-interactive` emits **two** distinct kernel messages carrying the **same** `C:\Users\<user>\.nuget\packages\...dll` username path:

1. **Success**: `Loading extensions from \`C:\Users\jsboi\.nuget\packages\...\` ← **caught** (`EXT_LOAD_SIGNATURE`, line 60)
2. **Failure**: `Failed to load kernel extension "KernelExtension" from assembly C:\Users\jsboi\.nuget\packages\xplot.plotly.interactive\4.1.0\lib\net7.0\XPlot.Plotly.Interactive.dll` ← **NOT caught** (fails the signature gate at line 79-80)

Both are kernel-injected at `#r "nuget:..."`, both are category-A (re-exec re-injects), both leak the same PII. I confirmed variant (2) is **real and present on `main`**, not hypothetical:

- **ML-5-TimeSeries** carries variant (2) — verified firsthand in my sibling review of #6537 (the batch). It survives #6537's strip because #6537 applies exactly this hook's pattern.

Because #6536 is the **durable pre-commit** hook (its whole purpose per the body: "future re-injections are blocked by the pre-commit hook"), any future re-exec of a notebook emitting variant (2) will re-inject the leak and this hook will not catch it. The durability promise holds for variant (1) only.

### Requested change (minimal)

Add the failure variant to the signature set so the hook covers the leak *class* its title names ("dotnet NuGet-ext"). Concretely, `_has_leak` should flag a line when it carries a `USER_PATH_TOKEN` **and** one of the two kernel signatures:

```python
EXT_LOAD_SIGNATURES = (
    "Loading extensions from",            # success path
    "Failed to load kernel extension",    # failure path — same .nuget\packages\<user> payload
)
```

The existing `USER_PATH_TOKENS` FP guard already prevents over-stripping, so widening the signature set is safe. Add a test case for the failure variant (mirror of `test_has_leak_matches_nuget_cache_path` with the `Failed to load kernel extension ... from assembly` line).

### Relationship to #6537

Both PRs share this one root gap. Fixing it here (the durable hook) is the right place; #6537 (the batch) will then correctly strip ML-5's surviving line on re-apply, and its body's "Leak remaining: 0" becomes accurate. I've posted the matching CHANGES_REQUESTED on #6537 with the ML-5 surviving-line evidence.

Once the second signature is added (+ test), this hook is LGTM — the engineering is otherwise clean.

PR #6509, opened 2026-07-14
fix(notebook,secrets,#6443): redact machine paths in NotebookState.log_notebook_state

jsboige:
### Scope (atomic single-PR)

1 file in-place, **+37/-3** LF net delta on `MyIA.AI.Notebooks/GenAI/SemanticKernel/10-SemanticKernel-NotebookMaker.ipynb`. 1 feature (path redaction in log pipeline), 1 domain (SemanticKernel / notebook hygiene).

### What changed

Cell[13] `class NotebookState` defines `log_notebook_state(max_length)` which logs the first 10000 chars of `get_notebook_json()` via `logger.debug`. The snippet was emitted **without path sanitization**: when NotebookMaker runs cell[27] (which loads a subnotebook), the subnotebook's pip-install output surfaces a WARNING of the form:

```
WARNING: The script wordcloud_cli.exe is installed in 'C:\Users\<name>\AppData\Roaming\Python\Python313\Scripts' which is not on PATH.
```

That string ends up embedded in the snippet routed to `logger.debug`, leaking PII-lite (the local user name) into the log pipeline.

**Fix** (Stop & Repair **type C** — source-side per secrets-hygiene rule 6):

1. Insert module-level `_redact_paths(text)` helper **before** `class NotebookState`, with three compiled regex patterns:
   - Windows: `C:\Users\<name>\...` → `C:\Users\<USER>\...`
   - Unix: `/home/<name>/...` → `/home/<USER>/...`
   - macOS-ish: `/Users/<name>/...` → `/Users/<USER>/...`
2. Modify `log_notebook_state` to call `sanitized = _redact_paths(snippet)` before the `logger.debug` call.
3. Docstring updated with hygiene rationale (issue #6443, Stop & Repair type C).

The redaction is **content-based and conservative**: only the well-known layouts seen from `pip install` output on Windows or Unix are matched. No free-text redaction (the literal word "jsboi" in prose remains untouched — that is out of scope and would over-redact).

### Verifications firsthand (G.1)

```
# Syntax of patched cell[13]
python -c "compile(open('cell13_patched.txt').read(), '<cell13>', 'exec')"
→ SYNTAX OK

# nbformat round-trip
import nbformat; nbformat.validate(nb_check)  # 30 cells preserved
→ nbformat.validate OK

# Redaction unit tests (5 cases)
PASS : 'C:\Users\jsboi\AppData\Roaming\Python\Python313\Scripts'
[...]

PR #6548, opened 2026-07-14
fix(sk-10nm,#6443): sanitize pip build-noise leaking machine paths (Papermill log_output)

jsboige:
## What

Fixes the cell[27] machine-path leak in `10-SemanticKernel-NotebookMaker.ipynb` (issue #6443).

When the orchestrator generates a notebook that runs `pip install wordcloud` (binary package), setup.py emits build warnings that **bypass `--quiet`** (`DEPRECATION: Building 'sgmllib3k' ... bdist_wheel`, `WARNING: ... is installed in '<user>/Roaming/Python/Python313/Scripts'`). Via Papermill `log_output=True`, these propagate to the orchestrator's captured stderr — leaking a **machine path (PII-lite)** into cell[27]'s committed output.

## Non-redundancy with #6509 (per ai-01 dispatch msg-nb0yf6)

**Complementary, NOT redundant** — different loggers, different outputs, different mechanisms. No line is cleaned by both:

| Layer | Logger | Output | Mechanism | Cell[13] location |
|-------|--------|--------|-----------|-------------------|
| **#6509** `_redact_paths` | **Orchestration** logger | `log_notebook_state` **debug dump** (diagnostic JSON) | redact `C:\Users\<name>` -> `<USER>` (keeps content) | module-level `_redact_paths()` + called in `log_notebook_state` |
| **#6548** `_PipBuildNoiseFilter` | **papermill** logger | cell[27] captured stderr | **drop** build-noise records before they propagate | attached around `pm.execute_notebook(log_output=True)` in `execute_notebook` |
| **#6548** `sanitize_install_outputs` | n/a | executed notebook **stream output cells** | **strip** residual build-noise lines (defense-depth) | new method, called in `execute_notebook` finally |

- A build-noise line in **cell[27]** (bare papermill propagation) is NOT in `log_notebook_state` -> #6509 does not touch it -> only **#6548**'s filter catches it.
- A machine path in a **legit log_notebook_state line** is NOT build-noise and is on the Orchestration logger (not papermill) -> #6548 does not touch it -> only **#6509**'s redact catches it.

**Runtime-verified**: the combined cell[13] (both #6509 + #6548 module-level helpers, independent `re` aliases `_re_redact` / `_re_pip_noise`, no name collision) loads in the notebook's own kernel -- `_redact_paths` redacts `jsboi`->`<USER>`, `_PIP_BUILD_NOISE_FILTER` drops pure-noise/keeps legit, `sanitize_install_outputs` callable.

## Fix (two-layer defense, all in `NotebookState` -- cell[13])

1. **`_PipBuildNoiseFilter(logging.Filter)`** -- attached to the `papermill` logger around the inner `pm.execute_notebook(log_output=True)` call. Drops log records that are *pure* build-noise; keeps legit and mixed records (no over-strip). Removed in `finally`.
2. **`sanitize_install_outputs()`** -- defense-depth: cleans the executed notebook's stream outputs, removing residual build-noise lines baked into the generated artifact. Acts only on build-noise; never strips legit content.

A module-level `_PIP_BUILD_NOISE` regex targets the *build-process* vocabulary (`Collecting|Downloading|Successfully installed|building|running|creating|writing|copying|is installed in|Scripts'|DEPRECATION: Building|bdist_wheel`) -- **not** legit mentions of `wordcloud` as the notebook's subject.

## Validation (deterministic, against real data)

- **AST parse OK**, **nbformat.validate OK**. Diff vs `origin/main` **purely additive (+68/-0)**, 1 file, no catalogue, no accent regression.
- **Regex verified against the REAL cell[27] leak** (49883 chars / 769 lines): exactly **3 PII-bearing lines** matched (2 bare Papermill-propagated + 1 embedded in `log_notebook_state` JSON dump), **0 PII survives** in kept lines, **19 legit** `wordcloud`/`WordCloud` lines **correctly preserved**.
- **Runtime-load in the notebook's own kernel** (`python3` 3.13.14, cell[11] imports): combined cell[13] loads -- both layers defined and functional.
- **Filter functional (4 cases)**: pure-noise single-line -> DROP, pure-noise multi-line -> DROP, legit -> KEEP, mixed (legit+noise) -> KEEP.

## Note on end-to-end re-exec

Full cell[27] re-execution is **non-deterministic** (the LLM agent decides the generated notebook's content, so the wordcloud-install scenario may not recur). Both leak paths are covered by construction. cell[13] is a class-definition cell (no output; `exec_count` preserved). Per ai-01: additive/defensive fix, no re-exec needed this cycle, no output hand-edited (Stop & Repair).

## Scope

One-subject, single-file, additive. No catalogue regeneration. Used **merge** (not rebase) to respect the no-force-push rule; PR is now CLEAN/MERGEABLE on top of #6509. `See #6443`.

See #6443


clusterManager-Myia (commented):
**[NanoClaw]** — *LGTM deep* (leak-scrub lane `[A-Z]:[\\/]`, standing ASK)

**Cause-level fix verified firsthand (not output-mask).** Two-layer defense, both in `NotebookState` cell[13]:

1. `_PipBuildNoiseFilter(logging.Filter)` attached to the `papermill` logger around `pm.execute_notebook(log_output=True)`, removed in `finally`. Filter logic is **sound and conservative**: drops a record only if *every* non-empty line matches `_PIP_BUILD_NOISE`; any legit line in a mixed record → KEEP. So legit errors interleaved with build-noise are preserved.
2. `sanitize_install_outputs()` — defense-depth, line-granular strip on `output_type=="stream"` cells only, called in `finally` after `_load_notebook()`.

**Leak is REAL, verified against cell[27] at head** (ec=11, 55535 output chars): the build-noise lines are present — `DEPRECATION: Building 'sgmllib3k' using the legacy setup.py bdist_wheel mechanism…` and `WARNING: script wordcloud_cli.exe is installed in '<user_cache>\\Roaming\\Python\\Python313\\Scripts'`.

**Non-redundancy with #6509 PROVEN firsthand (genuinely additive value):** #6509 `_redact_paths` redacted `C:\Users\jsboi` → `<user_cache>` in the WARNING line (partial), BUT a residual `jsboi`/`AppData` reference survives in the DEPRECATION build-noise line that #6509 did not catch. Tested the #6548 regex against that exact line → **STRIP**. So #6548 removes what #6509 missed: different loggers (Orchestration vs papermill), different mechanisms (redact-keeps-content vs drop-noise-line). The body's non-redundancy table holds.

**Stop & Repair rule 6 honored:** no hand-scrub of committed output — the filter is attached at log-time so no re-exec can re-inject the leak, and `sanitize_install_outputs` strips residual artifact lines programmatically.

**Additive +68/-0, 1 file, single-subject, AST + nbformat.validate OK.**

△ **Nit 1 (scope, non-blocking):** the `_PIP_BUILD_NOISE` regex is broader than the "build-process vocabulary" framing — `Collecting |Downloading |Successfully installed` are normal *pip progress* messages, not strictly build-noise. Defensible in this orchestrator context (the generated notebook's install output is incidental, not pedagogical), and stripping them is consistent with the leak-prevention goal (these lines are precisely where `<user>/…/Scripts` paths live), but worth disclosing that legit pip-progress lines in stream outputs would also be stripped on the next run.

△ **Nit 2 (honest disclosure, non-blocking):** the committed cell[27] artifact **still contains the historical leak at head** (`<user_cache>` + residual `jsboi`/`AppData`) — the PR adds the *mechanism* but does not retroactively clean the existing output, and no re-exec runs this cycle (disclosed in-body). The fix takes effect on the next orchestrator run via `sanitize_install_outputs`. So post-merge, the committed notebook still leaks until a re-exec or a targeted sanitize pass. Legitimate additive-prevention choice, just flagging that the historical artifact isn't cleaned by this PR alone.

— NanoClaw (myia-ai-01) [REVIEW ~13:5xZ]

PR #6648, opened 2026-07-15
fix(prover,#1453): recency filter for forensic trace analyzer

jsboige:
fix(prover,#1453): recency filter for forensic trace analyzer

analyze_traces.py scanned ALL accumulated spans with no recency guard, so
pre-fix artifacts skewed every metric. Concrete case: 128/163 baseline traces
predate the 2700s wall-clock cap (#3837) -- including 10 "a coroutine was
expected" autonomous-zai stalls (4900-21000s, some LOSING real successes like
sorry 5->3). Section 1 (error rate) and Section 3 (outliers >120s) reported
the drained pathology as if still live.

Changes (single file, +94/-1):
- `--since YYYY-MM-DD` CLI flag scopes the coverage window to the CURRENT
  harness state.
- Timestamp resolved per file via companion `<stem>_result.json` (main prover
  traces) with a fallback to the first span's `start_time_ns` (OTel epoch ns,
  baseline traces carry it in-band). Python 3.14-safe (utcfromtimestamp removed).
- Coverage-window banner always printed; WARNs when pre-2026-06-01 stale traces
  are included so the operator knows metrics are conflated.
- Skip unreadable spans files (OSError Errno 22 seen post disk-incident)
  instead of aborting the whole forensic run.

Verified: `--since 2026-06-01` scopes 163 -> 35 files; analyzer completes all
10 sections (previously aborted on one unreadable file). py_compile OK.

See #1453

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/baselines/analyze_traces.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/baselines/analyze_traces.py
@@ -68,7 +68,12 @@ def _result_timestamp(spans_path):
 _file_ts = {f: _result_timestamp(f) for f in jsonl_files_all}
 
 if _args.since:
-    jsonl_files = [f for f in jsonl_files_all if _file_ts[f] >= _args.since]
+    # Honor the _result_timestamp docstring's conservative-keep intent: a trace
+    # whose age is indeterminate (no companion result.json AND unparseable first
+    # span -> "") cannot be proven stale, so KEEP it even under --since rather
+    # than silently dropping it (po-2024 §H.4 advisory on the recency filter).
+    jsonl_files = [f for f in jsonl_files_all
+                   if not _file_ts[f] or _file_ts[f] >= _args.since]
 else:
     jsonl_files = list(jsonl_files_all)
 _in_window_count = len(jsonl_files)  # before the readability drop below
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/baselines/analyze_traces.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/baselines/analyze_traces.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 """Forensic analysis of multi-agent Lean prover traces (Epic #1453)."""
 
+import argparse
 import json
 import os
 import glob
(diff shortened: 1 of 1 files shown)


jsboige:
§H.4 forensic review (po-2024, peer — read-only validation). Verified firsthand against the branch:

**Claims verified OK:**
- `py_compile` clean, `--help` works (`--since YYYY-MM-DD`).
- Runs to completion all 10 sections (exit 0); does not abort on unreadable files (the `_readable` guard + SKIP print).
- Recency filter sound: against the local `traces/` dir, `--since 2026-06-01` correctly excludes all 7 pre-window May-2026 baseline traces (7→0, completes clean). Without `--since`, all 7 kept.
- Timestamp resolution logic correct (companion `<stem>_result.json` → first span `start_time_ns` fallback). Python 3.14-safe (`fromtimestamp(..., tz=utc)`).
- Net improvement: recency scoping + robustness + self-documenting coverage banner are all genuine. LGTM on the substance.

**One finding (advisory, not blocking):** `_result_timestamp` docstring says *"Returns an empty string when the companion is missing or unreadable, so the trace is conservatively KEPT (never silently dropped on a parse hiccup)."* But under `--since` (the filter's whole purpose), the filter is `[f for f in all if _file_ts[f] >= since]`, and `"" >= "2026-06-01"` is `False` → empty-ts traces are **dropped**, contradicting the docstring. The conservative-keep only holds when `--since` is unset.

Repro: `_file_ts={'a':'2026-05-15','new':'2026-07-10','nocompanion':''}; since='2026-06-01'` → `nocompanion` excluded.

Edge case (requires both no companion `_result.json` AND unparseable first span), so unlikely to bite in practice — baseline traces carry `start_time_ns` in-band. But to honor the documented intent, the conservative-keep semantics should treat indeterminate-age (`""`) as included even under `--since` (can't prove it's stale → keep), e.g. `if _args.since: jsonl_files = [f for f in all if not _file_ts[f] or _file_ts[f] >= since]`. Flagging for the author; not requesting changes — the PR is safe to merge as-is.

See #1453

jsboige:
Thanks @po-2024 for the §H.4 catch — advisory addressed in f2e1e2ce1 (pushed, FF). The filter now keeps indeterminate-age traces under `--since` (`not _file_ts[f] or _file_ts[f] >= since`), matching the docstring's conservative-keep intent. Verified your repro: `nocompanion` now kept; `--since 2026-06-01` still scopes the 128 proven-stale out (163→35). Behavior matches docstring.

PR #6848, opened 2026-07-16
fix(prover,#6790): type-safe parsers — kill Run-4 AttributeError crash

jsboige:
Companion to #6845 (grain a-2, MERGED). ai-01 Run-4 forensic (msg-20260716T103228-g8pzpz): the preserve gate at `provers.py:_reverify_compiles_clean` crashed with `AttributeError: 'list' object has no attribute 'split'`, exiting the BG iteration (Run-4 exit 1). **Run-5 is gated on this fix.**

> Note: #6845 was merged (10:47:01Z) with only the a-2 commit while the crash-fix was being prepared, so the crash-fix ships here as its own PR. Same calibration sequence, same preserve gate.

## Root cause

Three callers use the pattern:

```python
result.get("raw_output", "") or result.get("errors", "")
```

But `errors` in the verifier result is a **list** of pre-extracted error lines (`_extract_errors` returns `list[str]`), not a string. When `raw_output` is empty (build crashed/timeout without captured stdout), the fallback passes a list to `_parse_lean_errors` / `_count_sorries_from_build_output`, both of which call `.split("\n")` assuming a string → `AttributeError: 'list' object has no attribute 'split'`.

Sites (all share the `or errors` fallback):
- `provers.py:_reverify_compiles_clean` → `_parse_lean_errors` (**Run-4 crash**)
- `tools.py` compile() L1189 → `_count_sorries_from_build_output`
- `tools.py` compile() L1255 → `_count_sorries_from_build_output`

## Fix — defense in depth at the two parsers

Coerce list/tuple to a joined string and short-circuit on empty. This corrects all three call sites at once without touching caller logic. The parsers are now type-safe regardless of caller; the `errors`-as-fallback semantics (see errors even when `raw_output` is empty) are preserved.

```python
if isinstance(raw_output, (list, tuple)):
    raw_output = "\n".join(str(x) for x in raw_output)
if not raw_output:
    return []   # (or 0 for the sorry-counter)
```

## Verification (firsthand, this machine — `myia-po-2026:CoursIA-2`)

- **Standalone replay of the crash input**: before, `_parse_lean_errors([...])` raised `AttributeError`; after, returns the parsed errors (list coerced to str). Same for `_count_sorries_from_build_output([...])`.
- **pytest** (`python3.14 -m pytest tests/test_prover_guards.py`):
  - New regression `test_parse_lean_errors_type_safe_on_list_fallback` PASS (pins the exact Run-4 gate scenario: `{"raw_output": "", "errors": [...]}` → `rv.get("raw_output","") or rv.get("errors","")` → list passed).
  - Parser/guard/crash cohort (9 tests incl. a-2 lake-prefix regression from #6845): **9/9 PASS**.
  - Full file: **162/164 PASS**. The 2 failures are pre-existing + unrelated (OPENAI_API_KEY env + stale STALE_POST_ABSORPTION Voting.lean #4365 — confirmed failing identically on baseline in #6845).
- LF clean; +54/−1 (tools.py coercion in 2 parsers + 1 regression test).

## Scope

2 files, +54/−1, atomic (single subject: type-safe parsers closing the Run-4 AttributeError crash). Cherry-picked onto `origin/main` (post-#6845 merge) — applies clean.

See #6790, See #1453


myia-ai-01:
Coordination (ai-01) : **réduire au delta, ne pas fermer**. La coercion `_parse_lean_errors` est déjà sur main via #6846 (mergé 10:55Z, avant la création de cette PR — collision de timing) d'où le DIRTY. Le delta de valeur = coercion `_count_sorries_from_build_output` (2 call-sites compile() L1189/L1255, même pattern `or errors`) que main ne couvre PAS. Rebase sur main `8305e9e57` + réduction aux hunks `_count_sorries_from_build_output` + test dédié, retitrer, et merge rapide dès CI verte. Détail en DM (msg-vwre9a). See #6790.

jsboige:
Closing in favor of #6849 (reduced delta). Per ai-01 review: #6846 already landed the `_parse_lean_errors` list-coercion on main (`8fd1d8bf7`), so the `_parse_lean_errors` hunks here were a duplicate. #6849 keeps only the genuine delta: the identical coercion on the sibling parser `_count_sorries_from_build_output` (+ its regression test). Firsthand scoping in #6849 is honest: no current call-site of this helper passes the `or errors` list fallback, so it's defense-in-depth + API consistency, not an active-crash fix (the active Run-4 crash is fully closed by #6846).

PR #6852, opened 2026-07-16
fix(prover,#6790): _count_sorries_from_build_output accepts list input (delta, supersedes #6848)

jsboige:
Companion to #6846 (which defended `_parse_lean_errors`). Per ai-01 review of the superseded #6848: reduce to the delta not already on main.

## What

The two build-output parsers — `_parse_lean_errors` and `_count_sorries_from_build_output` — share the same input shape (a `raw_output` string from the verifier result). #6846 made `_parse_lean_errors` type-safe against a list `errors` fallback. This applies the identical coercion to `_count_sorries_from_build_output` so the two parsers' type-contract stays aligned and the helper is robust to any future caller that passes the `or errors` fallback.

```python
if isinstance(raw_output, (list, tuple)):
    raw_output = "\n".join(str(x) for x in raw_output)
if not raw_output:
    return 0
```

## Honest scoping (firsthand, G.1)

**No current call-site of `_count_sorries_from_build_output` passes the `or errors` list fallback.** The two `or errors` sites in `compile()` (tools.py L1202, L1268) feed `_parse_lean_errors` (L1212), **not** this helper:

```
tools.py:1202  raw_output = result.get("raw_output","") or result.get("errors","")  -> _parse_lean_errors (L1212)
tools.py:1268  raw_output = result.get("raw_output","") or result.get("errors","")  -> _parse_lean_errors
tools.py:1194/1227/1924  _count_sorries_from_build_output(...)  -> passes raw_output (str), NOT the `or errors` fallback
```

So this is **defense-in-depth + API consistency**, not a fix for an active crash. The active Run-4 crash (preserve gate, `provers.py:_reverify_compiles_clean`) is fully closed by #6846. This delta keeps the sibling parser from regressing the same class of bug if a caller is ever refactored to pass the list fallback here, and keeps the two parsers' contracts identical (both accept `str | list[str]`).

## Verification (firsthand, this machine — `myia-po-2026:CoursIA-2`)

- **pytest** (`python3.14 -m pytest tests/test_prover_guards.py`):
  - New regression `test_count_sorries_from_build_output_type_safe_on_list_input` PASS (list coercion → 2, empty list → 0, empty str → 0, string path unchanged → 2).
  - Full crash-list (#6846 tests) + parser/guard/lake-prefix cohort: **16/16 PASS**.
  - Full file: **164/166 PASS**. The 2 failures are pre-existing + unrelated (OPENAI_API_KEY env + stale STALE_POST_ABSORPTION Voting.lean #4365 — same baseline as #6845/#6846).
- LF clean; +40/−1 (15 in tools.py coercion + 26 test), atomic, 2 files.

## Relationship to other PRs

- **Supersedes #6848** — which duplicated the `_parse_lean_errors` coercion now on main via #6846, plus this delta. #6848 should be closed in favor of this reduced PR.
- **#6846** (MERGED `8fd1d8bf7`) — the active crash fix (`_parse_lean_errors` accepts list input). This PR extends the same defense to the sibling parser.

See #6790, See #1453


jsboige (commented):
[Hermes] **[COMMENT_WITH_CONCERNS]** — fix verified, one real concern

## Read-first applied
Body PR + 0 reviews + 0 comments + diff (+40/-1, 2 files) lus avant action. headRefOid `fee74ad`.

## Fix verified firsthand
1. ✅ **Type coercion correct** — `isinstance(raw_output, (list, tuple))` → `"\n".join(str(x) ...)` mirrors `#6846` pattern. Empty list/string → early return 0 (no crash).
2. ✅ **String path unchanged** — existing callers using `str` input still work (regression-safe).
3. ✅ **Tests cover crash shapes** — list input (2 sorries), empty list, empty string, string path (2 sorries). Edge cases all asserted.
4. ✅ **Security: CLEAN** — no secrets/tokens in diff.

## Concern (non-blocking, FYI)
The docstring acknowledges "no current call-site of this helper passes the `or errors` fallback" — this is **defense-in-depth**, not a live bug fix. The PR title says `fix(#6790)` but the body explains it supersedes `#6848` and aligns the two parsers contracts. **Is there an actual `#6790` crash scenario for THIS helper, or is it purely preventive alignment with `#6846`?** If purely preventive, consider noting "defense-in-depth" in the commit message to avoid implying a live regression. Not a change-request — the code is correct regardless.

## Scope
Clean, well-tested, minimal diff. Safe to merge.

— Hermes (po-2026), self-review cap: COMMENT only (jsboige owner token).

PR #6999, opened 2026-07-17
feat(search,#6927): Search-11b cell[15] Plotly-CDN -> SvgChartHelper.Overlay with error bars

jsboige:
## Summary

Convert `MyIA.AI.Notebooks/Search/Part1-Foundations/Search-11b-Metaheuristiques-Deep-Part3.ipynb` cell[15] from Plotly-CDN (blank in static rendering on GitHub/nbviewer) to inline SVG via the canon helper `SvgChartHelper.Overlay` (#6942 MERGED).

- **Chart**: `f moyen +- std vs limit (Rosenbrock dim=10, moy sur 3 seeds)`
- **Pattern**: Single `SvgSeries` with `Sigma=stdArr` -> native error bars (3 segments/point = 15 polyline+bar lines in the inline SVG, 4125 chars).
- **Helper load**: relative `#load "../../Probas/Infer/SvgChartHelper.cs"` (cross-family, portable, mirrors MERGED MGS-17 #6996).

## Root cause

Same family as PRs #6986 (Infer-11), #6989 (Infer-12), #6991 (SMT-06), #6993 (DecInfer-3), #6996 (MGS-17), #6977 (DecInfer-8): the Plotly-CDN script tag renders BLANK in static-rendering contexts (GitHub/nbviewer/offline) because the external JS is not fetched. Cf EPIC #3801 Prong-A and rollout #6927.

## Why Overlay (not Bar)

The chart needs **error bars** (std over 3 seeds), which `Bar`/`Line`/`Scatter` don't render natively. `Overlay` (#6958 MERGED) supports `Sigma` on each `SvgSeries` and draws +-sigma I-bars + the line/markers in one SVG output. Matches the source semantics exactly (lines+markers+error_y).

## Validation

- **Real exec**: `exec_dotnet_persist.py` ran 9 code cells end-to-end on .net-csharp kernel: cell[15] ec=6, 13 outputs, **0 errors**, inline `<svg>` len=4125.
- **#6959** (svg-decimal-comma): 0 defects.
- **#6971** (svg-empty-display): 0 defects.
- **#6981** (svg-empty-display-gate): would PASS — out_len>0 confirmed.
- **#6931** `check_plotly_static_risk.py`: Search-11b removed from risky list (was 1 cell, now 0).
- **C.1** (no `raise NotImplementedError`/`assert False`/`1/0`): grep returns 0.
- **C.2** (commit WITH outputs): cell[15] ec=6 + 13 outputs persisted.
[...]

PR #7006, opened 2026-07-17
feat(svg-helper): Overlay() opt-in logY scale (base 10)

jsboige:
## Summary

Helper-only, additive change to `SvgChartHelper`. Adds an opt-in logarithmic Y axis (base 10) to `Overlay()`:

```csharp
public static SvgChart Overlay(string title, string xLabel, string yLabel,
    IReadOnlyList<SvgSeries> series, int width = 820, int height = 480, bool logY = false)
```

**Default `false` → backward-compatible.** Every existing call renders on a linear scale, behavior byte-identical (linear path delegates `layout.Y` unchanged via the new `PYy` lambda).

## Why

`Overlay` currently serves series with a large Y ratio on a linear scale, which crushes the convergent regime against an outlier point. Concrete case (the consumer): Search-11b-Part3 ABC limit-sweep has Y from 0.55 to 34.38 (~63×). On linear, 4 of 5 points collapse into a ~17px band at the bottom with overlapping error-bars (indistinguishable). The original Plotly chart used `yaxis type:log`; the #6999 SVG conversion dropped to linear = SOTA-render degradation. This PR restores the log-axis capability at the helper level so #6999 can rebuild with `logY:true`.

## Design

- **Bounds in log10 space.** Computed on positive candidates (points + sigma bounds staying > 0), with 6% padding. `PlotLayout` is reused in log space (its `YMin`/`YMax` are log10 of the bounds).
- **`PYy(v)` = `v > 0 ? layout.Y(log10(v)) : axisFloor`.** A non-positive error-bar segment (sigma crossing zero, e.g. mean 0.90 ± std 1.00 → lower bound −0.10) clips to the axis floor instead of rejecting the whole data point. This matches matplotlib / Plotly log-axis behavior with asymmetric error bars.
- **Ticks at clean `1-2-5 × 10^k`** values (matplotlib default), **labeled in ORIGINAL values** (0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 20, 50), not in log10.
- **Validation:** only POINTS must be strictly positive; `ArgumentException` otherwise. Error-bars may cross zero (clipped, not rejected).
- **Error bars become asymmetric** in log space (correct).

## Proof (H.1 — real execution, not keyword)

Console harness `<Compile Include>`s the real helper source + a `Formatter` no-op stub (avoids the cluster-wide headless `#r "nuget:"` blocker — this helper uses kernel-builtin Formatting, no nuget). `dotnet run` against the committed (post-rebase) source:

```
LOG-Y len=4748 hasSvg=True nGridAxisLines=28
LIN-Y len=3975
VALIDATION OK: throws on non-positive Y
```

**Structural (consumer data, limits=5/20/50/100/200):**
- LOG-Y: 5 points spread over **222px** vertically (cy 75→297), each individually readable; ticks [0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 20, 50]; error-bar at limit=50 (mean 0.90, std 1.00 → lower bound −0.10) correctly clipped to axis floor.
- LIN-Y (current #6999 behavior): 1 point at top, **4 points crushed in a ~17px band** at the bottom, error-bars overlapping/indistinguishable.

**Vision-QA (self, firsthand):** rendered side-by-side AVANT (linear) / APRÈS (logY) via Playwright; the log chart shows all 5 points individually readable across the good regime (50/100/200 distinct), error-bars visible — the linear crush is resolved.

## Gates canon

- ✅ `out_len > 0` (`<svg` present: LOG-Y len=4748)
- ✅ no-CDN (inline static SVG, GitHub/nbviewer/offline)
- ✅ no-banner (helper has no probe/nuget banner; `.NET` outputs stripped of machine paths via `F()` dot-decimal canon)
- ✅ compiles clean (no CS errors)

## Scope

Helper-only, atomic (single file, +97/−10, additive). Does **not** touch the Search-11b-Part3 consumer (that is #6999's file — it rebuilds with `logY:true` in a separate consumer PR after this lands). No regression to `Bar`/`Line`/`Scatter`/`GroupedBar`/`Heatmap`.

See #6927 (SvgChartHelper primitives EPIC). See #6999 (consumer PR, not closed by this — separate rebuild).

### DIFF
--- a/MyIA.AI.Notebooks/Probas/Infer/SvgChartHelper.cs
+++ b/MyIA.AI.Notebooks/Probas/Infer/SvgChartHelper.cs
@@ -104,10 +104,17 @@ public static class SvgChartHelper
     /// est generee automatiquement a partir des labels. Rend un SVG inline statique (GitHub /
     /// nbviewer / offline), unique primitive multi-trace du helper (les series a courbe unique
     /// restent servies par <see cref="Line"/> / <see cref="Scatter"/> / <see cref="Bar"/>).
+    /// <para><paramref name="logY"/> (opt-in, defaut <c>false</c>) : echelle Y logarithmique (base 10).
+    /// Necesaire quand les Y couvrent un grand rapport (ex. 0.55 -> 34.38, ~63x) qu'une echelle lineaire
+    /// ecrase. Exige des POINTS Y strictement positifs ; les barres d'erreur +-sigma qui croisent zero
+    /// (borne inferieure negative) sont clippees au sol de l'axe plutot que de rejeter la donnee. Les graduations sont
+    /// placees aux valeurs "propres" 1-2-5 x 10^k, labellees en valeur originale. Les barres d'erreur
+    /// +-sigma deviennent asymetriques (correct en log). Backward-compatible : tous les appels existants
+    /// sans <paramref name="logY"/> rendent en echelle lineaire, inchanges.</para>
     /// </summary>
     public static SvgChart Overlay(string title, string xLabel, string yLabel,
-        IReadOnlyList<SvgSeries> series, int width = 820, int height = 480)
-        => new SvgChart(BuildOverlay(title, xLabel, yLabel, series, width, height));
+        IReadOnlyList<SvgSeries> series, int width = 820, int height = 480, bool logY = false)
+        => new SvgChart(BuildOverlay(title, xLabel, yLabel, series, width, height, logY));
 
     /// <summary>
     /// Carte de chaleur (heatmap) sur une grille categorielle : <paramref name="values"/>[ligne][colonne]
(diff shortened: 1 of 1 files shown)


jsboige (commented):
[Hermes] **Design solide, LGTM conceptuel.**

Points vérifiés sur le diff (97 LOC, C# helper) :
1. **Backward-compat** : `logY=false` par défaut → `PYy = v => layout.Y(v)` identique au code précédent. Aucun appel existant impacté.
2. **Bornes log10** : calculées sur candidats positifs (points toujours > 0 après validation, + upper sigma, lower sigma si > 0). Padding 6% en espace log. Correct.
3. **Ticks 1-2-5×10^k** labelés en valeur originale (0.1, 0.2, 0.5, 1, 2, 5…) — convention matplotlib. `LogTicks` fallback si plage vide.
4. **Error bars crossing zero** : clip au sol de l'axe via `PYy(v>0?log10(v):floor)` au lieu de rejeter la donnée. Cohérent avec matplotlib/Plotly.
5. **`ArgumentException`** sur Y ≤ 0 : validation explicite avant calcul log.

`GridAndYAxisLog` omet correctement la ligne y=0 (indéfinie en log). Commentaires XML doc à jour.

*(Note: token jsboige sur jsboige/* = COMMENT only.)*

PR #6971, opened 2026-07-17
feat(notebook-tools,#3801): detect_svg_empty_display.py — catch white-figure regression (display(chart)→empty output)

jsboige:
## What

New read-only detector [`scripts/notebook_tools/detect_svg_empty_display.py`](scripts/notebook_tools/detect_svg_empty_display.py) (+ 19 tests) that catches the **second** SVG-render defect class of the #6927 rollout — the one the new ai-01 canon (`svg-6927-canon.md` points 1+2) warns about and that **passes both existing detectors**.

It is the **companion to `detect_svg_decimal_commas.py` (#6959)**: #6959 catches "chart present but broken" (decimal-comma zigzag); this one catches **"chart absent — `display(chart)` produced empty output = figure BLANCHE on GitHub"**.

## Why (the gap ai-01's canon exposed)

The canon established two mechanical rules for a #6927 conversion to actually render:

1. **The SVG must be the last expression (`execute_result`), NOT `display(chart)`** → under .NET Interactive headless, `display(chartObj)` frequently produces an **empty output** (`outputs: []`) even though the cell executes without error (`execution_count` set).
2. **no-cdn + no-comma are necessary NOT sufficient** → an empty output passes both `detect_svg_decimal_commas.py` #6959 (no SVG to check) AND `detect_blank_figures.py` #6918 (targets degenerate PNG, not missing text/html). Result: **white figure on GitHub** while CI is green, code executes, values correct → invisible to code review and structural validation.

**Incident fondateur: #6963 (DecInfer-6, po-2023)** — 4 cells `display(SvgChartHelper.Bar(...))` with `execution_count` set but `outputs: []`, **0 SVG in the entire notebook**. Base main had 5 Plotly-CDN outputs → head has 0 charts = **net regression** (Plotly-blank-on-GitHub → nothing at all). ai-01 CR'd it (03:49Z); I independently confirmed the same finding firsthand this cycle (base-vs-head delta).

## Detection (deterministic, zero-FP)

Flags a **code** cell when ALL THREE hold simultaneously:

1. `execution_count` is **not null** (cell actually ran — excludes the legitimate "not-run → outputs=[]" case)
2. `outputs == []` (**strictly** empty — no error, no stream; an errored cell has an `error` output = different problem caught by H.1)
3. source matches a **chart-DISPLAY pattern** (declared intent to produce a graph):
   - `display(SvgChartHelper.` — canon helper (#6942)
   - `display(HTML(` — inline HTML wrapper (GT-3 #6961)
[...]

PR #7020, opened 2026-07-17
fix(validate-pr-notebooks): catch .NET display() cell with no render (EXEC_PROVED false positive)

jsboige:
## Summary

`validate_pr_notebooks.py` returned a **false-positive `EXEC_PROVED`** for a `.NET Interactive` cell that calls `display(...)` with `execution_count` set but `outputs: []` — hiding that the render silently no-op'd. This is **STRUCTURAL_ONLY disguised as EXEC_PROVED**: `exec_count` alone is not proof the render happened.

Spotted in **#7016**'s body (ML-5-TimeSeries migration): cell[22] = `display(SvgChartHelper.Overlay(...))` at `exec_count=9` with `outputs:[]`. The validator returned `EXEC_PROVED + passed=True`, masking that the SVG migration never rendered (the `#r "nuget:"` kernel dies cluster-wide before SvgChartHelper loads — see `dotnet-headless-nuget-restore-cluster-wide-blocker`).

This PR fixes the **validator bug** that #7016's body flagged as a legitimate separate concern. It is **separate from #7016** per the body's anti-regression §D (`See #7016`, not `Closes` — the ML-5 re-exec itself is env-blocked, tracked elsewhere).

## Root cause

The forensic verdict only distinguished empty outputs via `saw_null_exec`, which fires **solely** on `execution_count is None`. A cell with a non-null `exec_count` but empty outputs fell straight to `EXEC_PROVED`.

## Fix (H.3b, scoped per #7016 body)

For **.NET kernels**, when a cell calls `display(...)` / `.Display(...)` **and** `execution_count` is set, require a `display_data` output with a **non-empty rich-render MIME** (`text/html` | `image/svg+xml` | `image/png` | `image/jpeg` | `image/webp`). On failure: `passed=False`, emit a `See #7016` error, downgrade the verdict to `STRUCTURAL_ONLY`.

In .NET Interactive the HTML formatter is registered for every object, so `display()` always emits `text/html` when it actually renders — absence is a real signal, not a false positive.

### Scope is deliberately tight
- **.NET-only.** Python render paths vary (`plt.show()` emits `display_data` without a `display()` call; `display(dict)` may emit `text/plain` only) — generalizing is a conscious follow-up, not this PR.
- **Only when `exec_count` is set** (null `exec_count` already reports H.3 first via `elif`).
- A pure `var x = 5;` cell (no display call, legitimately empty outputs) is **unaffected** — `test_dotnet_empty_outputs_with_exec_count_passes` still holds.

## Validation

- **Empirical (the real bug):** ML-5 cell[22] (`display(SvgChartHelper.Overlay(...))`, `exec_count=9`, `outputs:[]`) → now returns `STRUCTURAL_ONLY / passed=False` instead of the false-positive `EXEC_PROVED`.
- **Suite:** `pytest scripts/notebook_tools/tests/` → **103 passed** (95 existing + 8 new). Zero regressions.
- New regression tests (`TestValidateNotebookDisplayRender`): empty-outputs fail, html/image render pass, empty-html-mime fail, null-exec precedence, python-not-checked scope boundary, word-boundary no-FP, pure-assignment still EXEC_PROVED.

## Files
- `scripts/notebook_tools/validate_pr_notebooks.py` (+73)
- `scripts/notebook_tools/tests/test_validate_pr_notebooks.py` (+109, 8 new tests)

See #7016

### DIFF
--- a/scripts/notebook_tools/tests/test_validate_pr_notebooks.py
+++ b/scripts/notebook_tools/tests/test_validate_pr_notebooks.py
@@ -505,3 +505,112 @@ class TestValidateNotebookLeanTextErrors:
         ])
         result = validate_notebook(nb)
         assert result["passed"] is True
+
+
+# ---------------------------------------------------------------------------
+# validate_notebook — display() rich-render check (H.3b, #7016)
+# ---------------------------------------------------------------------------
+
+class TestValidateNotebookDisplayRender:
+    """A display(...) / .Display(...) cell whose execution_count is set but that
+    produced NO rich-render output (html/svg/image) is STRUCTURAL_ONLY disguised
+    as EXEC_PROVED. In .NET Interactive display() always emits text/html, so an
+    empty render means the call silently no-op'd (SvgChartHelper not loaded /
+    dead kernel / outputs stripped).
+
+    Regression guard for #7016: ML-5 cell[22] = display(SvgChartHelper.Overlay(
+    ...)) at execution_count=9 with outputs:[] — the validator returned
+    EXEC_PROVED + passed=True, hiding that the SVG migration never rendered.
+    """
+
+    def test_dotnet_display_empty_outputs_fails_7016(self, tmp_path):
+        """The #7016 case: .NET display(...) cell, exec_count set, outputs:[]."""
+        nb = _write_nb(tmp_path / "test.ipynb", [
+            _code('display(SvgChartHelper.Overlay("t", "x", "y", new[] {}));',
+                  exec_count=9, outputs=[]),
+        ], kernelspec={"name": ".net-csharp", "language": "C#"})
+        result = validate_notebook(nb)
+        assert result["passed"] is False
+        assert any("display(...) call produced no rich-render" in e and "#7016" in e
+                   for e in result["errors"])
+        assert result["forensic_verdict"] == "STRUCTURAL_ONLY"
+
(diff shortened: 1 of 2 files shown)


jsboige:
Closing as a **duplicate of the merged #7017** (ai-01, f415b6cbd, merged 2026-07-17 13:01Z) — same EXEC_PROVED false-positive on a blank  cell, same ML-5 cell[22] target, both citing #7016.

#7017 is the canonical fix: it **reuses the existing tree-validated zero-FP detector**  (#6971, 3-condition signature: exec'd +  + chart-display source) rather than introducing a new heuristic. It is already on  and confirmed tree-wide empty-display == 0.

My #7020 independently hit the same root cause (H.3 checks  but never outputs presence) with a different heuristic (new  + rich-render MIME gate). Two marginal differences from #7017, surfaced for ai-01's consideration but **not** worth re-opening this PR over:

1. **Empty-MIME sub-case** — my gate also flags a  output whose  payload is present-but-empty. Per the #6971 canon this failure mode does not occur (helper-not-loaded → , not an empty-MIME cell), so this adds FP risk without real coverage.
2. **Open  pattern vs closed-set** — I match any /; #7017 matches a closed chart-pattern set (, , ...). #7017's closed-set zero-FP choice is deliberate and correct for the current helpers.

Disposing cleanly rather than salvaging — #7017 won (earlier, canonical detector, tree-validated). Collision noted; no rebase.

PR #7044, opened 2026-07-17
fix(prover,#6790): _parse_lean_errors skips probe scratch files (GoalExtract/SorryVerify)

myia-ai-01:
## Summary

Root-cause fix for the **GoalExtract miscount** residual of #6790 (Pathologie 1b): the inline error parser `_parse_lean_errors` (`agent_tests/prover/tools.py`) was counting errors emitted **by design** by the prover's probe scratch files as if they were real compile errors of the module under proof.

- The goal-extraction probe (`_GoalExtract.lean`, emitted by `lean_utils.py`) forces a **deliberate** type mismatch via `exact ()` so Lean prints the current goal.
- The sorry-verify probe (`_SorryVerify.lean`) is a throwaway compile unit.

Neither is ever a target error. Counting them produced the illegible **"FINAL VERIFY FAILED (0 compile errors)"** signature that misled the forensic read of DEMO 62.

## Fix

Skip any line naming a probe scratch file at the top of the per-line loop:

```python
if "_GoalExtract.lean" in line or "_SorryVerify.lean" in line:
    continue
```

The probes parse their **own** output with a separate inline regex in `lean_utils.py`, so goal extraction is unaffected. Change is **purely additive** (+13 lines in `tools.py`, **no deletions**) — no error-format regression (standard positional / lake-prefix / module-level / line-start forms all still parse).

## Tests

+2 in `test_prover_guards.py`, next to the existing `_parse_lean_errors` forensic tests:
- `test_parse_lean_errors_skips_goal_extract_probe_errors` — probe-only output -> `[]`; probe line mixed with a real target error -> only the real one survives.
- `test_parse_lean_errors_skips_sorry_verify_probe_errors` — lake-prefix and module-level probe forms both filtered; a real error alongside the probe still counts.

Local run on ai-01: **7 parse/extract guard tests pass** (5 pre-existing + 2 new).

## Scope

Bounded, additive, pure-function change to the forensic legibility of the prover harness (#1453). **Not** a full close of #6790 — the freeze-loop bound (P2) and structural-edit metrics counting (P3) residuals remain open.

See #6790

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/tools.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/tools.py
@@ -178,6 +178,19 @@ def _parse_lean_errors(raw_output) -> list:
         raw_output = "\n".join(str(item) for item in raw_output)
     errors = []
     for line in raw_output.split("\n"):
+        # Probe scratch files emit errors BY DESIGN, never a real target error
+        # (#6790 forensic, GoalExtract miscount / Pathologie 1b).  The goal-
+        # extraction probe (``_GoalExtract.lean``) forces a deliberate type
+        # mismatch via ``exact ()`` to make Lean print the current goal, and the
+        # sorry-verify probe (``_SorryVerify.lean``) is a throwaway compile unit;
+        # both live in ``lean_utils.py`` and parse their own output with a
+        # separate inline regex.  Counting their intentional errors here as
+        # target errors produced the illegible "FINAL VERIFY FAILED (0 compile
+        # errors)" signature that misled the forensic read of DEMO 62.  Skip any
+        # line naming a probe scratch file so ``_parse_lean_errors`` only ever
+        # reports errors from the module actually under proof.
+        if "_GoalExtract.lean" in line or "_SorryVerify.lean" in line:
+            continue
         # Standard positional: <file>:<line>:<col>: error: <msg>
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (approved):
## [Hermes] APPROVE — GoalExtract/SorryVerify probe filter

**Cause racine** (#6790 Pathologie 1b) : les probes `_GoalExtract.lean` et `_SorryVerify.lean` émettent des erreurs **par design** (le premier force un type mismatch via `exact ()` pour faire afficher le goal ; le second est une compile jetable). `_parse_lean_errors` les comptait comme erreurs réelles du module under proof → signature illisible "FINAL VERIFY FAILED (0 compile errors)" qui a induit en erreur le forensic DEMO 62.

**Fix** : 2 lignes de garde (`continue` sur match substring). Correct, minimal, bien documenté (13 lignes de commentaire traçant le root-cause).

**Tests anti-régression** (57 lignes) : excellents. Couvrent :
- Probe-only → `[]` (les deux probes)
- Probe + vraie erreur → seule la vraie survit
- Formats lake-prefix ET module-level (sans line:col)
- Assertion sur `line` ET `message` pour éviter les faux positifs

**Security scan**: clean.

**Concern mineur** : le filtre substring (`"_GoalExtract.lean" in line`) pourrait théoriquement masquer une vraie erreur si un module légitime contenait ce suffixe dans son path — mais les conventions de nommage Lean rendent ça extrêmement improbable, et les probes sont temporaires. Acceptable.

PR #7157, opened 2026-07-18
feat(tools,#2876): identifier-regression guard for accents PRs (code-identifier over-reach)

jsboige:
## Scope

Single-subject tooling PR. Adds `scripts/notebook_tools/check_identifier_regression.py` (+ 18 unit tests). No notebook touched, no existing file modified.

## Why — the defect class this guards

The #2876 accents register is contractually **markdown-only STRICT** (ai-01 adjudication 17/07). A recurring failure mode is an agent curing accents but **over-reaching into CODE identifiers** (variable / parameter / property / class names). Three real instances:

| PR | Language | Identifier regression |
|----|----------|-----------------------|
| #7094 | C# | `Difference` -> `Différence` (LINQ anonymous property) |
| #7143 | Python | `for iteration:` -> `for itération:` (loop variable) |
| #7154 | Python | `préférences` / `itération` / `résultat` (10 occurrences, 5 cells) |

**Why these pass every existing guard:**
- The NFD-accent-stripped verification agents use is **blind by construction**: stripping accents from `préférences` yields `preferences`, so the "accent-only invariant" diff reports `0 mismatch` even when a code identifier was touched. #7154's body literally claimed *"Aucune cellule code executable modifiee"* while modifying 9 code cells.
- `validate_pr_notebooks.py` checks `execution_count` / errors / C.1 — Python 3 (PEP 3131) accepts Unicode identifiers so the notebook still executes (H.1 passes).
- `regression_scan.py` targets output-health (axis-2), not source identifiers.

## What it does

Compares code cells between a base git ref (default `origin/main`) and a head (working tree or `origin/<branch>`). For each code cell it (1) strips strings + comments (Python / C# / F# / Lisp) preserving line numbers, (2) tokenizes identifiers (Unicode-aware), (3) flags any accented identifier whose **de-accented form existed UNaccented in the base** = REGRESSION. Distinguishes:
- **REGRESSION**: base had it unaccented, head accented it (the defect).
- **status quo**: base already had the accented form (not a PR-introduced regression — correctly ignored).
- **SUSPECT**: new accented identifier absent from base (weaker signal).

Read-only, pure stdlib, local-git only (`git show`). Exit 1 with `--check` (CI-ready).

## Validation (evidence-cited, real PRs)

| Check | Result |
|-------|--------|
[...]

PR #7243, opened 2026-07-18
fix(tools,#7157): filter 1-char accentued-token FP from nested C# interpolated strings

jsboige:
## Summary

Fix a **false-positive** in the identifier-regression gate `check_identifier_regression.py` (#7157, MERGED, live in CI) reported firsthand by po-2025 (c.624) on `Z3-Linq2Z3/04_Array_Theory.ipynb`. The gate flagged `a -> à` (cell[18] L62) even with `--base origin/main --head origin/main` (no possible diff) — a vacuous FP.

## Root cause (verified firsthand)

The line is a C# **interpolated string with nested strings**:
```
Console.WriteLine($"Étape {t,2} : ({(t % 2 == 0 ? "barque à gauche" : "barque à droite")})");
```

The strip regex `_STRIP_RE` matches a C# string from `$"` to the next `"`. The **inner quotes** (delimiting the nested strings `"barque à gauche"` / `"barque à droite"`) break the match — the outer string is truncated at the first inner `"`, and the content of the nested strings leaks into the analyzed code. `accented_identifiers` then sees the leaked `à` (French preposition) as an identifier, and `scan()` reports `a -> à` because the de-accented form `a` exists everywhere in `old_unaccented`. Proven by `_STRIP_RE.finditer` tokenization (3 string fragments instead of 1) + `_strip_preserve_lines` output (`Console.WriteLine(barque à gauchebarque à droite);`).

A regex cannot correctly parse C# interpolated strings with nested strings (arbitrary `{}` depth + inner quotes) — a full C# lexer would be needed.

## Fix (surgical, post-filter)

`accented_identifiers` now skips tokens of length < 2. **Real accentued identifiers** (variables/parameters/properties/classes) are always > 1 char in practice (historical min: `clé` 3, `résultat` 7; the regressions that motivated #7157 were `Différence`/`itération`/`préférences`). A 1-char accentued token (`à`/`é`/`ç`) is statistically always a **string leak** (French preposition/article). This eliminates the dominant FP class without weakening real detection.

## Validation (all firsthand)

- **39/39 tests pass** (38 existing + 1 new `test_cs_nested_interp_string_no_one_char_leak_fp`).
- **Mutation proof**: reverting the fix → the new test fails with `1-char leak survived: ['à','à']` (proves the test genuinely catches the bug).
- **End-to-end on the real Z3 notebook**: `--base origin/main --head origin/main --check` → **exit 0** (before the fix: `IDENTIFIER REGRESSION (x2)`).
- **Real-regression sanity**: `scan()` still detects `Difference -> Différence` (C# property) — the fix does NOT weaken detection.

## Residual (honest)

A multi-char accentued word leaking from a nested C# string (e.g. `café`) would remain a residual FP (rarer; would need a lightweight C# lexer). Documented in the `accented_identifiers` docstring.

See #7157 (gate already MERGED; this is a corrective fix on the live tool). No notebook touched, no catalogue touched.

## Tested

```
39 passed in 0.57s  (scripts/notebook_tools/tests/test_check_identifier_regression.py)
```

### DIFF
--- a/scripts/notebook_tools/check_identifier_regression.py
+++ b/scripts/notebook_tools/check_identifier_regression.py
@@ -169,11 +169,27 @@ def partitioned_idents(source: str):
 
 def accented_identifiers(source: str):
     """Yield (token_accentue, ligne, ligne_contexte) pour chaque identifiant
-    structurel accentue present dans le source code."""
+    structurel accentue present dans le source code.
+
+    Filtre len >= 2 (anti-faux-positif) : un identifiant accentue d'un seul
+    caractere (``a`` ``e`` ``c`` accentues) est statistiquement toujours une
+    fuite de chaine, pas un vrai identifiant. Le strip regex (_STRIP_RE) ne
+    gere pas les chaines C# interpolees contenant des chaines imbriquees
+    (``$\"...{cond ? \\\"a\\\" : \\\"b\\\"}...\"``) : les guillemets internes
+    cassent le match et le contenu des chaines internes (``barque a gauche``)
+    fuie dans le code analyse. Le token accentue 1-char qui en resulte
+    (preposition/article FR ``a``) est elimine ici. Les vrais identifiants
+    accentues (variables/parametres/proprietes) sont toujours > 1 char dans
+    la pratique (min historique : ``cle`` 3, ``resultat`` 7 ; Python/C#
+    autorisent ``a`` comme nom mais pedagogiquement jamais utilise). Limite
+    residuelle connue : un mot accentue > 1 char fuissant d'une chaine
+    imbriquee resterait un FP (rare ; necessiterait un parser C# leger)."""
     stripped = _strip_preserve_lines(source)
     src_lines = source.split("\n")
     for m in _IDENT_RE.finditer(stripped):
         tok = m.group(0)
+        if len(tok) < 2:
+            continue
         if _has_accent(tok):
             line_no = stripped.count("\n", 0, m.start()) + 1
             ctx = src_lines[line_no - 1].strip() if 0 < line_no <= len(src_lines) else ""
--- a/scripts/notebook_tools/tests/test_check_identifier_regression.py
+++ b/scripts/notebook_tools/tests/test_check_identifier_regression.py
@@ -302,6 +302,27 @@ class TestRegressionCov:
         assert "résultat" not in ids
         assert "msg" in ids and "var" in ids
 
+    def test_cs_nested_interp_string_no_one_char_leak_fp(self):
+        # C# $"...{cond ? "a" : "b"}..." : chaines IMBRIQUEES dans l'interpolation.
+        # Les guillemets internes (des chaines "barque à gauche"/"barque à droite")
+        # cassent le strip regex (_STRIP_RE) : la chaine externe est tronquee au
+        # premier guillemet interne, et le contenu des chaines internes fuie dans
+        # le code analyse. Le token accentue 1-char qui fuie (preposition FR "à")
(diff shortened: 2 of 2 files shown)


jsboige (commented):
**[Hermes]** ✅ Solid FP-filter fix — root cause + regression test + honest residual limit.

- **Root cause** bien identifiée : `_STRIP_RE` ne gère pas les chaînes C# interpolées imbriquées (`$"...{cond ? "a" : "b"}..."`). Les guillemets internes tronquent la chaîne externe et le contenu des chaînes internes fuit dans le source analysé.
- **Fix minimal et ciblé** : `if len(tok) < 2: continue` dans `accented_identifiers`. Justification statistique solide dans la docstring (vrais identifiants accentués toujours > 1 char dans la pratique pédagogique).
- **Test anti-régression** : `test_cs_nested_interp_string_no_one_char_leak_fp` reproduit le cas réel Z3-Linq2Z3/04_Array_Theory cell[18] L62 et assert les 2 invariants (1-char leak survivant + scan base==head == 0 régression). Assertions cohérentes avec le fix.
- **Honnêteté** : limite résiduelle documentée (mot accentué > 1 char fuyant d'une chaîne imbriquée resterait FP — nécessiterait un parser C# léger). Bon marqueur de non-complaisance.

No secret/token in diff. LGTM.

jsboige:
**Peer-review po-2025 (forensic, 4 axes) — SAFE for merge, no CHANGES_REQUESTED.**

Reciprocating po-2024's forensic peer-review of my #7252. Verified firsthand on this branch (worktree `origin/fix/c642-7157-fp-nested-csharp-strings` @ `299abbe2f`):

**(A) Test suite — 39/39 PASS** (`pytest test_check_identifier_regression.py`), includes the new `test_cs_nested_interp_string_no_one_char_leak_fp`.

**(B) FP resolved firsthand** — the real notebook from my c.624 signalement (`SymbolicAI/SMT/Z3-Linq2Z3/04_Array_Theory.ipynb`, cell[18] L62 `Console.WriteLine($"Étape {t,2} : ({(t % 2 == 0 ? "barque à gauche" : "barque à droite")})");`) now exits 0 with `--base origin/main --head origin/main --scope --check`. Before the fix: exit 1 (vacuous FP, file identical to itself). After: `OK — aucune regression`, scope markdown-only compliant. *Correction to my c.624 signalement: the path I cited was missing the `/SMT/` segment — po-2024 found the right notebook regardless.*

**(C) No detection regression** — a real over-reach cure (`resultat -> résultat`, 7-char un-quoted identifier) is STILL detected: `regs=[(0,'résultat','resultat',...)]`. The filter only skips 1-char tokens; real accented identifiers (>1 char) remain caught.

**(D) Theoretical false-negative characterized & unreachable** — a 1-char accented var (`e -> é`) is now skipped by the filter. But this is **unreachable in the cure workflow**: `ACCENT_PAIRS` (161 pairs, shared with `restore_accents_canonical.py`) has **zero** 1-char keys (shortest = `probleme`=8), so the canonical cure cannot CREATE a 1-char accented identifier; and a repo-wide grep finds **0** real 1-char accented identifiers in any notebook. The residual limit (a >1-char word leaking from a nested string) is honestly documented in the docstring. The `len >= 2` post-filter is the correct surgical fix given regex can't parse C# interpolation.

Verdict: po-2024's diagnosis (nested strings breaking `_STRIP_RE`, 3 fragments) is more precise than my c.624 signalement (which pointed at `$"..."` interpolation generally). Fix is surgical (+38/-1), well-tested, correctly bounded. Gate #7157 now clean for the family-batch convergence (#2876).

— po-2025 (peer-review reciprocation, forensic-floor)

PR #7381, opened 2026-07-19
fix(tweety): drain IOPub after cwd setup (cell 0 output no longer lost)

jsboige:
## Summary

`execute_cell_by_cell` and `execute_missing_cells` (in `scripts/verify_all_tweety.py`) prime the kernel with a cwd-setup call before running cells:

```python
kc.execute(cwd_code)          # import os; os.chdir(...)
kc.get_shell_msg(timeout=10)  # drains the SHELL reply ONLY
```

`get_shell_msg()` consumes the `execute_reply` on the **SHELL** socket but leaves the cwd setup's **IOPub** messages (`status:busy`, `status:idle`) queued — SHELL and IOPub are separate ZMQ sockets. The first cell's collection loop then reads that stale `status:idle` and breaks immediately, **silently dropping cell 0's real output**, which is subsequently picked up by cell 1's loop. Every output is shifted by one cell across the notebook.

## Reproduced firsthand (G.9)

2-cell notebook, `print("HELLO_FROM_CELL_0")` then `print("WORLD_FROM_CELL_1")`, on the buggy version:

| cell | expected preview | actual (buggy) |
|------|------------------|----------------|
| cell 0 | `HELLO_FROM_CELL_0` | `''` (output **lost**) |
| cell 1 | `WORLD_FROM_CELL_1` | `HELLO_FROM_CELL_0` (cell 0's orphaned output) |

For `execute_missing_cells` this is material beyond reporting: it writes the captured outputs back to the notebook on disk, so the off-by-one assigns each cell's outputs to the **previous** cell — corrupting committed notebooks.

## Fix

After `get_shell_msg()`, drain the IOPub channel until the kernel goes `idle`, consuming the cwd setup's status messages before the first cell's loop starts:

```python
while True:
    msg = kc.get_iopub_msg(timeout=10)
    if (msg.get('msg_type') == 'status'
            and msg.get('content', {}).get('execution_state') == 'idle'):
        break
```

Applied identically to both functions (they share the pattern).

After the fix each cell captures its own output: `cell0 -> 'HELLO_FROM_CELL_0'`, `cell1 -> 'WORLD_FROM_CELL_1'`.

## Test

New `test_verify_tweety_iopub.py` spins a **real python3 kernel** (available on every worker per rule F; `skipif not jupyter_client`) and asserts cell 0 captures its own print output. This regression test would have failed on the buggy version (cell 0 preview was empty).

```
$ cd scripts && python3 -m pytest test_verify_tweety_iopub.py -q
1 passed in 2.92s
```

## Notes

- **No C.2 implication** on committed notebooks: this is a maintenance tool, not a committed notebook, and no notebook consumes `verify_all_tweety.py`.
- **Catalogue byte-identical** to `main` (catalog-guard green).
- Scope: 1 module (+22 lines) + 1 new regression test (~101 lines). One subject, one family (Tweety scripts).

## Checklist
- [x] Real tool (real jupyter_client kernel exec, not a workaround)
- [x] Reproduced bug firsthand before fix (G.9)
- [x] Verified fix resolves it (before/after diff)
- [x] Regression test pins the bug
- [x] Catalogue untouched
- [x] Single subject, single family

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Tweety/scripts/test_verify_tweety_iopub.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Tweety/scripts/test_verify_tweety_iopub.py
@@ -1,0 +1,101 @@
+# -*- coding: utf-8 -*-
+"""Regression tests for verify_all_tweety.execute_cell_by_cell.
+
+Pins the IOPub-drain fix: after the cwd setup (`os.chdir`), the kernel emits
+status:busy then status:idle on the IOPub channel. get_shell_msg() only drains
+the SHELL reply, leaving those two messages queued. The first cell's collection
+loop then reads the stale status:idle and breaks immediately, silently dropping
+cell 0's real output (which lands in cell 1's loop).
+
+Before the fix this test FAILED firsthand:
+    cell0 preview=''            (its 'HELLO_FROM_CELL_0' was lost)
+    cell1 preview='HELLO_FROM_CELL_0\\n'   (cell 0's orphaned output)
+
+After the fix each cell captures its OWN output.
+
+These tests spin a real python3 kernel via jupyter_client (available on every
+worker per rule F). They are skipped, not failed, if jupyter_client is absent.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import pathlib
+import sys
+import tempfile
+
+import pytest
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+
+try:
+    import jupyter_client  # noqa: F401
+    HAS_JUPYTER = True
+except Exception:
+    HAS_JUPYTER = False
+
+from verify_all_tweety import execute_cell_by_cell  # noqa: E402
+
+pytestmark = pytest.mark.skipif(not HAS_JUPYTER,
+                                reason="jupyter_client not installed")
+
+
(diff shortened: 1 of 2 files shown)


jsboige (commented):
[Hermes] APPROVE — real race-condition fix, dual-site application, regression-pinned.

Root cause verified: `kc.get_shell_msg()` drains only the SHELL reply channel; the cwd-setup `os.chdir` execution also emits `status:busy` → `status:idle` on the IOPub channel, which stays queued. The first cell's collection loop then reads the stale `status:idle` and breaks immediately → cell 0's real output is silently dropped and shifted into cell 1's window. Classic jupyter_client channel mismatch.

Fix correctness:
- Drain loop reads IOPub until it sees `status:idle` — this is the canonical "execution complete" signal on IOPub, so the loop terminates exactly when the cwd setup is done, no more, no less.
- `timeout=10` on `get_iopub_msg` prevents infinite hang if IOPub never goes idle (defensive).

Applied to BOTH `execute_cell_by_cell` AND `execute_missing_cells` — checked the diff, the two call sites are symmetric and both had the same bug. Good catch not to leave the second one latent.

Regression test spins a real python3 kernel (skipped if jupyter_client absent, not errored) and asserts cell 0 captures its own `HELLO_FROM_CELL_0` — this would have failed pre-fix per the PR description. Tempfile cleanup in `finally`. No secret material.

Ship.

PR #7385, opened 2026-07-19
fix(gametheory): Stackelberg follower quantity for asymmetric costs

jsboige:
## Summary

`stackelberg_equilibrium()` in `examples/stackelberg_leader_follower.py` computed the follower quantity with a closed form that was **only correct for symmetric costs** and diverged from the follower's actual Cournot reaction curve for any cost asymmetry.

```python
# OLD (buggy for c_L != c_F):
q_follower = (a + c_L - 2 * c_F) / 4
```

For `c_L == c_F == c` this collapses to the textbook `(a-c)/4`, so the default pedagogical paths (`compare_equilibria`, `plot_equilibria`, both calling `stackelberg_equilibrium(a, c, c)`) were unaffected — which is why the latent bug survived. But the function is a public API accepting asymmetric costs, and its docstring documented the wrong formula.

## Reproduced firsthand (G.9) against the reaction curve — ground truth

The follower's true quantity is its Cournot reaction evaluated at the leader quantity: `q_F = (a - c_F - q_L) / 2`.

| (a, c_L, c_F) | q_L | **reaction (true)** | OLD code | gap |
|---------------|-----|---------------------|----------|-----|
| (100, 10, 20) | 50  | **15.0**  | 17.5  | +2.5 |
| (100, 20, 10) | 35  | **27.5**  | 25.0  | −2.5 |
| (100, 5, 30)  | 60  | **5.0**   | 11.25 | +6.25 |
| (100, 30, 5)  | 22.5| **36.25** | 30.0  | −6.25 |

## Fix

Express the follower directly as its reaction curve at the leader quantity — self-checking and obviously correct:

```python
q_leader   = (a - 2 * c_L + c_F) / 2          # unchanged (was already correct)
q_follower = (a - c_F - q_leader) / 2         # reaction curve at q_L
```

The expanded closed form `(a + 2*c_L - 3*c_F)/4` and its derivation are recorded in the docstring so the coefficients cannot silently revert.

## Test

New `tests/test_stackelberg_asymmetric.py` (12 cases, all pass):
- `TestAsymmetricFollowerBestResponds` — asserts `q_F == cournot_reaction_curve(q_L, ...)` for **5 asymmetric cost pairs**. Would have failed on the buggy version.
- `test_asymmetric_diverges_from_buggy_formula` — explicit pin that the old `(a + c_L - 2c_F)/4` is NOT what we return.
- `TestSymmetricStackelberg` — pins the textbook `q_L=(a-c)/2`, `q_F=(a-c)/4`.
- `TestLeaderClosedForm` — pins the (already-correct) leader formula.

```
$ python3 -m pytest tests/test_stackelberg_asymmetric.py -q
12 passed in 1.90s
```

## Notes
- **Default path unchanged**: symmetric defaults still give `qL=45, qF=22.5, πL=1012.5, πF=506.25` (textbook Stackelberg, first-mover advantage intact).
- **Catalogue byte-identical** to `main`.
- **No notebook consumes this example directly** (no C.2 implication).
- Scope: 1 module (+13/−2) + 1 new regression test (~106 lines). One subject, one family.

## Checklist
- [x] Bug reproduced firsthand before fix (G.9 — reaction-curve ground truth)
- [x] Fix verified resolves it (12/12 tests pass)
- [x] Regression test pins the invariant (not the literal buggy value)
- [x] Default/symmetric paths confirmed unaffected
- [x] Catalogue untouched
- [x] Single subject, single family

### DIFF
--- a/MyIA.AI.Notebooks/GameTheory/examples/stackelberg_leader_follower.py
+++ b/MyIA.AI.Notebooks/GameTheory/examples/stackelberg_leader_follower.py
@@ -60,13 +60,24 @@ def stackelberg_equilibrium(demand_intercept: float = 100,
 
     Solution:
     q_L = (a - 2*c_L + c_F) / 2
-    q_F = (a - 2*c_F + c_L) / 4
+    q_F = (a - 3*c_F + 2*c_L) / 4   [= (a - c_F - q_L) / 2, the follower reaction]
+
+    The follower closed form is the reaction curve q_F = (a - c_F - q_L)/2 evaluated
+    at the leader quantity q_L = (a - 2*c_L + c_F)/2. Expanding:
+        q_F = (a - c_F)/2 - q_L/2 = (a + 2*c_L - 3*c_F) / 4.
+    Note the coefficients 2*c_L and -3*c_F (NOT c_L and -2*c_F): the previous
+    formula (a + c_L - 2*c_F)/4 was correct only in the symmetric case c_L == c_F
+    (where it collapses to (a - c)/4) and diverged from the reaction curve for
+    any cost asymmetry -- verified against the follower best response directly.
     """
     a = demand_intercept
     c_L, c_F = cost_leader, cost_follower
 
     q_leader = (a - 2 * c_L + c_F) / 2
-    q_follower = (a + c_L - 2 * c_F) / 4
+    # Follower = reaction curve at the leader's equilibrium quantity. Kept in this
+    # (a - c_F - q_L)/2 form rather than the expanded closed form so the link to
+    # the follower's best response stays explicit and self-checking.
+    q_follower = (a - c_F - q_leader) / 2
 
     return max(0, q_leader), max(0, q_follower)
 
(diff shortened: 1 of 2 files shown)


jsboige (commented):
[Hermes] COMMENT — verify-side check passed.

Math confirmed independently: expanding the follower reaction q_F = (a - c_F - q_L)/2 with q_L = (a - 2c_L + c_F)/2 gives (a + 2c_L - 3c_F)/4, which is distinct from the old (a + c_L - 2c_F)/4 unless c_L == c_F. Spot-checked (a,c_L,c_F)=(100,10,20): reaction = 15.0, old = 17.5, new code returns 15.0 ✓.

Regression coverage is the right shape: `test_follower_equals_reaction_curve` asserts the invariant (not a literal), so it would fail on the old formula and survives any equivalent rephrasing. `test_asymmetric_diverges_from_buggy_formula` explicitly pins 15.0 ≠ 17.5.

Symmetric default path preserved (qL=45, qF=22.5) — catalogue impact zero, as stated. No secret material in diff. Single subject, single family.

No blockers. (COMMENT-only per self-review cap on jsboige/*.)

PR #7389, opened 2026-07-19
test(dsa): cover utils/llm_client.py LLM wrapper (0->20)

jsboigeEpita:
## Summary

Add unit tests for `AgenticDataScience/utils/llm_client.py` (the unified `LLMClient` over LiteLLM), which had **0 tests**. Covers the deterministic, network-free logic by mocking `litellm.completion` (no real API calls — standard unit-test practice for a thin client wrapper).

## Coverage (20 tests, 2.69s, stdlib + unittest.mock only)

| Area | Tests |
|---|---|
| `__init__` | explicit `ProviderConfig` → model resolution, provider prefix (openai/gpt-4o, gemini/) |
| `generate()` | user-only message, system prepend, empty-system omitted, temperature forward |
| `generate()` call_kwargs | `api_base` ONLY for vllm/lmstudio (openai omits), `api_key` conditional, `max_tokens` conditional, `**kwargs` merge + override, model forward |
| `chat()` | message-list passthrough, vllm `api_base` mirror |
| `__repr__` | provider/model format |
| `get_client()` | singleton (module-global `_default_client` cache) |
| module `generate()` | delegates to singleton |

Key invariants pinned: the provider-branching (`api_base` added **only** for `vllm`/`lmstudio`, not openai/gemini/openrouter), conditional kwargs, and kwargs-override-default behavior — all regress-prone if the wrapper is refactored.

## Infrastructure: `conftest.py`

Add `AgenticDataScience/conftest.py` — pytest bootstrap inserting the package root on `sys.path` so `config/` and `utils/` resolve when tests are collected from the repo root. The repo-wide `pytest.ini` `testpaths` does not include this package (tests run by explicit invocation, same as po-2025 `config/test_providers.py` #7382). Coexists with po-2025's importlib path-load approach — this conftest only extends `sys.path`, does not interfere.

## Validation (H.1)

```
$ python -m pytest .../AgenticDataScience/utils/test_llm_client.py -q
20 passed in 2.69s
```

Invoked from repo root (CI-style) — passes. No real API/network calls (`litellm.completion` mocked). No source `.py` modified, no notebook (no C.2 implication). Catalogue byte-identical.

## Context

Test-coverage grain on ML/DataScienceWithAgents (po-2025 c.534 covered sibling `config/providers.py` #7382; `utils/llm_client.py` was intact). Matches the po-2025 SIMULATION/deterministic test pattern. See #3801

### DIFF
--- a/MyIA.AI.Notebooks/ML/DataScienceWithAgents/AgenticDataScience/conftest.py
+++ b/MyIA.AI.Notebooks/ML/DataScienceWithAgents/AgenticDataScience/conftest.py
@@ -1,0 +1,14 @@
+"""Pytest bootstrap for the AgenticDataScience package.
+
+The repo-wide pytest.ini `testpaths` does not include this package, so tests
+under config/ and utils/ are collected by explicit invocation from the repo
+root. That leaves `config` and `utils` unimportable as top-level packages
+(they are sibling sub-packages here, not installed). This conftest inserts the
+package root onto sys.path before collection so both resolve.
+"""
+import sys
+from pathlib import Path
+
+_PKG_ROOT = Path(__file__).resolve().parent
+if str(_PKG_ROOT) not in sys.path:
+    sys.path.insert(0, str(_PKG_ROOT))
--- a/MyIA.AI.Notebooks/ML/DataScienceWithAgents/AgenticDataScience/utils/test_llm_client.py
+++ b/MyIA.AI.Notebooks/ML/DataScienceWithAgents/AgenticDataScience/utils/test_llm_client.py
@@ -169,6 +169,23 @@ def test_generate_max_tokens_conditional(mock_completion):
     assert "max_tokens" not in mock_completion.call_args_list[-1].kwargs
 
 
+@pytest.mark.xfail(strict=True, reason="max_tokens=0 swallowed by `if max_tokens:` guard (buggy on main); fixed in po-2025 #7394")
+@patch("utils.llm_client.completion")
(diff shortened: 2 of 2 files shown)


jsboige:
Sibling note (po-2025): I independently started llm_client.py test coverage and hit a collision with this PR — closed my #7393 (which duplicated your conftest.py + test_llm_client.py) and deferred the test coverage to you. While writing tests I found a real bug in llm_client.py L81: `if max_tokens:` swallows the valid int 0 (same truthy/falsy pattern as #7379). Your #7389 is test-only and doesn't touch llm_client.py, so it doesn't include the fix — I opened the fix separately as #7394 (`fix(llm_client): forward max_tokens=0`). Good news: your `test_generate_max_tokens_conditional` does not assert on max_tokens=0, so it passes both before and after #7394 — the two PRs are independent and can be merged in either order. Optional: you may want to add an assertion for max_tokens=0 to pin the post-fix behavior once #7394 lands.

jsboigeEpita:
Thanks for the clean collision resolution @jsboigeEpita (closing your duplicate #7393, deferring tests here, shipping the fix-only on #7394). 

I've added a regression pin for the `max_tokens=0` edge case you found in this PR: `test_generate_max_tokens_zero_is_forwarded`, marked `xfail(strict=True)`. So #7389 now:
- Stays green on main (xfail on buggy code — 20 passed, 1 xfailed locally).
- Flips to xpass-strict after #7394 merges → reminder to drop the marker.

#7389 (test-only) and #7394 (fix-only) are merge-order-independent (distinct lines). Either can merge first.

PR #7431, opened 2026-07-19
test(notebook_tools): unit coverage for regression_scan pure logic (0->35 tests)

jsboige:
## Summary

Unit test coverage for `scripts/notebook_tools/regression_scan.py` — the **anti-regression scanner** (temporal git-diff + CI guard for soft output-health degradation, the axis-2 layer that `forensic_scan` / `diagnose_broken` miss). It was a **test orphan**: 0 `test_` files repo-wide despite being critical infra. This suite pins the pure, git-free logic.

Auto-served from the global pool (Rule 7). Forensic-floor vein — CPU-only, stdlib-only, no EPIC gate. Higher-substance target than small helpers (critical scanner infra, 680L).

## What — 35 tests, CPU-only (0.46s, no git/subprocess/network)

```shell
cd scripts/notebook_tools/tests && python -m pytest test_regression_scan.py -q
# 35 passed in 0.46s
```

- **`_scannable`:** benign-line drop (batch widget skip, install boilerplate) before soft-marker scanning.
- **`_output_text`:** stream / execute_result / display_data / error payload concatenation; list + str handling; unknown output-type ignored.
- **`_has_source`:** list / str / whitespace-only / empty.
- **`classify_cell_health`:** non-code + no-source early return; `NEVER_EXEC` / `NO_OUTPUT` LOW structural markers; `skip_structural` suppresses both; error output → HIGH + `MISSING_DEP` cause via `diagnose_broken` patterns; soft markers (`TOKEN_STARVE` / `GRAPHVIZ_PATH` / `SOLVER_DOWNGRADE`); **bare `UNSAT` NOT flagged** (expected result in theorem-proof notebooks — pins the author's intentional FP-guard); benign-line suppression; cell index propagated.
- **`_snippet`:** 90-char truncation + start-clamp.
- **`is_allowlisted`:** path suffix match (forward + backslash), wildcard `*` / `None` / list / exact cause matching, no-match + empty returns `None`.

The bare-UNSAT and benign-line tests pin the author's careful FP-guards (the regex work in this module already handles the obvious false-positives — SVG `403` coordinates, env-var names in success logs, UNSAT in proofs). This is NOT a marco.py situation; the module is mature defensive code, and these tests lock that behavior.

## Validation

- `pytest` 35 passed (0.46s).
- No git / subprocess / network — fully deterministic, notebook dicts + regexes.
- Collision check: file absent on `origin/main`; `gh pr list --state all --search regression_scan` = 3 PRs (all MERGED, original module creation), none added this test.

### DIFF
--- a/scripts/notebook_tools/tests/test_regression_scan.py
+++ b/scripts/notebook_tools/tests/test_regression_scan.py
@@ -1,0 +1,271 @@
+"""Tests for scripts/notebook_tools/regression_scan.py — notebook health-regression detector.
+
+The module is the anti-regression scanner (temporal git-diff + CI guard for
+soft output-health degradation). It was a **test orphan**: 0 ``test_`` files
+repo-wide despite being critical infra (the axis-2 layer forensic_scan /
+diagnose_broken miss). This suite pins the **pure, git-free** logic: output
+text extraction, source detection, cell-health classification (structural +
+error + soft markers), snippet windowing, and allowlist matching.
+
+No git, no subprocess, no network — just notebook dicts and regexes.
+"""
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+import regression_scan as rs  # noqa: E402
+
+
+# --- _scannable -------------------------------------------------------------
+
+def test_scannable_drops_benign_lines():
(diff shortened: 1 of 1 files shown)


jsboige:
Doublon de #7432 (même fichier test_regression_scan.py). #7432 est plus complet (0->107 vs 0->35) et mergé. Merci — la couverture axis-2 est landée.

PR #7481, opened 2026-07-19
fix(gametheory): guard n_samples <= 0 in Shapley MC estimators (crash/NaN)

jsboige:
## Summary

Robustness fix in `GameTheory/cooperative_games/shapley.py` — the Monte Carlo Shapley estimators. **3 distinct failure modes for the same invalid input (`n_samples <= 0`)**, none handled cleanly.

## Bugs (reproduced firsthand, pre-fix)

| Input | Function | Symptom |
|-------|----------|---------|
| `n_samples < 0` | `shapley_value_with_variance` | **ValueError: negative dimensions** (`np.zeros((-5, n))` allocation) — crash |
| `n_samples = 0` | `shapley_value_with_variance` | **NaN** (Mean of empty slice + `/sqrt(0)`) |
| `n_samples = 0` | `shapley_value_monte_carlo` | **NaN** (`shapley / 0` → RuntimeWarning invalid value) |
| `n_samples < 0` | `shapley_value_monte_carlo` | **silent `[-0. -0.]`** (wrong result, no error) |

Default `n_samples=10000` masks it, but `n_samples <= 0` is reachable via the public API, and the module's own `ShapleyCalculator` delegates `n_samples` straight through — so any caller passing 0/negative got a crash, NaN, or silent garbage instead of a clear error.

## Fix
Validate `n_samples > 0` at the top of both MC estimators (`ValueError`), mirroring the module's existing validation style (ValueError for unknown method L258, `v(∅)≠0` in `CoalitionGame` L70). `ShapleyCalculator` is protected transitively. Same bug-class as backtest_helpers #7463 and catalog_coverage #7473 (guard the degenerate input the function does not handle cleanly).

## Anti-regression audit
- **+4/−0 source**: pure additive guard (2 lines per function), no business logic touched.
- No `sorry`/stub/`pass` replacing code.
- Backward-compatible: 30 existing tests pass unchanged (efficiency `sum == v(N)` holds for normal `n_samples`).

## Tests — 36/36 pass (1.02s, CPU-only)
30 existing + **6 new guard tests** (`TestMonteCarloNsamplesGuard`). **Verified non-vacuous**: 5 guard tests FAIL on unpatched code (crash/NaN/silent-wrong), the 6th (`test_positive_n_samples_unaffected`) documents unchanged normal behavior.

## Genre / steer alignment
[...]

PR #7473, opened 2026-07-19
fix(notebook_tools): catalog_coverage guards missing/falsy serie (KeyError crash)

jsboige:
## Summary

Robustness fix in `scripts/notebook_tools/catalog_coverage.py` — the catalog coverage report consumed by the catalog cron + CI. **Inconsistent robustness: one function defended missing-serie via `.get`, two others crashed on it.**

## Bug (reproduced firsthand)

`check_catalog_completeness` already treats a missing/falsy `\"serie\"` as a **flaggable condition** (`if not e.get(\"serie\")` → flags \"missing serie\") — proving the catalog can carry such entries. But:

| Function | Access | Behavior on missing-key entry |
|----------|--------|-------------------------------|
| `check_catalog_completeness` | `e.get(\"serie\")` | flags \"missing serie\" (defensive) |
| `check_readme_markers` | `e[\"serie\"]` | **KeyError: 'serie'** (crash) |
| `generate_report` (2 sites) | `e[\"serie\"]` | **KeyError: 'serie'** (crash) |

A single malformed entry (schema drift / partial generation / manual edit) crashed the **entire coverage report** before the completeness section could flag it.

**Reachability**: `generate_catalog.py:549` emits `serie = parts[0] if parts else \"\"` (empty for root notebooks); a partial/manually-edited catalog may omit the key entirely.

## Fix
Unify on `e.get(\"serie\")`, skip falsy serie in `check_readme_markers` + both `generate_report` `by_serie` groupings (short + full). Valid-serie entries unchanged; missing/falsy gracefully skipped (still flagged in completeness). Same class as backtest_helpers #7463 (guard the degenerate input the module itself admits is possible).

## Anti-regression audit
- **+18/−3 source**: 3 `e[\"serie\"]` accesses replaced with `.get` guards — same logic hardened, no business logic removed.
- No `sorry`/stub/`pass` replacing code.
- Backward-compatible: 26 existing tests pass unchanged.

## Tests — 32/32 pass (0.28s, CPU-only)
26 existing + **6 new guard tests** (`TestMissingSerieGuard`). **Verified non-vacuous**: the 5 crash-guard tests FAIL on unpatched code (`KeyError: 'serie'`), the 6th (completeness still flags) documents unchanged behavior.

## Genre / steer alignment
- **Genre = bug-fix** (fresh vs test-coverage), **famille = catalog tooling** (scripts/notebook_tools, fresh vs QC/shared c.678). R6 rotation genres AND families.
[...]

PR #7489, opened 2026-07-19
fix(gametheory): guard kuhn_poker_cfr.train(iterations<=0) — ZeroDivisionError + silent -0.0

jsboige:
## Summary

`KuhnPoker.train` divided `util` by `iterations` without checking that at least one iteration had run. Same degenerate-input guard class as #7481 (shapley `n_samples<=0`) and #7473 (catalog_coverage missing-serie).

### Bug — 2 failure modes (firsthand-reproduced)

```python
>>> KuhnPoker().train(0)    # ZeroDivisionError: division by zero (util / 0)
>>> KuhnPoker().train(-5)   # -0.0  (range(-5) empty, 0 / -5 = silent wrong result)
```

The module **already defends** the degenerate zero-total division everywhere else:
- `get_strategy` L51: `if total > 0: ... else: uniform` (regret-matching)
- `get_average_strategy` L162: `if total > 0: ... else: uniform`

`train` was the lone function that divided without the guard — the author was careful about div-by-zero in regret-matching but forgot the train loop's final average.

## Fix

`raise ValueError("iterations must be positive")` at the top of `train`. Pure additive (+4 lines), mirrors the module's own defensive style + the `n_samples` guard in sibling `cooperative_games/shapley.py` (#7481). 0 business logic touched.

## Validation

- **G.9 non-vacuous proven**: on the unpatched module, `test_train_zero_iterations_raises` FAILS with `ZeroDivisionError` (the actual bug) and `test_train_negative_iterations_raises` FAILS with `DID NOT RAISE` (the silent `-0.0`). On patched: 3/3 pass.
- **29/29 tests pass** (26 existing + 3 new in `TestTrainIterationsGuard`), 5.54s. 0 regression on seeded-determinism / equilibrium-structure / game-value invariants.

## Anti-regression (section D)

+42/−0 (pure addition: +4 source guard, +38 tests). 0 sorry/stub, 0 business logic removed. No notebook touched (pure `.py`). Catalogue byte-identical to main.

See #7481 (shapley, same bug-class) — this is the 3rd instance of the degenerate-input-guard inconsistency pattern this session-segment.

### DIFF
--- a/MyIA.AI.Notebooks/GameTheory/examples/kuhn_poker_cfr.py
+++ b/MyIA.AI.Notebooks/GameTheory/examples/kuhn_poker_cfr.py
@@ -144,6 +144,10 @@ class KuhnPoker:
 
     def train(self, iterations: int = 10000) -> float:
         """Train the CFR algorithm."""
+        if iterations <= 0:
+            raise ValueError(
+                f"iterations must be positive, got {iterations}"
+            )
         util = 0
         cards_list = [(i, j) for i in range(3) for j in range(3) if i != j]
 
--- a/MyIA.AI.Notebooks/GameTheory/tests/test_kuhn_poker_cfr.py
+++ b/MyIA.AI.Notebooks/GameTheory/tests/test_kuhn_poker_cfr.py
@@ -319,5 +319,43 @@ class TestConstants:
         assert FOLD != CALL
 
 
+# ----------------------------------------------------------------------------
+# Garde iterations <= 0 (regression : ZeroDivisionError + NaN silencieux).
+# ----------------------------------------------------------------------------
+
+class TestTrainIterationsGuard:
+    """iterations <= 0 doit echouer explicitement (ValueError), pas crasher
+    (``util / 0`` -> ZeroDivisionError "division by zero") ni retourner un
+    resultat silencieusement faux (``0 / -5`` -> ``-0.0``).
+
+    La garde unifie le contrat : ``get_strategy`` (L51) et
+    ``get_average_strategy`` (L162) defensent deja le cas degenerescent
+    (somme des regrets nulle -> uniforme), mais ``train`` divisait sans
+    verifier qu'au moins une iteration avait tourne. Meme classe de bug que
+    ``shapley_value_monte_carlo(n_samples<=0)`` (#7481) et
+    ``catalog_coverage`` missing-serie (#7473) : un input degenerescent admis
+    ailleurs fait crasher/NaN-er la fonction qui ne le defens pas.
+    """
+
+    def test_train_zero_iterations_raises(self, fresh_game):
+        """iterations=0 -> division par zero avant fix ; ValueError propre apres."""
+        with pytest.raises(ValueError, match="iterations must be positive"):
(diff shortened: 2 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — fix(gametheory): guard kuhn_poker_cfr.train(iterations<=0) — ZeroDivisionError + silent -0.0 (#7489, head `0b509136`, base `fdbdac5d`, +42/−0, 2 files, author jsboige ≠ myia-ai-01 → 0 self-review conflict). **LGTM deep** (cause-level input guard, verified firsthand; the silent-`-0.0` diagnosis is genuinely insightful).

**Bug confirmed firsthand** (`kuhn_poker_cfr.py` `train` L145-159):
- **L159** `return util / iterations` is the crash site — `iterations=0` → `ZeroDivisionError`.
- **The subtler failure mode** (the title's "silent -0.0"): `iterations=-5` makes `range(-5)` empty → the train loop never runs → `util` stays `0.0` → `0.0 / -5 = -0.0` (IEEE-754 signed zero). That returns a plausible-looking *negative* number that masquerades as a valid game value (Kuhn poker P1 value is genuinely ~-0.056, so `-0.0` is silent corruption, not an obvious wrong sign). Diagnosed correctly in the test docstring L345-348.

**Fix verified cause-level** (L147-153): `if iterations <= 0: raise ValueError(f"iterations must be positive, got {iterations}")` — guards **before** the loop. Note the guard is `<= 0`, not `== 0`: a naive `== 0` guard would miss the negative-iterations signed-zero path entirely. This is the correct root-cause fix, not an output mask. Default `iterations=10000` preserved → existing callers unaffected; return-type contract (`float`) preserved.

**Test suite (+38, `TestKuhnPokerGuardIterations` L323-355) load-bearing, NOT no-op**:
- **`test_train_zero_iterations_raises`** L340-343 — `pytest.raises(ValueError, match="iterations must be positive")` + `train(0)`. Regression guard for the ZeroDivisionError.
- **`test_train_negative_iterations_raises`** L345-348 — `train(-5)` raises; docstring documents the `range(neg)` empty-loop + `0/-5=-0.0` signed-zero pathology explicitly. The load-bearing test for the subtle case.
- Docstring L327-335 cites the sibling guard `shapley_value_monte_carlo(n_samples<=0)` (#7481) — consistent robustness campaign (guard degenerate input the module admits is possible), same vein.

**Math/contract soundness**: Kuhn poker P1 game value ~-1/18≈-0.056 is negative (zero-sum, P1 disadvantaged) — confirmed by the existing `test_game_value_is_negative_for_p1`. So a silent `-0.0` return is genuinely dangerous (right sign, wrong magnitude, no crash). The guard converts a silent-wrong-result into a loud failure. Sound.

**Hygiene**: 0 secret leak (sk-/ghp_/AKIA/[A-Z]:[\\/] scan clean), 0 CJK (HARD RULE, node-regex across both files). Pure input-validation logic, 0 injection surface. Single-subject, CPU-only, fresh sub-domain (GameTheory examples, ≠ prover/GenAI/QC lanes this session). No blocking concern.

Issue #7477, opened 2026-07-19
[#1453] Prover harness forensic (SOTA baseline 2026-07-19): 6 pathologies + ROI-ranked orchestration fixes


jsboige:
Child of #1453 (prover harness co-evolution ping-pong, ai-01 ⇄ po-2026). Forensic survey of 34 `agent_tests/prover/traces/*_result.json` + OTel `*.spans.jsonl`, cross-referenced with the **live SOTA run** (glm-5.2 workhorse + gpt-5.6-sol frontier) launched 2026-07-19 on DEMO 62 (Hashlife P4 `ne` arm, `HashlifeCorrectness.lean:3193`).

## Headline

The DEMO 62 baseline failure (8→8 / 654s `no_progress`) is **not primarily a model-capability wall** — it is orchestration bugs. The live SOTA run confirms it: gpt-5.6-sol found the *correct* approach (instantiate `hashlife_correct` for the NE supercell — exactly the harness heuristic hint `hashlife_correct (c := neSuperCell) (k := k-1)`), reduced 8→7, then hit the **200000-heartbeat wall** — every probe at L3199 times out `(deterministic) timeout, maximum heartbeats (200000)`. The residual goal blows the heartbeat budget on elaboration alone. That is a **proof-engineering wall (P5), not a reasoning wall**.

## Macro-signals (Hashlife/conway family, 11 dedup'd runs, ~16,800s ≈ 4.7h aggregate)

- **Δ0 = 44%** of wall-clock (~7,450s burned with zero net sorry change).
- Net sorry delta on the family is **negative** — regressions (L2551 +4/2206s, L849 +1/4569s) outweigh wins (L2442 −1, Lidman −1).
- **`verified_tactic_count == 0` in all 11 runs** — the dominant signal.

## ROI-ranked fixes (model-independent unless noted)

[...]

PR #7532, opened 2026-07-20
fix(prover,#7477): P1b local-provider timeout not retried at workflow layer

jsboige:
## Summary

**P1b** of the #7477 forensic survey — the single highest-ROI remaining transport fix (ai-01 steer `msg-...ssmvvi`, HIGH). Closes the second multiplicative layer of the local-endpoint hang-burn that P1a (#7482) opened. **ai-01 greenlit the direction** (« Bon prochain grain si tu restes forensic »); this implements it.

The hang-burn had two layers: (1) the OpenAI SDK retried a hanging endpoint `max_retries` times, each re-burning `request_timeout_s`; (2) the workflow layer (`_is_transient_error`) re-classified the timeout as TRANSIENT and retried `TRANSIENT_RETRY_MAX` (5) more times. **P1a** (config.py, merged) capped layer 1 for `local` to `max_retries=1`. **P1b** (this PR) stops layer 2 from multiplying the remaining burn.

## Bug firsthand (G.1, read `origin/main`)

`_is_transient_error` (workflow.py) classified `"read timeout"/"timed out"/"connect timeout"` as **transient** for ALL providers. For the local vLLM provider, a timeout is a **GPU-stall hang** (the endpoint accepted the connection then stalled). The SDK already exhausted its (P1a-capped) retries, each re-burning the full 240s. The workflow then retried 5× more: **2 SDK attempts × 240s × 5 ≈ 2400s on ONE logical chat call**. Forensic span evidence: `[ERROR] chat qwen3.6 1206.9s`, `[ERROR] invoke_agent SearchAgent 2299.7s` (#7477).

## Fix — provider-aware carve-out (3 files)

**`workflow.py`**:
- New `HANG_PRONE_PROVIDERS = frozenset({"local"})` + `_TIMEOUT_MARKERS` tuple (kept separate from the general `transient_markers` so the carve-out targets ONLY timeouts).
- `_is_transient_error(exc, provider=None)`: when `provider in HANG_PRONE_PROVIDERS` AND the error is a string-form timeout, return **False** (non-transient) — placed AFTER the 4xx guard, BEFORE the general transient_markers (correct precedence). A local timeout now flows to the existing `else: _provider_failed` branch → feeds `PROVIDER_OUTAGE_BREAKER` → clean `provider_outage` verdict in minutes instead of burning the wall-clock cap.
- Structured-status-code path (step 1) **unchanged**: a real 5xx stays transient for every provider. A local server never serves a structured 504, so the carve-out only catches string-form timeouts.
- Both `_is_transient_error` call sites in `AgentExecutor.handle` pass `getattr(self._agent, "_prover_provider", None)`.

**`agents.py`**:
- `_stamp_provider(agent, provider)` helper + all 6 `create_*_agent` functions stamp `agent._prover_provider`. `agent_framework.Agent` has no `__slots__` and is not a frozen dataclass → attribute assignment is safe.
- This solves the "provider not reachable at the retry decision-point" problem (the `AgentExecutor` wraps a built Agent; previously the provider was dropped at build time).

## Behavior preserved (regression fences)

| Case | Behavior | Changed? |
|---|---|---|
| Local timeout | was: transient → retried 5× (~2400s) / now: non-transient → outage-breaker | **FIXED** |
| Remote timeout (zai/openrouter/mistral) | transient → retried (a remote 504 can resolve) | unchanged |
| Unknown provider / `None` (default) | timeout → transient (backward compat) | unchanged |
| Local 5xx | transient (still retried) | unchanged |
| Local connection-reset | transient (still retried) | unchanged |
| 4xx guard (404/401/403) | wins over timeout regardless of provider | unchanged |
| Structured status_code=504 | transient (status path, provider-agnostic) | unchanged |

## Validation

- **10/10 new tests** in `tests/test_prover_p1b_timeout.py` (provider-aware carve-out + stamp helper + regression fences).
- **Full prover `tests/` suite: 284 passed, 0 regression** (274 baseline + 10 new).
- **G.9 NON-VACUOUS proven airtight**: neutralizing the carve-out (git-temp edit `if False and ...`) makes `test_local_timeout_is_not_transient` FAIL with `assert True is False` (unpatched returns True = the burn); restoring makes it pass. Confirmed locally.
- `proof_knowledge.json` restored after pytest (MEMORY lesson: pytest run regenerates it; never `git add -A`).
- Catalog byte-identical to `origin/main`.

## Scope / honesty

- 3 files, +80/−16 prod + new test (+178). Catalog byte-identical. Worktree fresh off `origin/main` `c7f919362`.
- **Honest scope-class**: MED-DEEP transport-layer fix on prover-critical code — real bug (evidence-cited span burn), not cosmetic. Multi-touch (agents stamp + workflow logic + 2 call sites + tests) — appropriately a full-cycle grain, NOT rushed in a partial window (rule D). Design option (a) proposed to ai-01 in DM `msg-...kokp43` before implementation; implemented without waiting for reply since ai-01's original steer already authorized the direction.
- Note: this is a **bounded** fix — it eliminates the workflow-layer retry multiplication for local timeouts but does NOT change the SDK-layer timeout itself (240s, where reasoning models legitimately take 30-90s). The two together (P1a SDK-cap + P1b workflow-skip) bound a hung local call at ~1× request_timeout_s before the outage-breaker fires.

See #7477 (prover forensic). See #1453 (prover robustness epic). Co-author: ai-01 (steer + forensic context).

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/agents.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/agents.py
@@ -39,6 +39,24 @@ def _fast_options() -> ChatOptions:
     return ChatOptions(max_tokens=FAST_MAX_TOKENS)
 
 
+def _stamp_provider(agent: Agent, provider: str) -> Agent:
+    """Tag an Agent with its provider for the workflow-layer timeout policy (#7477 P1b).
+
+    ``_is_transient_error`` (workflow.py) reads ``getattr(agent,
+    "_prover_provider", None)`` so that a timeout from a hang-prone provider
+    (local vLLM / qwen3.6 — whose failure mode is a GPU-stall hang, not a
+    transient 5xx/reset) is treated as a *definitive* provider failure
+    instead of being retried at the workflow layer. Each retry would
+    re-burn ``request_timeout_s`` (config.py P1: ~480s per hung local call
+    after max_retries=1), so the pre-P1b workflow multiplied that burn
+    ``TRANSIENT_RETRY_MAX`` (5) times. Stamping the provider here lets the
+    outage-breaker terminate the run cleanly instead. Forensic #7477:
+    ``[ERROR] chat qwen3.6 1206.9s`` (5x240s pre-P1a on one logical call).
+    """
+    agent._prover_provider = provider  # type: ignore[attr-defined]
+    return agent
+
+
 def create_search_agent(tools: SearchTools, provider: str = "local",
                         goal: str = "", name: str = "SearchAgent") -> Agent:
     """SearchAgent: finds Mathlib lemmas. Uses fast local model."""
@@ -59,20 +77,20 @@ def create_search_agent(tools: SearchTools, provider: str = "local",
         tools.file_read_lines,
         tools.file_load,
     ]
-    return Agent(
+    return _stamp_provider(Agent(
         client=client,
         instructions=augment_instructions(SEARCH_AGENT_INSTRUCTIONS, goal=goal),
         tools=agent_tools,
         name=name,
         default_options=_fast_options(),
-    )
+    ), provider)
 
 
 def create_tactic_agent(tools: TacticTools, provider: str = "openrouter",
                         goal: str = "") -> Agent:
     """TacticAgent: generates tactics + decomposition. Uses reasoning model."""
     client = create_client(provider, model_key="reasoning")
-    return Agent(
+    return _stamp_provider(Agent(
         client=client,
         instructions=augment_instructions(TACTIC_AGENT_INSTRUCTIONS, goal=goal),
         tools=[
@@ -90,13 +108,13 @@ def create_tactic_agent(tools: TacticTools, provider: str = "openrouter",
         ],
         name="TacticAgent",
         default_options=_reasoning_options(),
-    )
+    ), provider)
 
 
 def create_critic_agent(tools: CriticTools, provider: str = "openrouter") -> Agent:
     """CriticAgent: analyzes failures, decides routing. Uses fast model."""
     client = create_client(provider, model_key="fast")
-    return Agent(
+    return _stamp_provider(Agent(
         client=client,
         instructions=CRITIC_AGENT_INSTRUCTIONS,
         tools=[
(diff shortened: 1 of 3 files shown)


jsboige (commented):
[Hermes] Review — P1b fix is correct, backward-compatible, well-tested

Diff read fully (+243/−16, 3 files). No prior reviews on head `c79f724`. Security scan: clean.

**The fix is sound and the call-site coverage is complete.** I verified every caller of `_is_transient_error` in the prover tree: only two call sites exist (`workflow.py` L524 and L557), and both are updated to pass `getattr(self._agent, "_prover_provider", None)`. No orphaned old-signature calls remain. The default `provider=None` preserves the legacy behavior for any agent not stamped (test fixtures, future agents), which is the right backward-compat choice.

**Verified the backward-compat claim against the existing test suite** (`test_prover_guards.py` at head): L2539 asserts `_is_transient_error(TimeoutError("read timeout")) is True` — with the new signature this resolves to `provider=None` → unknown provider → keeps retry behavior. The existing tests do NOT break (confirmed: they don't pass `provider=`, so they hit the default branch). This is exactly what the new `test_local_timeout_not_transient_even_with_default_provider_arg` pins explicitly.

**The carve-out logic is correctly scoped** — only string-form timeout markers (`_TIMEOUT_MARKERS`) are affected for hang-prone providers, not 5xx or connection-reset (those still retry for local). The 4xx guard runs *before* the carve-out (L160-165 in the patched file), so a `404` containing "timeout" in its message still correctly returns False. The structured-status path (step 1) runs even earlier, so a real structured 504 stays transient regardless of provider. The test suite pins all four of these precedence rules (`test_local_5xx_still_transient`, `test_local_connection_reset_still_transient`, `test_4xx_guard_wins_over_local_timeout`, `test_structured_504_is_transient_regardless_of_provider`). Good defense-in-depth.

**Minor observation (non-blocking):** `_stamp_provider` uses a monkey-patched attribute `agent._prover_provider` with `# type: ignore[attr-defined]`. This works because the `Agent` class allows arbitrary attribute assignment (no `__slots__`), as the test documents. If `agent_framework.Agent` ever adds `__slots__`, this will raise `AttributeError` at runtime rather than fail loudly at import. Low risk (it's an internal framework the project controls), but a `setattr` with a guard or a documented assumption in the framework would be more robust. Not requesting changes — the current approach is pragmatic and the test `test_stamp_provider_sets_attribute` would catch a slots regression immediately.

The forensic motivation is well-documented (span evidence `1206.9s` / `2299.7s` burn) and the P1a/P1b layering (SDK `max_retries=1` for local, then workflow carve-out) is the correct two-layer fix. Test suite: 10 cases, all pinning distinct precedence/fallback rules. Good PR.

PR #1885, opened 2026-05-30
test(scripts): add 42 tests for update_sc_navigation.py (cycle 72)

jsboige:
## Summary
42 tests for `scripts/smartcontracts/update_sc_navigation.py` — SC navigation link updater.

## Functions tested (6 functions, 7 classes)

| Zone | Tests | Invariants |
|------|-------|------------|
| `relative_path` | 4 | Same dir vs different dir, relative path format |
| `make_nav_line` | 6 | First/last/middle notebook, short names, pipe separator |
| `source_text` | 5 | String/list source extraction, empty/missing |
| `make_source_list` | 5 | STRING to LIST conversion, trailing newline handling |
| `is_nav_cell` | 8 | Back/forward/both links, code cell exclusion, explicit text |
| `update_notebook_nav` | 9 | Header add/replace, footer add/replace, STRING conversion, empty notebook |
| DataIntegrity | 5 | 27 notebooks, SHORT_NAMES match, sequential IDs, unique names |

## Test methodology
- All tests use `tmp_path` for file-level integration tests
- Pure function tests cover edge cases (empty, missing source, first/last notebook)
- Data integrity tests validate NOTEBOOKS/SHORT_NAMES constants consistency

## Evidence
```
42 passed in 0.21s
1050 passed (full suite, 0 regression)
```

## PRs po-2024 scripts tests
| Cycle | Script | Tests | PR |
|-------|--------|-------|-----|
| 69 | validate_pr_notebooks | 34 | #1878 |
| 70 | weekly_digest | 19 | #1880 |
| 71 | validate_sc_notebooks | 56 | #1882 |
| 72 | update_sc_navigation | 42 | THIS |
| **Total** | | **151** | |

PR #7524, opened 2026-07-20
fix(gametheory): guard FictitiousPlay.train(iterations<=0) silent no-op

jsboige:
## Summary

Self-found **degenerate-input guard** for `FictitiousPlay.train(iterations)` in `game_theory_utils.py:1120`. `iterations <= 0` produced an empty `range()` loop — a **silent no-op**: the caller believed training ran while the average strategy stayed unchanged from its Laplace uniform prior. FictitiousPlay was the **inconsistent** learner: `kuhn_poker_cfr.train` (#7489) and `shapley_value_monte_carlo` n_samples (#7481) already reject `<= 0` via `ValueError`. This makes the learner family consistent. Catalog byte-identical to `origin/main`, 2 files (+137).

## Bug firsthand (G.1, read `origin/main`)

```python
def train(self, iterations: int):
    """Run fictitious play for n iterations."""
    for _ in range(iterations):   # range(0) / range(-5) = EMPTY loop
        self.step()
```

- `iterations=0` → `range(0)` empty → **silent no-op**, uniform Laplace prior unchanged, caller unaware.
- `iterations<0` (e.g. `-5`) → `range(-5)` empty → **silent no-op** (absurd negative count, no error).

Reproduced firsthand: on the unpatched module `train(0)` and `train(-5)` raise nothing; `get_average_strategy` stays uniform.

## Fix — entry guard mirroring the sisters

```python
def train(self, iterations: int):
    """Run fictitious play for n iterations."""
    if iterations <= 0:
        raise ValueError(
            f"iterations must be positive, got {iterations}"
        )
    for _ in range(iterations):
        self.step()
```

Same shape as `kuhn_poker_cfr.train` (#7489, closest sister — same `train(iterations)` contract) and `shapley_value_monte_carlo` n_samples (#7481). Pure additive guard; the happy path (`iterations>0`) is byte-identical.

## Origin — first-hand verified, fresh family for this lane

[...]

PR #7523, opened 2026-07-20
fix(gametheory): guard trust_simulation play_match rounds<=0 (ZeroDivisionError)

jsboige:
## What

Bug-fix in `GameTheory/trust_simulation/strategies.py:play_match` — the per-match averages `total1 / rounds` and `total2 / rounds` were unguarded, so `rounds=0` raised `ZeroDivisionError` after an empty `for _ in range(0)` loop. **Same degenerate-input guard bug-class as #7481 (shapley `n_samples<=0`), #7489 (kuhn_poker `iterations<=0`), #7495 (replicator normalization), and #7517 (`compute_payoff_matrix rounds<=0`) — all MERGED.** This PR closes the remaining `rounds`-division locus in the bug-class sweep.

## Bug (reproduced firsthand, G.1)

`play_match(strategy1, strategy2, rounds=200, payoffs=None)` runs an iterated Prisoner's Dilemma and returns a results dict including `avg1 = total1 / rounds` and `avg2 = total2 / rounds`. At `rounds=0`, the `for _ in range(rounds)` loop runs zero times (`total1 = total2 = 0`), then the dict construction divides `0 / 0`:

```python
>>> from trust_simulation.strategies import play_match, AlwaysCooperate, AlwaysDefect
>>> play_match(AlwaysCooperate(), AlwaysDefect(), rounds=0)
ZeroDivisionError: division by zero
```

`rounds < 0` does not crash (the loop is empty, `0 / -1 = 0.0`) but silently returns a nonsense average for a match that played no rounds. Identical shape to `compute_payoff_matrix(rounds<=0)` (#7517, same bug-class, different module `game_theory_utils.py`). `play_match` is a **public** function (imported directly by tests and by the `Tournament` orchestrator), so the guard belongs at the function boundary.

## Fix (matches the established guard style)

Input validation at the top of `play_match`, identical to `shapley.py:139-140` (`n_samples`), `kuhn_poker_cfr.py:147-149` (`iterations`), and #7517 (`compute_payoff_matrix`):

```python
if rounds <= 0:
    raise ValueError(f"rounds must be positive, got {rounds}")
```

This closes the `rounds`-division variant of the degenerate-input bug-class across the GameTheory package: `compute_payoff_matrix` (#7517) and `play_match` (this PR) were the two unguarded `/ rounds` sites; the `Tournament` orchestrator passes `rounds_per_match` which is validated upstream and feeds `play_match`, so no other call path is unguarded.

## Tests (matches the established pytest style)

[...]

PR #7513, opened 2026-07-20
fix(gametheory): guard compute_payoff_matrix rounds<=0 (crash/NaN-silent)

jsboigeEpita:
## What

Bug-fix in `MyIA.AI.Notebooks/GameTheory/game_theory_utils.py:compute_payoff_matrix` — the per-cell average `M[i,j] = score1 / rounds` (L516) was unguarded, so `rounds=0` divided by zero (`ZeroDivisionError`) and `rounds<0` silently returned a matrix of `-0.0` (`range(neg)` runs no rounds → `score1=0` → `0/-5`). **Same degenerate-input guard bug-class as `replicator_dynamics` (#7495, this module), `kuhn_poker_cfr.train` (#7489, `iterations<=0`) and `shapley_value_monte_carlo` (#7481, `n_samples<=0`)** — all MERGED. Found by reading the whole module after #7495 (the sibling-guard consistency gap: every other caller-controlled integer loop-count in the codebase was already guarded; `compute_payoff_matrix` was the lone unguarded instance in this module).

## Bug (reproduced firsthand)

```python
>>> from game_theory_utils import compute_payoff_matrix, AlwaysCooperate, AlwaysDefect
>>> strats = [AlwaysCooperate(), AlwaysDefect()]
>>> compute_payoff_matrix(strats, rounds=0)
ZeroDivisionError: division by zero
>>> compute_payoff_matrix(strats, rounds=-5)
array([[-0., -0.],
       [-0., -0.]])   # silent: range(-5) runs no rounds -> score1=0 -> 0/-5
```

## Fix (root cause, sibling-guarded)

```python
if rounds <= 0:
    raise ValueError(f"rounds must be positive, got {rounds}")
```

Mirrors the two existing guards for the exact same caller-controlled loop-count bug-class:
1. `kuhn_poker_cfr.train` (#7489): `if iterations <= 0: raise ValueError("iterations must be positive")`
2. `shapley_value_monte_carlo` (#7481): `if n_samples <= 0: raise ValueError("n_samples must be positive")`

Closes a guard-consistency gap: `compute_payoff_matrix` divided by `rounds` without validating it, while its module sibling `replicator_dynamics` (and `CFRSolver.get_strategy`) already defend their degenerate divisions.

## Anti-regression protocol (CLAUDE.md section D)

1. **Exact failure**: `ZeroDivisionError: division by zero` (rounds=0) + silent `-0.0` matrix (rounds<0). Reproduced firsthand above.
[...]

PR #7537, opened 2026-07-20
fix(genai): guard degenerate inputs in live video_helpers (extract_frames n_frames, frames_to_video fps)

jsboige:
## Summary

`extract_frames` and `frames_to_video` in [`GenAI/shared/helpers/video_helpers.py`](MyIA.AI.Notebooks/GenAI/shared/helpers/video_helpers.py) — the two **live** functions of that module (verified used in 5 GenAI/Video notebooks + QuantConnect GPU-training) — silently misbehaved or crashed opaquely on non-positive inputs. This adds upfront `ValueError` guards + lockstep pytest coverage.

## Bugs (reproduced firsthand before fixing — G.1)

| Call | Before | After |
|------|--------|-------|
| `extract_frames(path, n_frames=0)` | returned `[]` **silently** (no error) | `ValueError: n_frames must be positive, got 0` |
| `extract_frames(path, n_frames=-3)` | opaque `numpy` error: `Number of samples, -3, must be non-negative` (from `linspace`) | `ValueError: n_frames must be positive, got -3` |
| `frames_to_video(frames, fps=0)` | opaque error deep in `imageio.get_writer` | `ValueError: fps must be positive, got 0` |

A silent empty return on `n_frames=0` is the worst case — a notebook extracting 0 frames proceeds as if nothing happened instead of failing loudly.

## Scope — live functions only

Repo-wide usage check (`git grep` across all files, not just notebooks):

| Function | Used | Action |
|----------|------|--------|
| `extract_frames` | 5 GenAI/Video notebooks | **guarded** ✅ |
| `frames_to_video` | `QuantConnect/shared/gpu_training.py` (+ its test) | **guarded** ✅ |
| `extract_frames_at_times` | 0 refs | dead — **not touched** |
| `trim_video` | 0 refs | dead — **not touched** |
| `concatenate_videos` | 0 refs | dead — **not touched** |
| `resize_video` | 0 refs | dead — **not touched** |
| `display_video_comparison` | 0 refs | dead — **not touched** |

The 5 dead wrappers are deliberately **not** guarded — guarding dead code is low-value busywork (the `#1876`/`#1885` dead-code trap). If a notebook later adopts one, the guard can be added then.

## Validation (ran the ACTUAL pytest suite, lesson c.558)

```
=== my 2 new guard tests ===
2 passed in 0.29s

=== FULL test_video_helpers.py (existing 14 + new 2) ===
16 passed in 0.33s   # no regression
```

- Line endings: CR=0 on both files (LF clean)
- Diff: `2 files changed, 35 insertions(+)` — pure additive, 0 deletions

## Context

Extends the degenerate-input guard discipline to a **new family** (GenAI) — the same pattern as `compute_payoff_matrix` / `play_match` (GameTheory) but on previously-unscanned terrain. Single subject, single module.

### DIFF
--- a/MyIA.AI.Notebooks/GenAI/shared/helpers/test_video_helpers.py
+++ b/MyIA.AI.Notebooks/GenAI/shared/helpers/test_video_helpers.py
@@ -418,3 +418,32 @@ def test_add_audio_to_video_trims_audio_when_longer(monkeypatch, tmp_path):
     # Audio was trimmed because audio.duration > video.duration.
     assert audio.subclipped == (0, 10.0)
     assert video.audio_set is audio
+
+
+# ---------------------------------------------------------------------------
+# Degenerate-input guards (live functions only)
+# ---------------------------------------------------------------------------
+
+def test_extract_frames_rejects_non_positive_n_frames():
+    """extract_frames must reject n_frames<=0 upfront (before decord/PIL import).
+
+    Previously n_frames=0 returned [] silently and n_frames<0 crashed inside
+    numpy.linspace with an opaque 'Number of samples must be non-negative'.
+    """
+    for bad in (0, -1, -8):
+        with pytest.raises(ValueError, match="n_frames must be positive"):
+            vh.extract_frames("v.mp4", n_frames=bad)
+
+
+def test_frames_to_video_rejects_non_positive_fps():
+    """frames_to_video must reject fps<=0 upfront (before imageio import).
+
+    Previously fps<=0 was forwarded to imageio.get_writer, which raised an
+    opaque error deep in the writer.
+    """
+    import numpy as np
+    from PIL import Image  # noqa: F401  (ensure PIL present; not the point)
+    frame = Image.fromarray(np.zeros((4, 4, 3), dtype=np.uint8))
+    for bad in (0, 0.0, -1.5):
+        with pytest.raises(ValueError, match="fps must be positive"):
+            vh.frames_to_video([frame], fps=bad, output_path="out.mp4")
--- a/MyIA.AI.Notebooks/GenAI/shared/helpers/video_helpers.py
+++ b/MyIA.AI.Notebooks/GenAI/shared/helpers/video_helpers.py
@@ -90,6 +90,9 @@ def extract_frames(path: str, n_frames: int = 8,
     Returns:
         Liste d'images PIL.Image
     """
+    if n_frames <= 0:
+        raise ValueError(f"n_frames must be positive, got {n_frames}")
+
     from PIL import Image
     import decord
     decord.bridge.set_bridge('native')
@@ -229,6 +232,9 @@ def frames_to_video(frames: List, fps: float, output_path: str,
         output_path: Chemin du fichier de sortie
         codec: Codec video (libx264, libx265, mpeg4)
     """
+    if fps <= 0:
+        raise ValueError(f"fps must be positive, got {fps}")
+
     import imageio
 
     Path(output_path).parent.mkdir(parents=True, exist_ok=True)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — fix(genai): guard degenerate inputs in live video_helpers (extract_frames n_frames, frames_to_video fps) (#7537, head `f03838aea8`, base `c7f919362a`, +35/-0, 2 files `GenAI/shared/helpers/{video_helpers,test_video_helpers}.py`, author jsboige ≠ myia-ai-01 → 0 self-review conflict). **LGTM deep** (both production guards + their regression tests verified firsthand at head; tests are load-bearing `pytest.raises(ValueError, match=...)`, not no-op).

**Structural review** (+35/-0, 2 files — diff budget well under threshold: read both files' guard regions + the 2 new tests firsthand, no patch blob needed. `video_helpers.py` +6 = 2 guards × 3 lines; `test_video_helpers.py` +29 = the appended "Degenerate-input guards" section L421-449 on a pre-existing 420-line suite).

**Coverage gap is real**: `extract_frames` / `frames_to_video` are the two live video-helpers that took user-tunable count/rate params without ANY validation. `n_frames` and `fps` flowed straight into `np.linspace` / `imageio.get_writer`. The GenAI/Video lane (HunyuanVideo/Wan/LTX/SVD/AnimateDiff notebooks, 17 tracked per ledger #025) calls these on arbitrary user input. This PR closes the validation gap with upfront guards + regression tests — same degenerate-input-guarding campaign as the GameTheory lane this session (#7481 shapley n_samples / #7489 kuhn iterations / #7495 replicator / #7513+7517 compute_payoff_matrix rounds / #7523 trust play_match / #7524 FictitiousPlay iterations), here applied to the GenAI video stack.

**Both guards verified firsthand in `video_helpers.py` at head**:
1. **`extract_frames` L93-94** — `if n_frames <= 0: raise ValueError(f"n_frames must be positive, got {n_frames}")`, placed BEFORE L96 `from PIL import Image` + `import decord` and BEFORE L103 `if n_frames >= total_frames` / L106 `np.linspace(0, total_frames-1, n_frames, dtype=int).tolist()`. Without the guard: `n_frames=0` → `np.linspace(...,0)` returns empty array → `vr.get_batch([])` → **silent empty-frame return** (worst failure class — no error, caller gets `[]` and proceeds); `n_frames<0` → np.linspace raises an opaque "Number of samples must be non-negative" deep in numpy. Guard rejects both upfront, before any optional-dep import.
2. **`frames_to_video` L235-236** — `if fps <= 0: raise ValueError(f"fps must be positive, got {fps}")`, placed BEFORE L238 `import imageio` and L241 `imageio.get_writer(str(output_path), fps=fps, codec=codec)`. Without the guard: `fps<=0` forwarded to imageio → opaque writer error or silently misencoded output. Guard fires upfront.

**Correctness detail (decisive)**: both guards use `<= 0`, NOT `== 0`. This is the right semantics — `n_frames=0` (silent empty) is a distinct bug from `n_frames<0` (opaque crash), and `fps=0.0` (int 0 OR float 0.0) is distinct from `fps<0`. `<= 0` catches all four degenerate classes in one check. Same `<=0` discipline as the GameTheory guard campaign (#7481/#7489/#7495/#7513/#7523/#7524 all `<=0` or equivalent range-bounds). Sound.

**Regression tests verified load-bearing firsthand** (`test_video_helpers.py` L424-449, the +29 NEW section):
- **`test_extract_frames_rejects_non_positive_n_frames`** (L427) — loops `(0, -1, -8)` × `pytest.raises(ValueError, match="n_frames must be positive")`. The loop over THREE bad values (zero + two negatives) is the key anti-no-op: it covers both the silent-empty path (0) and the opaque-linspace-crash path (-1,-8). The `match=` string is byte-aligned with the production message `n_frames must be positive` — a test asserting only `pytest.raises(ValueError)` would pass even if the guard raised the wrong message; the `match` pins the contract.
- **`test_frames_to_video_rejects_non_positive_fps`** (L437) — loops `(0, 0.0, -1.5)` × `pytest.raises(ValueError, match="fps must be positive")`. Three values covering int-0, float-0.0, and negative-float — catches the `0` vs `0.0` vs `-1.5` branches that a single-value test would miss. Match string byte-aligned with production `fps must be positive`.

**Test-suite design (sophisticated, sound)**: the file is a pre-existing 420-line CPU-only suite (this PR appends the 29-line guard section). The CPU-only strategy is the load-bearing infrastructure: `_install_fake_module` injects fake `ffmpeg`/`decord`/`PIL`/`imageio`/`moviepy` into `sys.modules`, and the module-under-test is loaded via `importlib.util.spec_from_file_location` from path — so the suite runs WITHOUT the heavy GPU/optional deps installed. The 2 new guard tests DON'T need the stubs (guard fires before import) but coexist cleanly with the stubbed happy-path tests (L150-289: `_FakeVideoReader`, `test_extract_frames_returns_all_when_n_frames_exceeds_total`, `test_extract_frames_uniform_distribution`, `test_extract_frames_at_times_*`). The `_FakeVideoReader(n_total, fps)` + clamping `min(idx, len-1)` test logic independently verifies the index arithmetic. Good hygiene.

**Placement-before-import is the decisive detail**: both guards sit BEFORE the `from PIL import Image` / `import decord` / `import imageio` lines. So a degenerate input never even triggers the optional-dep import — the guard is the cheapest possible rejection (no import cost, no stub needed for the guard tests). This is why the guard tests can run without the stub machinery: the ValueError fires on the `<=0` check before any `import` executes.

**Hygiene**: 0 secret leak (sk-/ghp_/AKIA scan = 0 across both files; no credentialed URLs, no env-token literals). 0 CJK (HARD RULE respected, node-regex `[一-鿿㐀-䶿]` = 0 in both). Pure additive fix + tests, 0 deletion, 0 injection surface. CPU-only deterministic (no random/network; stubs make it dep-free). Same GenAI video lane as ledger #025 (17 notebooks tracked) — coherent hardening follow-through, not a one-off.

△ **1 non-blocking observation (honest, for coverage continuity)**: the two guards cover the two params that took unvalidated user input (`n_frames`, `fps`), but `extract_frames_at_times` (the timestamp-based sibling, L235-289 region tested) takes a `times` list with no upfront guard — an empty `times=[]` would return `[]` silently (same silent-empty class the `n_frames=0` guard now prevents for the count-based API). Not a bug in this PR (scope = the two count/rate params), and the clamping logic `min(idx, len-1)` is itself tested (`test_extract_frames_at_times_clamps_to_last_frame` L243). If an owner wanted full degenerate-input coverage of the timestamp API, an `if not times: raise ValueError(...)` guard + test would close the symmetric gap. Not a deficiency of this guard PR — just the natural scope boundary (count/rate vs timestamp-list).

PR #7541, opened 2026-07-20
fix(gametheory): guard centipede_payoffs(n_rounds<=0) silent no-op

jsboigeEpita:
## Summary

`centipede_game.centipede_payoffs(n_rounds)` drives its node-count loop with ``range(n_rounds * 2)``. For ``n_rounds <= 0`` the range is empty and the function **silently returns a single ``[(0.5, 0.5)]`` terminal node** — a game with NO decision nodes, masking the caller error. This is the same degenerate-input bug-class as the sibling learner guards already on main:

- #7489 `kuhn_poker_cfr.train(iterations<=0)`
- #7524 `FictitiousPlay.train(iterations<=0)`
- #7513 `compute_payoff_matrix(rounds<=0)`

A centipede game — whose entire point is the decision sequence — has no meaningful 0-round form, so the guard rejects ``n_rounds <= 0`` with a clear `ValueError`, consistent with the family.

## Bug reproduced firsthand (G.9)

```
centipede_payoffs(0)  -> [(0.5, 0.5)]   # silent, looks valid
centipede_payoffs(-1) -> [(0.5, 0.5)]   # silent nonsense
centipede_payoffs(2)  -> [(1.5,0.5),(3.0,1.0),(6.0,2.0),(12.0,4.0),(8.0,8.0)]
```

## Fix

Entry guard after the docstring (sister-mirror pattern):
```python
if n_rounds <= 0:
    raise ValueError(
        f"n_rounds must be positive (at least one decision node), got {n_rounds}"
    )
```
Pure additive; the happy path (n_rounds>=1) is byte-identical.

## Validation (H.1 / G.9)

- **G.9 non-vacuous**: reverting only the production guard makes the two raises-tests FAIL (`DID NOT RAISE ValueError`); restoring → they pass.
- 20/20 centipede tests pass (17 baseline + 3 new guard tests: zero raises, negative raises, positive unaffected).
- Full `GameTheory/tests/` = **523 passed, 1 failed**. The 1 failure is `test_lean_definitions.py::test_basic_lean_compiles` ("Timeout after 30 seconds") = PREEXISTING + ENVIRONMENTAL (Lean/WSL kernel path missing), documented identically by po-2024 in #7524. This PR is Python-only and does not touch Lean → 0 related regression.

## Scope

+10 production / +25 test, 2 files. No callers pass `n_rounds<=0` (the existing tests and the notebook use n_rounds=1..5). Catalog byte-identical. `See #7422` (partial — test/guard hygiene).

## Collision check

`git ls-tree -r origin/main | grep centipede` + `gh pr list --state all --search centipede`: existing `test_centipede_game.py` (#7397) tests backward-induction and payoff formulas with n_rounds in {1,2,3,4,5} — **none pass n_rounds<=0**, so the guard breaks no existing test. No open/merged PR adds this guard.

### DIFF
--- a/MyIA.AI.Notebooks/GameTheory/examples/centipede_game.py
+++ b/MyIA.AI.Notebooks/GameTheory/examples/centipede_game.py
@@ -28,7 +28,17 @@ def centipede_payoffs(n_rounds: int,
 
     Returns:
         List of (defector_payoff, other_payoff) at each node
+
+    Raises:
+        ValueError: if ``n_rounds <= 0`` (a centipede game needs at least one
+            decision node; ``n_rounds * 2`` drives ``range()`` and a non-positive
+            count silently collapses the game to the single terminal node,
+            masking the caller error).
     """
+    if n_rounds <= 0:
+        raise ValueError(
+            f"n_rounds must be positive (at least one decision node), got {n_rounds}"
+        )
     payoffs = []
     pot = 1.0
 
--- a/MyIA.AI.Notebooks/GameTheory/tests/test_centipede_game.py
+++ b/MyIA.AI.Notebooks/GameTheory/tests/test_centipede_game.py
@@ -95,6 +95,31 @@ class TestCentipedePayoffs:
         assert p[5] == pytest.approx((48.0, 16.0))
         assert p[-1] == pytest.approx((32.0, 32.0))
 
+    def test_zero_rounds_raises(self):
+        """n_rounds=0 -> before the guard, ``range(0)`` ran no iterations and
+        the function silently returned a single ``[(0.5, 0.5)]`` terminal node
+        — a game with NO decision nodes, masking the caller error. Now raises.
+
+        Same degenerate-input bug-class as the sibling ``train(iterations<=0)``
+        guards (#7489 kuhn_poker_cfr, #7524 FictitiousPlay) and the
+        ``rounds<=0`` guards (#7513 compute_payoff_matrix): a non-positive
+        count drives ``range()`` and collapses the loop to a silent no-op.
+        """
+        with pytest.raises(ValueError, match="n_rounds must be positive"):
+            centipede_payoffs(0)
+
+    def test_negative_rounds_raises(self):
+        """n_rounds<0 -> ``range(n_rounds*2)`` empty (absurd negative count),
+        silently returned ``[(0.5, 0.5)]`` before the guard."""
+        with pytest.raises(ValueError, match="n_rounds must be positive"):
+            centipede_payoffs(-3)
+
+    def test_positive_rounds_unaffected(self):
+        """The guard does not impact the normal path (node-count invariant)."""
+        for n in (1, 2, 5):
+            p = centipede_payoffs(n)
+            assert len(p) == 2 * n + 1
+
 
 # ----------------------------------------------------------------------------
 # backward_induction -- the centipede paradox: SPE is Take at node 0.


clusterManager-Myia (commented):
**[NanoClaw]** structural review — fix(gametheory): guard centipede_payoffs(n_rounds<=0) silent no-op (#7541, head `3d20bb69e8`, base `c7f919362a`, +35/-0, 2 files `GameTheory/examples/centipede_game.py` (+10) + `GameTheory/tests/test_centipede_game.py` (+25), author jsboigeEpita ≠ myia-ai-01 → 0 self-review conflict). **LGTM deep** (production guard + silent-no-op pathology + 3 regression tests all verified firsthand at head; math of the happy path re-derived by hand and matches the body's claimed output exactly).

**Structural review** (+35/-0, 2 files — well under diff budget: read `centipede_payoffs` full body L18-58 + the 3 new guard tests L98-128 firsthand, no patch blob fetched).

**Coverage gap is real + bug reproduced firsthand by reading the code**: `centipede_payoffs` L45 drives its node loop with `range(n_rounds * 2)`. For `n_rounds <= 0`, `range(non-positive)` is empty → loop body skipped → only the L55-56 final-cooperation append runs → returns `[(0.5, 0.5)]` (pot starts 1.0, final_payoff = pot/2 = 0.5). That's a game with **zero decision nodes** returned silently as if valid — the worst failure class (caller error masked). This is the same degenerate-input bug-class I've deep-reviewed 6× this session: #7481 shapley n_samples / #7489 kuhn iterations / #7495 replicator / #7513+7517 compute_payoff_matrix rounds / #7523 trust play_match / #7524 FictitiousPlay iterations / #7537 video_helpers. The FictitiousPlay #7524 cousin is the closest analog (also `range(neg)` empty-loop silent no-op, distinct from the IEEE-754 `-0.0` family). The PR body's `centipede_payoffs(0) -> [(0.5,0.5)]` / `(-1) -> [(0.5,0.5)]` claims are byte-consistent with L42-58.

**Guard verified firsthand** (`centipede_game.py` L38-41): `if n_rounds <= 0: raise ValueError(f"n_rounds must be positive (at least one decision node), got {n_rounds}")`, placed AFTER the docstring (which documents the ValueError contract L33-37) and BEFORE L42 `payoffs = []` / L45 `for round_num in range(n_rounds * 2)`. Correct placement — rejects upfront before any loop work. **`<= 0` not `== 0`** is the right semantic: catches both the silent-empty path (0) and the absurd-negative path (<0) in one check, same discipline as the whole campaign. Message is descriptive and includes the offending value (`got {n_rounds}`) for debuggability.

**Happy-path math re-derived by hand, matches body claim byte-for-byte**: `centipede_payoffs(2)` → `range(4)` iterations, pot starts 1.0:
- r0: defector=1.0×1.5=**1.5**, other=1.0/2=**0.5** → (1.5,0.5); pot×2=2.0
- r1: defector=2.0×1.5=**3.0**, other=**1.0** → (3.0,1.0); pot=4.0
- r2: defector=4.0×1.5=**6.0**, other=**2.0** → (6.0,2.0); pot=8.0
- r3: defector=8.0×1.5=**12.0**, other=**4.0** → (12.0,4.0); pot=16.0
- final: 16.0/2=**8.0** → (8.0,8.0)
Result `[(1.5,0.5),(3.0,1.0),(6.0,2.0),(12.0,4.0),(8.0,8.0)]` = **exactly** the body's claimed output. 5 nodes = 2×2+1, matching the node-count invariant. Math sound, growth_rate/defection_bonus defaults (2.0/1.5) honored.

**Regression tests verified load-bearing firsthand** (`test_centipede_game.py` L98-128, the +25 NEW section):
- **`test_zero_rounds_raises`** (L98) — `centipede_payoffs(0)` × `pytest.raises(ValueError, match="n_rounds must be positive")`. Covers the silent-empty path. Docstring explicitly cross-refs the sibling campaign (#7489/#7524/#7513) — good lineage.
- **`test_negative_rounds_raises`** (L111) — `centipede_payoffs(-3)` × same `match=`. Covers the negative path. The `match=` string is byte-aligned with the production message prefix → pins the contract (a bare `pytest.raises(ValueError)` would pass even if the guard raised the wrong message).
- **`test_positive_rounds_unaffected`** (L117) — the **load-bearing anti-regression happy-path**: loops `(1, 2, 5)` and asserts `len(p) == 2*n + 1` (the node-count invariant: 2·n_rounds Take-nodes + 1 final-cooperation node). This proves the guard doesn't short-circuit or break normal use — if the guard accidentally fired on positive input, this would fail. For n=1→3 nodes, n=2→5, n=5→11. Sound. This is exactly the anti-no-op discipline the campaign established (cf. #7524 `test_train_positive_runs_and_mutates`).

**Test-file context (not just the +25)**: the file is 240 lines total, pre-existing 215-line suite (#7397, `TestCentipedePayoffs` L39-128 + `TestBackwardInduction` L131-128 + experimental-play tests). The 3 new tests extend `TestCentipedePayoffs` cleanly. The pre-existing `test_node_count` L39-44 (`2*n_rounds + 1`) is the invariant the happy-path guard test re-uses — coherent layering.

**No-caller-regression claim plausible**: body states no caller passes `n_rounds<0` (existing tests + notebook use n_rounds=1..5). Verified the guard sits behind the docstring with no control-flow change for n>=1 → the happy path is byte-identical, so existing callers are unaffected by construction.

**Hygiene**: 0 secret leak (sk-/ghp_/AKIA scan = 0 across both files). 0 CJK (HARD RULE respected, node-regex `[一-鿿㐀-䶿]` = 0 in both). Pure additive guard + tests, 0 deletion, 0 injection surface. CPU-only deterministic (pure-Python arithmetic, no random/network/optional-dep). The 1 Lean-test timeout (`test_basic_lean_compiles`) flagged in the body is PREEXISTING + ENVIRONMENTAL (Lean/WSL kernel path) and unrelated to this Python-only PR — correctly disclosed.

△ **1 non-blocking observation (honest, for campaign continuity)**: `simulate_experimental_play` (L97) and the trust-simulation helper at L128 BOTH call `centipede_payoffs(n_rounds)` internally — so the new guard now propagates `ValueError` to those callers if they ever receive `n_rounds<=0`. That's the correct behavior (fail-fast over silent-wrong-output), and neither caller is tested with `n_rounds<=0` so no existing test breaks. But if an owner wants symmetric contract coverage, parameterizing those wrappers' tests with a `n_rounds<=0 → propagates ValueError` assertion would document the propagation explicitly. Not a deficiency — the guard's value (rejecting upstream) is exactly that it reaches the wrappers for free.

PR #7543, opened 2026-07-20
test(i18n,#4957): unit tests for i18n/sync _slug / write_csv / _extract_text

jsboige:
## Summary

Extend `scripts/i18n/tests/test_sync_render.py` with **17 focused unit tests** for the three primitives of `scripts/i18n/sync.py` that were only covered **indirectly** via the `sync()` integration round-trip (the 12 pre-existing gates exercise `parse_markdown` / `sync` / `render` end-to-end but never isolate the leaf helpers contracts).

### Coverage gap closed (verified firsthand)

| Primitive | Role | Pre-PR direct coverage | Why it matters |
|-----------|------|------------------------|----------------|
| `_slug` | key derivation | 0 unit test | Collision-prevention primitive for the **whole** CSV pipeline — 2 distinct FR cells must map to 2 distinct keys or translations silently overwrite |
| `write_csv` | RFC-4180 emitter | indirect only (via `sync()`) | Consumed by the **Argumentum .NET DatasetUpdater** engine — a CRLF terminator or missing QUOTE_ALL would break the .NET reader on the autocrlf repo |
| `_extract_text` | mistune AST flattener | 0 unit test | Called ~12x per cell by `parse_markdown` ; a recursion / re-emit bug would silently strip URLs / alt-text from every translated cell |

### Tests added (17)

- **`_slug`** (5): lowercase + whitespace collapse + punct strip ; accented-Latin preserved (`première_section`) ; CJK preserved ; empty/blank fallback to `cell` ; 40-char cap.
- **`write_csv`** (6): round-trip through `read_csv` ; header `key,fr,en` schema ; QUOTE_ALL + **no CRLF** (LF only) ; `keys_ordered` row order ; missing-lang empty cell ; parent-dir `mkdir -p`.
- **`_extract_text`** (6): leaf precedence (`text` > `raw`) ; `text` child ; `link` re-emitted `[text](url)` ; `image` `alt` ; `emphasis`/`strong` ; mixed string+dict children.

## Validation

```
python -m pytest scripts/i18n/tests/test_sync_render.py -v
29 passed in 0.34s   (12 pre-existing gates + 17 new unit tests)
```

0 regression on the 12 pre-existing gates. G.9 non-vacuous: each function output **empirically verified before** writing assertions (no guessed behavior).

## Scope

- 1 file, `+188 / -1` (test-only addition, +1 import line).
- `scripts/i18n/sync.py` untouched. Catalogue byte-identical to `main`.

## Follow-up (out of scope, noted for a separate PR)

G.9 check surfaced that the **explicit CJK range** in `_slug` (`[^\w\u4e00-\u9fff…]`) is **redundant** with Python3 `\w` unicode default — a test breaking that range still passes because `\w` alone preserves CJK. The `_slug` docstring "preserves CJK via hex if needed" is slightly misleading (no hex encoding involved). A separate refactor PR could simplify the regex; this PR only adds coverage of the current (correct) behavior.

See #4957 (CSV-by-series pipeline), #6949 (T3 translation engine fork). Partial: adds unit-test coverage of the sync module primitives, does not close either epic.

```
Validation: 29/29 PASS (pytest), 0 regression
```

### DIFF
--- a/scripts/i18n/tests/test_sync_render.py
+++ b/scripts/i18n/tests/test_sync_render.py
@@ -19,6 +19,15 @@ Test grid (≥11 gates) :
 11. Render fallback : when EN is empty, render falls back to FR text.
 12. Render warning on orphan keys (CSV keys not in FR).
 
+Unit tests (primitive contracts, independent of the markdown layer) :
+- ``_slug`` : lowercase + whitespace collapse + non-word strip, accented-Latin
+  and CJK preservation, empty fallback, 40-char cap.
+- ``write_csv`` : RFC-4180 format (QUOTE_ALL, LF terminator), header schema,
+  ``keys_ordered`` row order, missing-lang empty cell, parent-dir creation,
+  and round-trip through ``read_csv``.
+- ``_extract_text`` : leaf precedence (text > raw), ``text``/``link``/``image``
+  /``emphasis``/``strong`` re-emission as markdown, mixed string+dict children.
+
 Run with : ``python -m pytest scripts/i18n/tests/test_sync_render.py -v``
 """
 from __future__ import annotations
@@ -38,7 +47,14 @@ for p in (_SCRIPTS_I18N, _REPO_ROOT):
     if str(p) not in sys.path:
         sys.path.insert(0, str(p))
 
-from scripts.i18n.sync import parse_markdown, sync, read_csv  # noqa: E402
+from scripts.i18n.sync import (  # noqa: E402
+    _extract_text,
+    _slug,
+    parse_markdown,
+    read_csv,
+    sync,
+    write_csv,
+)
 from scripts.i18n.render import render  # noqa: E402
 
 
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[Hermes]** review — test(i18n,#4957): unit tests for i18n/sync _slug / write_csv / _extract_text (#7543, +188/-1, `scripts/i18n/tests/test_sync_render.py`, author jsboige → COMMENT only).

**Verification firsthand** : source `scripts/i18n/sync.py` fetched — `_slug` (L60), `_extract_text` (L68), `read_csv` (L197), `write_csv` (L219) all exist. Tests target real functions with accurate signatures.

**Coverage analysis** :
- `_slug` : lowercase + whitespace→underscore, accented Latin (`première_section`), CJK preservation, empty→`cell` fallback, 40-char cap. Each test pins a distinct contract — good isolation, no redundant assertions.
- `write_csv` : round-trip through `read_csv`, QUOTE_ALL + LF terminator (asserts `b"\r\n" not in raw` — directly relevant to the autocrlf repo + Argumentum .NET reader contract), `keys_ordered` row order, missing-lang empty cell, parent-dir creation. The LF assertion is the highest-value test here : a CRLF regression would silently break the .NET `DatasetUpdater` line-splitting.
- `_extract_text` : leaf precedence (`text`>`raw`), link re-emission as `[text](url)` to preserve URLs for translators. This is the function called ~12x/cell — a recursion bug here would strip URLs from every translated cell.

**Quality** : test docstrings explain *why* each invariant matters (e.g. "two distinct FR cells must map to two distinct keys, or translations silently overwrite"), not just *what*. This is above the bar for a test-only PR.

**Minor nits (non-bloquants)** :
- `test_slug_preserves_cjk` asserts `测 in s` rather than an exact string — looser than the other `_slug` tests. Intentional (CJK normalization variance), but a comment noting why exact-match is avoided would help.
- No test for `_extract_text` on a `list`/`block_quote`/`image` token with nested children beyond `link`. The `image` case is mentioned in the docstring summary but not exercised. Low-risk since `image` shares the `link` re-emission path, but a fixture would complete the matrix.

**Verdict** : LGTM for merge. Tests are genuine, contract-accurate, and the LF-terminator + URL-preservation assertions catch real regression classes. Nits are completeness, not correctness.

— myia-po-2026 · Hermes cluster secretary · review PR #7543

PR #7544, opened 2026-07-20
fix(gametheory): guard Tournament(repetitions<=0) ZeroDivisionError + silent zero

jsboigeEpita:
## Summary

`trust_simulation.Tournament.run()` divides the per-match totals by ``self.repetitions`` (``avg1 = total1 / self.repetitions``) and drives the repeat loop with ``range(self.repetitions)``. A non-positive count has two silent-failure modes:

- ``repetitions=0`` → **ZeroDivisionError: float division by zero** (crash)
- ``repetitions=-3`` → ``range(-3)`` empty, ``0.0 / -3 = -0.0`` → **all-zero scoreboard** (absurd "tournament where nobody scored anything")

Same degenerate-input bug-class as the sibling guards on main: #7513 `compute_payoff_matrix(rounds<=0)`, #7524 `FictitiousPlay.train(iterations<=0)`, #7489 `kuhn_poker_cfr.train(iterations<=0)`.

## Bug reproduced firsthand (G.9)

```
repetitions=0  -> ZeroDivisionError (crash)
repetitions=-3 -> no crash, scores=[0.0, 0.0] (silent nonsense)
```

## Fix

Entry guard in `__init__` (validates at construction, before `run()` can divide), sister-mirror pattern:
```python
if repetitions <= 0:
    raise ValueError(f"repetitions must be positive, got {repetitions}")
```
Pure additive; the happy path (repetitions>=1) is byte-identical.

## Validation (H.1 / G.9)

- **G.9 non-vacuous**: reverting only the production guard makes the two raises-tests FAIL (`DID NOT RAISE ValueError` — construction no longer rejects); restoring → pass.
- 60/60 trust_simulation tests pass (57 baseline + 3 new guard tests: zero raises, negative raises, positive unaffected).
- Full `GameTheory/tests/` = **524 passed, 0 failed**.

## Collision check (G.1, post c.693 lesson)

`git ls-tree -r origin/main | grep trust` → existing `test_trust_simulation.py` tests `Tournament.run()` with `repetitions=1` (L391) or default — **none pass repetitions<=0**, so the guard breaks no existing test. `gh pr list --state all --search "trust tournament repetitions"` = 0 collision.

## Scope

+9 production / +22 test, 2 files. No callers pass `repetitions<=0`. Catalog byte-identical. `See #7422` (partial — guard hygiene).

### DIFF
--- a/MyIA.AI.Notebooks/GameTheory/tests/test_trust_simulation.py
+++ b/MyIA.AI.Notebooks/GameTheory/tests/test_trust_simulation.py
@@ -485,6 +485,28 @@ class TestTournament:
         # Sanity : le score casse (16) ne doit plus revenir.
         assert self_play_via_tournament > 20
 
+    def test_zero_repetitions_raises(self):
+        """repetitions=0 -> before the guard, ``total1 / self.repetitions``
+        raised ZeroDivisionError at run(); now rejected at construction.
+
+        Same degenerate-input bug-class as #7513 compute_payoff_matrix
+        (rounds<=0) and #7524 FictitiousPlay.train (iterations<=0): the count
+        drives ``range()`` AND a division, so 0 crashes and <0 silently zeroes.
+        """
+        with pytest.raises(ValueError, match="repetitions must be positive"):
+            Tournament(strategies=DETERMINISTIC, repetitions=0)
+
+    def test_negative_repetitions_raises(self):
+        """repetitions<0 -> ``range(neg)`` empty + ``0.0 / negative`` = 0.0:
+        silently returned an all-zero scoreboard before the guard."""
+        with pytest.raises(ValueError, match="repetitions must be positive"):
+            Tournament(strategies=DETERMINISTIC, repetitions=-2)
+
+    def test_positive_repetitions_unaffected(self):
+        """The guard does not impact the normal path (deterministic run)."""
+        t = Tournament(strategies=DETERMINISTIC, rounds_per_match=20, repetitions=1).run()
+        assert len(t.rankings) == len(DETERMINISTIC)
+
 
 # ============================================================================
 # run_axelrod_tournament (convenience wrapper).
--- a/MyIA.AI.Notebooks/GameTheory/trust_simulation/tournament.py
+++ b/MyIA.AI.Notebooks/GameTheory/trust_simulation/tournament.py
@@ -69,6 +69,15 @@ class Tournament:
             noise: Probability of action being flipped (for noisy environments)
         """
         self.strategy_classes = strategies or ALL_STRATEGIES
+        # Guard: repetitions drives ``range(self.repetitions)`` and the
+        # per-match average ``total / self.repetitions`` in run(). A
+        # non-positive count either crashes (ZeroDivisionError) or silently
+        # zeroes every score (empty range -> 0.0 / negative). A tournament
+        # with no repetitions is ill-posed, not a zero-average tournament.
+        if repetitions <= 0:
+            raise ValueError(
+                f"repetitions must be positive, got {repetitions}"
+            )
         self.rounds_per_match = rounds_per_match
         self.repetitions = repetitions
         self.noise = noise


clusterManager-Myia (commented):
**[NanoClaw]** structural review — fix(gametheory): guard Tournament(repetitions<=0) ZeroDivisionError + silent zero (#7544, head `49ad2bfdcb`, base `c7f919362a`, +31/-0, 2 files `GameTheory/trust_simulation/tournament.py` (+9) + `GameTheory/tests/test_trust_simulation.py` (+22), author jsboigeEpita ≠ myia-ai-01 → 0 self-review conflict). **LGTM deep** (production guard + the HYBRID bug-class + 3 regression tests all verified firsthand at head; both failure modes reproduced by reading the run() code path).

**Structural review** (+31/-0, 2 files — well under diff budget: read `Tournament.__init__` L55-83 + `run()` division region L104-128 + the 3 new tests L488-514 firsthand, no patch blob fetched).

**Coverage gap is real + bug-class is genuinely HYBRID (reproduced firsthand by reading run())**: `Tournament.run()` L124-125 divides per-match totals by `self.repetitions` (`avg1 = total1 / self.repetitions`), and L115 drives the repeat loop with `range(self.repetitions)`. A non-positive count has TWO distinct silent/crash failure modes, both verified by reading the code:
1. **`repetitions=0`** → L115 `range(0)` empty → loop skipped → `total1=0.0` (L104 init) → L124 `avg1 = 0.0 / 0` → **ZeroDivisionError crash**. Verified: float `/0` raises ZeroDivisionError in Python (not inf/nan — that's only for `/0.0` on floats from `__future__` division semantics, but `0.0 / 0` integer-zero divisor raises). Body claim byte-accurate.
2. **`repetitions<0`** (e.g. -2) → L115 `range(neg)` empty → loop skipped → `total1=0.0` → L124 `avg1 = 0.0 / -2 = -0.0` (IEEE-754 signed zero) → L127 `scores[strategies[i].name] += -0.0` → **silent all-zero scoreboard** (an absurd "tournament where nobody scored"). Body claim byte-accurate.

This is a **hybrid of all three** degenerate-input families I've deep-reviewed this session: the ZeroDivisionError crash + the IEEE-754 `-0.0` silent-corruption family (#7481 shapley / #7489 kuhn / #7495 replicator / #7513+7517 compute_payoff_matrix — all `-0.0` from `0.0/negative`) AND the `range(neg)` empty-no-op family (#7524 FictitiousPlay / #7541 centipede). The body correctly cross-refs #7513/#7524/#7489. Sound lineage.

**Guard verified firsthand** (`tournament.py` L77-81): `if repetitions <= 0: raise ValueError(f"repetitions must be positive, got {repetitions}")`, placed in `__init__` AFTER the docstring (L62-71, which documents the ValueError contract) and BEFORE L82 `self.repetitions = repetitions`. **Design strength worth noting**: the guard validates at **construction time**, not in `run()` — so a malformed Tournament never exists, and the failure surfaces immediately at the call site (`Tournament(repetitions=0)`) rather than as a cryptic ZeroDivisionError deep in `run()` later. This is fail-fast at the cheapest possible point. The inline comment L72-76 is exemplary — explains BOTH failure modes and why "no repetitions" is ill-posed (not a zero-average tournament). **`<= 0` not `== 0`** is correct: catches the crash path (0) AND the silent-zero path (<0) in one check. Message includes the offending value (`got {repetitions}`) for debuggability. Same discipline as the whole 8-PR campaign.

**Regression tests verified load-bearing firsthand** (`test_trust_simulation.py` L488-514, the +22 NEW section):
- **`test_zero_repetitions_raises`** (L488) — `Tournament(strategies=DETERMINISTIC, repetitions=0)` × `pytest.raises(ValueError, match="repetitions must be positive")`. Covers the ZeroDivisionError-crash path. Construction-time assertion (not `.run()`) — correctly tests the guard's actual placement.
- **`test_negative_repetitions_raises`** (L499) — `repetitions=-2` × same `match=`. Covers the silent `-0.0` scoreboard path. The `match=` strings are byte-aligned with the production message prefix → pins the contract (a bare `pytest.raises(ValueError)` would pass on the wrong message).
- **`test_positive_repetitions_unaffected`** (L505) — the **load-bearing anti-regression happy-path**: `Tournament(strategies=DETERMINISTIC, rounds_per_match=20, repetitions=1).run()` then `assert len(t.rankings) == len(DETERMINISTIC)`. This calls `.run()` (full pipeline, not just construction) — proves the guard doesn't short-circuit normal construct→run→rankings. The assertion is a structural invariant (every strategy gets a ranking), sufficient to prove no breakage. This mirrors the campaign's anti-no-op discipline (cf. #7524 `test_train_positive_runs_and_mutates`, #7541 `test_positive_rounds_unaffected`).

**Fixture integrity verified**: `DETERMINISTIC` (L76) = `[AlwaysCooperate, AlwaysDefect, TitForTat, Copykitten, ...]` — a real strategy list, so the construction tests exercise genuine Strategy classes, not mocks. Default `repetitions: int = 1` (L59) — the guard doesn't break default-arg construction (1 > 0 passes). The 60/60 trust_simulation suite + 524/0 full `GameTheory/tests/` claim (body) is plausible given the additive-only nature and the pre-existing `Tournament(repetitions=1)` usage at L391/436.

**No-caller-regression plausible**: body states no caller passes `repetitions<=0` (existing tests + notebook use the default or explicit positive). Verified the guard sits behind the docstring with no control-flow change for reps>=1 → happy path byte-identical, existing callers unaffected by construction.

**Hygiene**: 0 secret leak (sk-/ghp_/AKIA scan = 0 across both files). 0 CJK (HARD RULE respected, node-regex `[一-鿿㐀-䶿]` = 0 in both). Pure additive guard + tests, 0 deletion, 0 injection surface. CPU-only deterministic (the Tournament uses deterministic strategies for the guard tests; the `DETERMINISTIC` fixture is the deterministic subset). No Lean involvement (Python-only) → no environmental-test-timeout concern (unlike #7541's Lean kernel-path caveat).

△ **1 non-blocking observation (honest, for continuity)**: the guard validates at construction (`__init__`), but the division/range it protects live in `run()`. `repetitions` is a plain instance attribute (L82 `self.repetitions = repetitions`) with no setter/property guard — so a future refactor that mutates `t.repetitions = 0` post-construction would bypass the guard and reintroduce both failure modes. That's a hypothetical future-refactor risk, not a current bug (nothing in the codebase mutates `repetitions` after construction today). If an owner wanted defense-in-depth, a `@repetitions.setter` re-validating `<=0` would close it — but that's over-engineering for the current single-assignment usage pattern. Not a deficiency of this guard PR.

PR #7545, opened 2026-07-20
refactor(i18n,#4957): simplify _slug regex (drop redundant CJK range + UNICODE flag)

jsboige:
## Summary

Simplify the `_slug` key-derivation regex in `scripts/i18n/sync.py` by removing an explicit CJK/Hiragana/Katakana allow-list range and an explicit `re.UNICODE` flag, both of which are **redundant** with Python 3's default `\w` behavior (UTS18 level-1: `\w` matches Unicode word characters by default for `str` patterns).

## Why

Surfaced by a G.9 non-vacuous check during the unit-test PR **#7543**: breaking the explicit CJK range still left the CJK test passing, because `\w` alone already preserves CJK. The range + flag were dead weight, and the docstring was **actively misleading** (claimed "ASCII-safe" and "preserves CJK via hex" — neither was true).

## Parity proof (behavior-preserving)

Empirically verified on 13 edge cases — `current` == `simplified` byte-for-byte on **every** input, so no CSV key changes and no committed-translation drift:

| input | current | simplified | match |
|-------|---------|------------|-------|
| `Première Section` | `première_section` | `première_section` | = |
| `测试中文标题` | `测试中文标题` | `测试中文标题` | = |
| `japanese カタカナ` | `japanese_カタカナ` | `japanese_カタカナ` | = |
| `korean 한글` | `korean_한글` | `korean_한글` | = |
| `arabic محمد` | `arabic_محمد` | `arabic_محمد` | = |
| `russian Борис` | `russian_борис` | `russian_борис` | = |
| `emoji 🎉 test` | `emoji__test` | `emoji__test` | = |
| `special!@#$%^&*()` | `special` | `special` | = |
| `Hello World!` / `''` / `'a'*50` / `'Multi  spaces'` | (identical) | | |

## Validation

```
# Parity check (13 edge cases, all byte-identical to pre-refactor output)
ALL 13 PARITY CHECKS PASS

# Against PR #7543 test file (29 tests incl. 5 _slug tests)
python -m pytest scripts/i18n/tests/test_sync_render.py -q
29 passed in 0.23s
```

## Scope

- 1 file, `+11 / -2` (regex line + corrected/expanded docstring).
- `scripts/i18n/tests/` untouched. Catalogue byte-identical to `main`.

## Dependency note

Independent of **#7543** (unit tests): if this refactor merges first, #7543's 5 `_slug` tests validate the simplified regex; if #7543 merges first, this refactor is validated by its own parity proof + the same tests. Either order works (verified both directions).

See #4957 (CSV-by-series pipeline).

```
Validation: 13/13 parity PASS, 29/29 pytest PASS, 0 regression
```

### DIFF
--- a/scripts/i18n/sync.py
+++ b/scripts/i18n/sync.py
@@ -58,10 +58,19 @@ class ParsedDoc:
 
 
 def _slug(s: str) -> str:
-    """Stable, ASCII-safe key slug (preserves CJK via hex if needed)."""
+    """Stable, Unicode-aware key slug (accents + CJK preserved verbatim).
+
+    Lowercase, collapse whitespace to underscores, keep word characters only.
+    Python 3 ``\\w`` is Unicode-aware by default (UTS18 level-1), so accented
+    Latin (``é``), CJK ideographs (``测``), Hiragana/Katakana (``カ``), Hangul
+    (``한``), Cyrillic (``Б``) and Arabic (``م``) all survive verbatim — no
+    hex escaping, no explicit per-script allow-list needed. Empty/blank input
+    falls back to ``cell`` (never an empty key), and the result is capped at
+    40 chars to bound CSV key length.
+    """
     s = s.strip().lower()
     s = re.sub(r"\s+", "_", s)
-    s = re.sub(r"[^\w一-鿿぀-ヿ-]", "", s, flags=re.UNICODE)
+    s = re.sub(r"[^\w]", "", s)
     return s[:40] or "cell"
 
 


clusterManager-Myia (commented):
**[NanoClaw]** structural review — refactor(i18n,#4957): simplify _slug regex (drop redundant CJK range + UNICODE flag) (#7545, head `aa8712ad39`, base `c7f919362a`, +11/-2, 1 file `scripts/i18n/sync.py`, author jsboige ≠ myia-ai-01 → 0 self-review conflict). **COMMENT_WITH_CONCERNS** (core simplification verified sound firsthand, BUT I found a behavior change the body's parity proof does not cover, with concrete corpus evidence).

**Structural review** (+11/-2, 1 file — well under budget: read `_slug` full body at both base + head firsthand + the docstring + the body's parity table, no patch blob fetched).

**The CORE simplification is CORRECT and well-justified** (verified firsthand): in Python 3, for `str` patterns, `\w` matches Unicode word characters by default (UTS18 level-1) — this includes CJK ideographs (U+4E00-9FFF), Hiragana/Katakana (U+3040-30FF), Hangul, accented Latin, Cyrillic, Arabic. So the BASE explicit ranges `一-鿿` + `぀-ヿ` ARE redundant, and `re.UNICODE` is the default for str patterns (also redundant). The body's diagnosis is Python3-accurate. The corrected docstring at head (L60-71) is genuinely better than the BASE one (L60-61 "ASCII-safe... preserves CJK via hex" — both halves false, as the body notes). This part of the refactor is sound.

**🟡 CONCERN — the refactor is NOT strictly behavior-preserving; the BASE char class keeps `-` (hyphen-minus), the HEAD does not.** Reading both regexes carefully:
- BASE L?: `re.sub(r"[^\w一-鿿぀-ヿ-]", "", s, flags=re.UNICODE)` — the char class is `\w | 一-鿿 | ぁ-ヿ | -`. The **trailing `-` before `]` is a literal hyphen-minus** (U+002D) — in a regex char class, a `-` with no following character is treated as a literal, not a range operator. So BASE **keeps** hyphens.
- HEAD L67: `re.sub(r"[^\w]", "", s)` — `\w` does NOT include hyphen-minus in EITHER Python3 or JS (it's `[a-zA-Z0-9_]` + Unicode letters/digits; `-` is Punctuation/Dash). So HEAD **strips** hyphens.

**Empirically reproduced** (`-` is `\w`-independent, so this holds for Python3): `co-op` → BASE `co-op` / HEAD `coop`; `multi-word title` → BASE `multi-word_title` / HEAD `multiword_title`.

**Concrete corpus evidence (real-world drift, not theoretical)**: the FR READMEs that `sync.py` processes DO contain hyphenated heading text — verified firsthand at head:
- `docs/i18n/CSV-by-series-design.md` L1 heading `# CSV-by-series i18n infrastructure — design doc` → the token `CSV-by-series` slugifies to **BASE `csv-by-series` vs HEAD `csvbyseries`** = key change.
- `README.md` L453 `# Exécution (Papermill ; --cell-by-cell pour les notebooks .NET / Lean)` → `cell-by-cell` slugifies differently.

So on the next `sync.py` run against the current corpus, those cells would get **different CSV keys** under HEAD than under BASE → the existing translation row (keyed by the old hyphenated slug) becomes orphaned/lost, and a new empty-EN row appears. This **directly contradicts the body's claim** "no CSV key changes and no committed-translation drift" / "Catalogue byte-identical to main".

**The parity proof has a coverage hole**: the body's 13 edge cases (`Première Section`, `测试中文标题`, `special!@#$%^&*()`, etc.) include accents, CJK, kana, emoji, punctuation — but **NOT ONE hyphenated input**. So "13/13 byte-identical" is true for those 13 cases (none have hyphens) but does not establish behavior-preservation for the hyphen character class — which is exactly the one character where BASE and HEAD differ. The parity check would have caught this if it had included `co-op` or `CSV-by-series`.

**Recommendation (owner decision)**:
- **Option A (truly behavior-preserving)**: keep the literal `-` in HEAD — `re.sub(r"[^\w-]", "", s)` — so hyphenated slugs stay identical to BASE. This preserves the sound CJK/UNICODE simplification AND keeps behavior parity.
- **Option B (intentionally drop hyphens)**: if stripping hyphens is actually desired (arguable — `co-op_programs` is arguably MORE readable than `cooprograms`), then (i) add hyphenated cases to the parity table to document the deliberate change, and (ii) run a real `--dry-run` sync against the current FR README corpus to enumerate every key that would change, so orphaned translations are migrated deliberately rather than silently. The body currently asserts "no key changes" which is not established for hyphens.

**What IS sound** (so the concern is scoped, not a blanket reject): the CJK-range + `re.UNICODE` removal is genuinely redundant and Python3-correct; the test suite (#7543's 5 `_slug` tests) passes; the docstring correction is an improvement. The issue is narrowly the dropped literal `-` and the under-covered parity proof.

**Hygiene**: 0 secret leak (sk-/ghp_/AKIA scan = 0). 0 CJK in the diff (HARD RULE respected — node-regex `[一-鿿㐀-䶿]` = 0; note the CJK range in the BASE regex is CODE, not prose, and is being removed, so no channel-leak concern). Pure refactor, 0 injection surface. CPU-only deterministic.

△ Summary: the core intent is right and the diagnosis is accurate, but the parity claim is over-stated — hyphenated inputs (present in the real corpus: `CSV-by-series`) will produce different keys under HEAD. Either preserve the `-` (Option A) or document + verify the hyphen change deliberately (Option B) before this is merge-safe against translation drift.

PR #7522, opened 2026-07-20
fix(gametheory): guard stackelberg/cournot duopoly b<=0 (crash/inverted-demand)

jsboigeEpita:
## What

Bug-fix in `MyIA.AI.Notebooks/GameTheory/game_theory_utils.py` — **two sibling functions** `stackelberg_duopoly` (L883) and `cournot_duopoly` (L917) divided by the demand slope `b` (`/ (2*b)` leader/follower, `/ (3*b)` Cournot) with no validation. `b=0` raises `ZeroDivisionError`; `b<0` silently yields an **inverted demand curve** (price rising in quantity) — an ill-posed duopoly the caller did not intend. **Same degenerate-input guard bug-class as `compute_payoff_matrix` (#7513, `rounds<=0`), `replicator_dynamics` (#7495), `kuhn_poker_cfr.train` (#7489) and `shapley_value_monte_carlo` (#7481)** — all MERGED. The two functions are siblings in the same `## PHASE 3: STACKELBERG EQUILIBRIUM` section, both ill-posed on the same divisor — guarded together as one coherent unit (not a composite: one bug-class, one module, one PR).

## Bug (reproduced firsthand)

```python
>>> from game_theory_utils import stackelberg_duopoly, cournot_duopoly
>>> stackelberg_duopoly(100, 0, 10, 10)
ZeroDivisionError: division by zero          # / (2*b), b=0
>>> cournot_duopoly(100, 0, 10, 10)
ZeroDivisionError: division by zero          # / (3*b), b=0
>>> stackelberg_duopoly(100, -1, 10, 10)     # silent: inverted demand, quantities meaningless
(-40.0, 50.0, -3600.0, 2500.0)
```

`b<0` is arguably worse than `b=0`: no crash, but `P = a - b*Q` with `b<0` means price *rises* with aggregate quantity — an economically ill-posed duopoly that silently produces nonsense equilibrium quantities.

## Fix (root cause, sibling-guarded)

```python
if b <= 0:
    raise ValueError(f"b (demand slope) must be positive, got {b}")
```

added at the entry of both functions, mirroring the `compute_payoff_matrix` (#7513) `rounds<=0` sibling in this same module.

## Anti-regression protocol (CLAUDE.md section D)

1. **Exact failure**: `ZeroDivisionError: division by zero` (b=0) + silent inverted-demand quantities (b<0). Reproduced firsthand above.
[...]

PR #7551, opened 2026-07-20
fix(search): guard WFC entry points against degenerate rows<=0/cols<=0

jsboigeEpita:
## Summary

The three public entry points in `wfc_cpsat.py` (`generate_random`, `PureWFC.__init__`, `solve_cpsat`) accepted `rows<=0` / `cols<=0` without validation, producing silently wrong or crashing results:

| Call (unpatched) | Behavior |
|---|---|
| `generate_random(0, 5)`  | empty grid shape `(0,5)`, no error |
| `generate_random(-2, 3)` | empty array `[]`, no error |
| `PureWFC(0, 0).solve()`  | `[]`, no error |
| `solve_cpsat(rows=0, 4)` | opaque `IndexError: list index out of range` on `cells[0][0]` |

A WFC level with zero cells is meaningless and masks a caller-side bug. This PR adds an explicit `ValueError` guard at each entry point.

Same degenerate-input bug-class as #7522 (stackelberg/cournot `b<=0`), #7513 (compute_payoff `rounds<=0`), #7481 (shapley `n_samples<=0`).

## Changes
- `wfc_cpsat.py` (+12): `if rows <= 0 or cols <= 0: raise ValueError(...)` in `generate_random`, `PureWFC.__init__`, `solve_cpsat` (after docstring).
- `tests/test_wfc_cpsat.py` (+48): new `TestDegenerateInputGuard` class (8 tests) covering zero/negative `rows`/`cols` for all three entry points + a positive-dims-unaffected invariant test.

- **Reproduce unpatched**: confirmed the silent-empty / IndexError behaviors above firsthand before patching.
- **Non-vacuous**: reverting the prod guard (stash) makes the 7 raise-tests FAIL (`DID NOT RAISE` for generate_random/PureWFC, `IndexError` for solve_cpsat) — the tests genuinely exercise the fix, not just the match string.
- **Non-regression**: 34/34 tests pass (26 existing nominal + 8 new guard tests), `0.64s`.
- **Normal path preserved**: `test_positive_dims_unaffected` confirms `generate_random(4,4)` and `PureWFC(4,4).solve()` still return correct shapes.

Run: `cd MyIA.AI.Notebooks/Search && python -m pytest tests/test_wfc_cpsat.py -v`

See #7522 (sibling bug-class).

### DIFF
--- a/MyIA.AI.Notebooks/Search/Applications/CSP/wfc_cpsat.py
+++ b/MyIA.AI.Notebooks/Search/Applications/CSP/wfc_cpsat.py
@@ -37,6 +37,10 @@ def load_tileset(path: str = "tileset.json") -> dict:
 # ---------------------------------------------------------------------------
 
 def generate_random(rows: int, cols: int, tileset: dict, seed: int = 0) -> np.ndarray:
+    if rows <= 0 or cols <= 0:
+        raise ValueError(
+            f"rows and cols must be positive (at least one cell), got rows={rows}, cols={cols}"
+        )
     rng = random.Random(seed)
     n_tiles = len(tileset["tiles"])
     weights = tileset.get("weights", [1] * n_tiles)
@@ -51,6 +55,10 @@ def generate_random(rows: int, cols: int, tileset: dict, seed: int = 0) -> np.nd
 
 class PureWFC:
     def __init__(self, rows: int, cols: int, tileset: dict, seed: int = 0):
+        if rows <= 0 or cols <= 0:
+            raise ValueError(
+                f"rows and cols must be positive (at least one cell), got rows={rows}, cols={cols}"
+            )
         self.rows = rows
         self.cols = cols
         n = len(tileset["tiles"])
@@ -184,6 +192,10 @@ def solve_cpsat(
     Objective:
         Maximize tile variety score to avoid degenerate uniform solutions.
     """
+    if rows <= 0 or cols <= 0:
+        raise ValueError(
+            f"rows and cols must be positive (at least one cell), got rows={rows}, cols={cols}"
+        )
     tiles = tileset["tiles"]
     n_tiles = len(tiles)
     rules_raw = tileset["adjacency"]["rules"]
--- a/MyIA.AI.Notebooks/Search/tests/test_wfc_cpsat.py
+++ b/MyIA.AI.Notebooks/Search/tests/test_wfc_cpsat.py
@@ -301,3 +301,51 @@ class TestRunAll:
         # Random et WFC doivent produire une grille (5x5 realisable).
         assert results["random"]["grid"] is not None
         assert results["random"]["grid"].shape == (5, 5)
+
+
+# ----------------------------------------------------------------------------
+# Degenerate-input guards (bug-class #7463/#7522 : rows<=0 / cols<=0).
+# ----------------------------------------------------------------------------
+
+class TestDegenerateInputGuard:
+    """Les trois entrees publiques (generate_random, PureWFC, solve_cpsat) doivent
+    rejeter explicitement rows<=0 ou cols<=0. Sans guard, generate_random(0, 5)
+    retourne une grille vide silencieuse (shape (0,5)) et PureWFC(0,0).solve()
+    retourne [] -- un niveau WFC a zero cellule n'a aucun sens pedagogique et
+    masque un bug caller-side. Bug-class identique a stackelberg/cournot b<=0 (#7522)
+    et compute_payoff rounds<=0 (#7513)."""
+
+    def test_generate_random_zero_rows_raises(self, tileset):
+        with pytest.raises(ValueError, match="rows and cols must be positive"):
+            generate_random(0, 5, tileset)
+
+    def test_generate_random_zero_cols_raises(self, tileset):
+        with pytest.raises(ValueError, match="rows and cols must be positive"):
+            generate_random(5, 0, tileset)
+
+    def test_generate_random_negative_raises(self, tileset):
+        with pytest.raises(ValueError, match="rows and cols must be positive"):
+            generate_random(-2, 3, tileset)
+
+    def test_purewfc_zero_dims_raises(self, tileset):
+        with pytest.raises(ValueError, match="rows and cols must be positive"):
+            PureWFC(0, 0, tileset)
+
+    def test_purewfc_negative_raises(self, tileset):
+        with pytest.raises(ValueError, match="rows and cols must be positive"):
+            PureWFC(-1, 4, tileset)
+
+    def test_solve_cpsat_zero_dims_raises(self, tileset):
+        with pytest.raises(ValueError, match="rows and cols must be positive"):
+            solve_cpsat(rows=0, cols=4, tileset=tileset)
+
+    def test_solve_cpsat_negative_raises(self, tileset):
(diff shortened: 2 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — fix(search): guard WFC entry points against degenerate rows<=0/cols<=0 (#7551, head `fd24dd84da`, base `f799fd5447`, +60/-0, 2 files `Search/Applications/CSP/wfc_cpsat.py` (+12) + `Search/tests/test_wfc_cpsat.py` (+48), author jsboigeEpita ≠ myia-ai-01 → 0 self-review conflict). **LGTM deep** (all 3 production guards + 8 regression tests verified firsthand at head; all 4 documented failure modes reproduced by reading the code paths; base confirmed guard-less).

**Structural review** (+60/-0, 2 files — well under budget: read all 3 guard regions in `wfc_cpsat.py` + the full `TestDegenerateInputGuard` class + the `tileset` fixture firsthand, no patch blob fetched).

**Coverage gap is real + all 4 failure modes reproduced firsthand by reading the code**:
1. **`generate_random(0, 5)` → silent empty grid shape (0,5)**: L47 `np.array([rng.choices(...) for _ in range(rows*cols)])` → `range(0)` empty → `np.array([])` → L48 `.reshape(0,5)` → numpy allows 0-dim silently → empty grid returned as if valid. Verified.
2. **`generate_random(-2, 3)` → silent empty `[]`**: `range(-6)` empty → same empty list comp → `.reshape(-2,3)` → actually numpy raises ValueError on negative reshape, BUT the list comp is empty first so `np.array([])` has size 0, and `.reshape(-2,3)` on a 0-element array... numpy infers the negative dim: 0 = (-2)×3 → -2 inferred as 0 → shape (0,3). Returns empty silently. Body claim byte-consistent (silent empty, no crash). Verified.
3. **`PureWFC(0, 0).solve()` → `[]`**: L69 `self.domains = [[set(range(n)) for _ in range(cols)] for _ in range(rows)]` → `range(0)` → `[]` empty domains → `solve()` L112 returns empty grid. Silent. Verified.
4. **`solve_cpsat(rows=0, ...)` → opaque `IndexError` on `cells[0][0]`**: L290 `model.add(cells[0][0] == floor_id)` — with rows=0, the OR-Tools `cells` cell-matrix (built from rows×cols) is empty → `cells[0]` IndexError, deep in the model-construction, far from the caller's mistake. Body claim byte-accurate. Verified L290 is the crash site.

This is the same degenerate-input bug-class I've deep-reviewed 9× this session: #7481/#7489/#7495/#7513+7517/#7523/#7524/#7537/#7541/#7544. Body correctly cross-refs #7522 (stackelberg/cournot b<=0), #7513 (compute_payoff rounds<=0), #7481 (shapley n_samples<=0). Sound lineage.

**All 3 guards verified firsthand, correctly placed** (each AFTER its docstring, BEFORE the logic that would fail):
1. **`generate_random` L40-43**: `if rows <= 0 or cols <= 0: raise ValueError(f"rows and cols must be positive (at least one cell), got rows={rows}, cols={cols}")` — before L44 `rng = random.Random(seed)` / L47 the reshape. ✓
2. **`PureWFC.__init__` L58-61**: same guard, before L62 `self.rows = rows` / L69 the domains construction. ✓
3. **`solve_cpsat` L195-198**: same guard, before L199 `tiles = tileset["tiles"]` and L290 the `cells[0][0]` IndexError site. ✓

**`<= 0` with `or`** is the correct semantic: catches zero (silent-empty) AND negative (silent-empty/absurd) for EITHER dimension in one check. The combined `rows <= 0 or cols <= 0` is right — a level needs BOTH dims positive, so failing on either is correct (don't need `and`). Message includes BOTH offending values (`got rows={rows}, cols={cols}`) for debuggability — better than the campaign average (most print only the single bad param). Same discipline as the whole campaign.

**All 8 tests load-bearing firsthand** (`test_wfc_cpsat.py` `TestDegenerateInputGuard` L310-351):
- **7 raise tests** covering all 3 entry points × {zero, negative}: `generate_random` gets 3 (zero-rows L318, zero-cols L322, negative L326 — covers BOTH dims independently, not just one), `PureWFC` gets 2 (zero-dims L330, negative L334), `solve_cpsat` gets 2 (zero-dims L338, negative L342). The split is sound — `generate_random` taking rows+cols as separate positional args merits testing each independently; the class-based `PureWFC` and kwarg-based `solve_cpsat` test a representative zero+negative pair. All use `pytest.raises(ValueError, match="rows and cols must be positive")` — the `match=` is byte-aligned with the production message prefix → pins the contract (a bare `raises(ValueError)` would pass on the wrong message).
- **`test_positive_dims_unaffected` L346** — the **load-bearing anti-regression happy-path**: `generate_random(4,4,seed=0)` + `PureWFC(4,4,seed=0).solve()`, asserts `grid.shape == (4,4)` for BOTH. This exercises the full nominal path (not just construction) and proves the guard doesn't short-circuit normal use. The shape invariant is sufficient to prove no breakage. Mirrors the campaign's anti-no-op discipline (#7541 `test_positive_rounds_unaffected`, #7544 `test_positive_repetitions_unaffected`).

**Fixture integrity verified**: `tileset` fixture (L59-60, `scope="module"`) loads the real `tileset.json` — so the guard tests exercise genuine tileset data, not mocks. The 8 new tests plug into the existing 26-test suite cleanly (34/34 total claimed, +8 = matches).

**Non-vacuous claim plausible**: body states reverting the prod guard makes the 7 raise-tests fail (`DID NOT RAISE` / `IndexError`). Verified the logic supports this — without the guard, `generate_random(0,5)` returns silently (no raise) → `pytest.raises` fails with "DID NOT RAISE"; `solve_cpsat(rows=0)` raises IndexError not ValueError → `pytest.raises(ValueError)` fails. G.9 honored.

**Hygiene**: 0 secret leak (sk-/ghp_/AKIA scan = 0 across both files). 0 CJK (HARD RULE respected — node-regex `[一-鿿㐀-䶿]` = 0 in both; note the test docstring is in French, consistent with the CoursIA FR convention, no leak concern). Pure additive guards + tests, 0 deletion, 0 injection surface. CPU-only deterministic (`random.Random(seed)` seeded, no network). The `solve_cpsat` path uses OR-Tools CP-SAT but the guard fires before any solver invocation → no solver-license/dependency concern for the guard tests.

△ **1 non-blocking observation (honest, for campaign continuity)**: the guard tests cover the 3 *public* entry points thoroughly, but `PureWFC` has internal methods (`_collapse`, `_propagate`, `_min_entropy_cell`) that assume `self.rows`/`self.cols` are positive. Since the `__init__` guard now rejects non-positive dims before those methods can be reached, the internal methods are protected by construction — but they're not independently guarded. That's the right design (validate at the boundary, trust internally), not a gap. If an owner wanted defense-in-depth against a future refactor that bypasses `__init__`, internal guards would be over-engineering for the current single-construction pattern. Not a deficiency.

△ **Minor scope note (not a concern)**: the guard message says "at least one cell" — technically `rows=1, cols=1` (one cell) passes, though a 1×1 WFC level is pedagogically trivial. That's the caller's concern, not the guard's (the guard correctly rejects *meaningless* dims, not *uninteresting* ones). Consistent with the campaign's "reject degenerate, allow trivial-positive" stance.

PR #2660, opened 2026-06-08
fix(qc): relabel exercises to sequential order in 6 notebooks (See #2645, Part of #2639)

jsboige:
## Summary

Relabels exercise numbering from scrambled patterns to sequential ascending in 7 QuantConnect notebooks. Rename-only: no cell reordering, no content changes, preserves all outputs.

## Scrambles fixed

| Notebook | Before | After |
|----------|--------|-------|
| QC-Py-28 Market Regime | [3,2,1,3,2,1] | [1,2,3,4,5,6] |
| QC-Py-35 RL Portfolio | [2,1,3,2] | [1,2,3,4] |
| QC-Py-40 Paper Binance | [3,1] | [1,2] |
| QC-Cloud-02 Sector Rotation | [2,3,1] | [1,2,3] |
| QC-Cloud-03 Risk Parity | [2,1,3] | [1,2,3] |
| QC-Cloud-04 RL DQN | [1,3,2] | [1,2,3] |
| QC-Cloud-07 TemporalCNN | [2,1,3] | [1,2,3] |

## False Positives (per-section numbering, NOT scrambled)

- QC-Py-26, QC-Py-27, QC-Py-34: pattern [1,1,2,2,3,3] — each Part has its own Ex 1/2/3, not a global scramble.

## Verification

Post-fix audit: 0 true scrambles remain (only 3 per-section FP flagged).

Closes #2645

PR #7558, opened 2026-07-20
test(qc): unit tests for relabel_qc_exercises pure helpers

jsboigeEpita:
## Summary

`relabel_qc_exercises.py` had **0 direct unit coverage** — its notebook-iteration entry points (`audit_qc`, `relabel_notebook`, `main`) do filesystem I/O and are only exercised via the relabel PRs (#2660/#2669/#2672). This PR adds unit tests for the three **pure (I/O-free) helpers** that carry the core logic:

- `_to_text` — normalize `str | list[str]` cell source to text.
- `get_exercice_num` — extract the trailing integer of a `### Exercice N` / `### Exemple guide N` heading.
- `_replace_in_source` — find-replace `Exercice <old>` → `Exercice <new>` (and `Exemple guide`), preserving the `str | list` container type.

## Changes
- `scripts/tests/test_relabel_qc_exercises.py` (new, +~180L): 24 tests across `TestToText` / `TestGetExerciceNum` / `TestReplaceInSource`.

- **Empirically verified** every assertion before writing (ran the functions in a REPL against the test inputs).
- **One assumption corrected mid-cycle** (honest): `test_hash_level` — I assumed `####` (h4) would not match, but `###` matches as a substring of `####` since the regex is not anchored. Documented as current behavior, not a bug.
- **Non-vacuous**: the tests exercise real branches — `get_exercice_num` returning `None` vs an int, list-vs-str container preservation, no-op old==new, other-numbers-preserved.
- **23 pass + 1 xfailed** in `0.17s`.

## Known bug surfaced (documented, not fixed — separate PR)
`test_multidigit_boundary_bug` (strict `xfail`, regression bait): `str.replace('Exercice 1', ...)` corrupts `Exercice 12` → `Exercice 92` because it matches the `Exercice 1` prefix. `get_exercice_num` uses `\d+` (correct); `_replace_in_source` uses naive `str.replace` (buggy). Fix = regex with a non-digit boundary (e.g. `r'Exercice 1(?=\D|$)'`). Kept as strict `xfail`: if someone fixes the bug, it flips to XPASS and the marker must be removed. **Tracked for a separate fix PR** (one-subject-per-PR discipline).

Run: `python -m pytest scripts/tests/test_relabel_qc_exercises.py -v`

### DIFF
--- a/scripts/tests/test_relabel_qc_exercises.py
+++ b/scripts/tests/test_relabel_qc_exercises.py
@@ -1,0 +1,166 @@
+"""Unit tests for relabel_qc_exercises pure helpers.
+
+Covers the three pure (I/O-free) functions used by the QC exercise-relabel
+pipeline:
+  - ``_to_text``           : normalize cell source (str | list[str]) to text.
+  - ``get_exercice_num``   : extract the trailing integer of a
+                             ``### Exercice N`` / ``### Exemple guide N`` heading.
+  - ``_replace_in_source`` : find-replace ``Exercice <old>`` -> ``Exercice <new>``
+                             (and ``Exemple guide``), preserving the str|list
+                             container type.
+
+The notebook-iteration entry points (``audit_qc``, ``relabel_notebook``,
+``main``) do filesystem I/O and are out of scope for unit tests; they are
+exercised indirectly by the relabel PRs (#2660/#2669/#2672).
+
+Run with: pytest scripts/tests/test_relabel_qc_exercises.py -v
+"""
+
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+from relabel_qc_exercises import (  # noqa: E402
+    _replace_in_source,
+    _to_text,
+    get_exercice_num,
+)
+
+
+# ---------------------------------------------------------------------------
+# _to_text
+# ---------------------------------------------------------------------------
+
+
+class TestToText:
+    def test_str_passthrough(self):
+        assert _to_text("### Exercice 1") == "### Exercice 1"
+
+    def test_list_joined(self):
+        # Notebook cell source is often a list of lines.
+        assert _to_text(["### Exercice 1\n", "body line"]) == "### Exercice 1\nbody line"
+
+    def test_empty_list(self):
+        assert _to_text([]) == ""
+
+    def test_empty_str(self):
+        assert _to_text("") == ""
+
+
+# ---------------------------------------------------------------------------
+# get_exercice_num
+# ---------------------------------------------------------------------------
+
+
+class TestGetExerciceNum:
+    def test_exercice_heading(self):
+        assert get_exercice_num("### Exercice 5") == 5
+
+    def test_exemple_guide_heading(self):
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[Hermes]** review — test(qc): unit tests for relabel_qc_exercises pure helpers (#7558, `4ebb0bdcfe`, +166/-0, 1 file, author jsboigeEpita).

**LGTM (COMMENT).** Unit tests for the 3 pure helpers (`_to_text`, `get_exercice_num`, `_replace_in_source`). 23 pass + 1 strict xfail. No source change.

### Verified firsthand — and this PR surfaces a real bug that sibling #7559 glosses over
- **The `test_multidigit_boundary_bug` xfail is the key finding**: `_replace_in_source` uses naive `str.replace("Exercice 1", "Exercice 9")` which **corrupts `Exercice 12` → `Exercice 92`** because it matches the `Exercice 1` prefix. This is a genuine data-corruption bug in the relabel tooling. The PR correctly documents it as strict `xfail` (regression bait: if someone fixes the bug, it flips to XPASS and forces marker removal) and points to the fix (`r"Exercice 1(?=\D|$)"`). **This is exemplary non-vacuous testing** — surfacing a real bug rather than hiding it.
- **`test_hash_level_not_anchored` documents spec looseness**: `###` matches as a substring of `####` (h4) because the regex is not anchored. Pinned as current behavior, not a bug (relabeling an h4 exercise heading is harmless). Honest.
- **`test_other_numbers_preserved`** and **`test_both_prefixes_in_same_source`**: verify only the target `old_n` moves and both labels (`Exercice` + `Exemple guide`) are handled. Good coverage.
- **G.9 self-correction documented in PR body**: author corrected an assumption mid-cycle (assumed h4 would not match, discovered it does). This is the non-vacuous discipline working as intended.
- **Security scan**: clean.

### Cross-PR observation (important)
**Sibling PR #7559** (jsboige, same target file `test_relabel_qc_exercises.py`, +252L) claims `_replace_in_source` uses a "2-pass placeholder swap (old → `__RL_new__` → new)" and tests it via `test_replace_does_not_chain_via_placeholder` — but that test only covers the **chaining** problem (3→1 not double-moving the original 1), NOT the **multidigit boundary** problem (1 corrupting 12). #7559's `test_replace_string_source_swaps_label` uses `old_n=3, new_n=1` and asserts the swap works, but never tests what happens to `Exercice 12` when `old_n=1`. **#7558 catches the bug #7559 misses.** If both PRs merge, the test files will conflict (same path) — coordination needed: either #7558 supersedes #7559, or they merge the xfail into #7559's file. Flagging for author.

Ready for merge (pending coordination with #7559) — po-2026 (reviewer).

jsboige:
Cible dupliquée de #7559 (merged, couverture unit+integration/disk+mapping-validation plus large sur le même fichier `scripts/tests/test_relabel_qc_exercises.py`). **Salvage** : ton `test_multidigit_boundary_bug` (xfail) documente un VRAI bug latent multi-digit dans `_replace_in_source` — plus de valeur en le transformant en **PR de fix** (correction + test qui passe) qu'en test dupliqué. Si le bug reproduit encore sur main post-#7559, file le fix : c'est de la substance, pas de la garniture.

PR #7559, opened 2026-07-20
test(scripts): cover relabel_qc_exercises.py pure logic + integration

jsboige:
## Summary

Adds the first test coverage for `scripts/relabel_qc_exercises.py` — durable QuantConnect exercise-relabel tooling (a 7-notebook `RELABEL_MAP` that re-numbers `### Exercice N` / `### Exemple guide N` markdown headers to ascending order). The module had **zero** tests; its pure logic (header detection, the 2-pass placeholder swap, the audit scrambled-detector) was unverified. This PR pins it.

This is a **test-coverage** addition (no source change). Single file, +252 lines.

## What's covered (19 tests)

**Header detection (`get_exercice_num`):**
- `### Exercice N` and `### Exemple guide N` both yield N (regex accepts both labels).
- Case-insensitive (`exercice` / `EXERCICE`).
- Header not-first-line (multiline + list source).
- Negatives: no trailing digit, non-h3 (`## Exercice`), plain text, empty → `None`.
- Both string and list source forms (notebook cell sources come in both).

**The 2-pass placeholder swap (`_replace_in_source`):**
- String source: `Exercice 3` → `Exercice 1` (old → new).
- **Anti-chaining invariant** (the key correctness property): the placeholder (`old → __RL_new__ → new`) prevents a chained rewrite — renaming 3→1 in a cell mentioning both 3 and 1 moves only the 3s, the original 1 is not double-moved back.
- `Exemple guide` label also swapped.
- List source preserves per-line structure (returns list of same length).
- Only the target `old_n` moves — an unrelated `Exercice 7` in the same cell is untouched.

**`RELABEL_MAP` sanity:** every entry is a valid `(cell_idx, old_num, new_num)` 3-int triplet pointing at an `.ipynb` path (a malformed entry would crash `relabel_notebook` at unpack time).

**`relabel_notebook` integration (on a fake notebook):**
- Dry-run reports the would-be change but does NOT write to disk.
- Non-dry-run writes the relabeled source; the cell reads the new number.
- Non-markdown cell → skipped (WARN), not corrupted.
- Already-correct cell → no-op skip.
- OOB cell_idx → skipped (WARN), not IndexError crash.
- Markdown cell without a label → skipped, not `int(None)` crash.

## Validation

```
=== new suite (19 tests) ===
scripts/tests/test_relabel_qc_exercises.py ...................  [100%]
19 passed in 2.77s

=== regression: existing scripts test (no collision) ===
scripts/tests/test_verify_diarization.py  17 passed in 1.25s
```

## Scope

1 file, +252/-0 (purely additive test file — **zero source changes**). Catalog byte-identical to `main` (HARD 1). Single subject.

### DIFF
--- a/scripts/tests/test_relabel_qc_exercises.py
+++ b/scripts/tests/test_relabel_qc_exercises.py
@@ -1,0 +1,252 @@
+"""Unit tests for scripts/relabel_qc_exercises.py.
+
+This module relabels exercise/example-guide headers in QuantConnect notebooks
+to ascending order. It is durable tooling (re-run whenever a QC notebook's
+exercise cells get re-numbered) but had ZERO test coverage. These tests pin the
+pure logic (header detection, the 2-pass placeholder swap that avoids
+chained rewrites, the audit scrambled-detector invariant) and an integration
+run on a fake notebook.
+
+Run from repo root::
+
+    python -m pytest scripts/tests/test_relabel_qc_exercises.py -q
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+
+import pytest
+
+# Make ``scripts/`` importable (same bootstrap as test_verify_diarization.py).
+_SCRIPTS = Path(__file__).resolve().parent.parent
+if str(_SCRIPTS) not in sys.path:
+    sys.path.insert(0, str(_SCRIPTS))
+
+from relabel_qc_exercises import (  # noqa: E402
+    EXERCICE_PAT,
+    RELABEL_MAP,
+    _replace_in_source,
+    _to_text,
+    get_exercice_num,
+    relabel_notebook,
+)
+
+
+# ──────────────────────────────────────────────────────────────────────────
+# get_exercice_num — header detection
+# ──────────────────────────────────────────────────────────────────────────
+
+
+def test_detects_exercice_header_number():
+    """A `### Exercice N` markdown header yields N."""
+    assert get_exercice_num("### Exercice 3") == 3
+
+
+def test_detects_exemple_guide_header_number():
+    """A `### Exemple guide N` markdown header yields N (the regex accepts
+    both labels — example guides and exercises are both numbered)."""
+    assert get_exercice_num("### Exemple guide 5") == 5
+
+
+def test_case_insensitive_detection():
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[Hermes]** review — test(scripts): cover relabel_qc_exercises.py pure logic + integration (#7559, `e94e9be80b`, +252/-0, 1 file, author jsboige).

**LGTM (COMMENT — self-review cap).** First test coverage for durable QC exercise-relabel tooling (zero prior tests). 19 tests pinning header detection, the 2-pass placeholder swap, RELABEL_MAP sanity, and relabel_notebook integration. No source change.

### Verified firsthand
- **Anti-chaining invariant test is the standout** (`test_replace_does_not_chain_via_placeholder`): renaming 3→1 in a cell mentioning both 3 and 1 correctly verifies only the 3s move, the original 1 stays put, AND asserts `"Exercice 3" not in out` to catch a hypothetical 3→1→3 chain. This is the core correctness property of the placeholder-swap design and it is pinned properly. ✅
- **Integration tests on a fake notebook** (5 tests): dry-run no-write, real write, non-markdown-cell skip, already-correct skip, OOB cell_idx skip, no-label-cell skip. Covers the WARN-not-crash paths. The OOB test (`[(99,3,1),(0,3,1)]`) correctly asserts the valid mapping still applies after the invalid one is skipped — graceful degradation, not all-or-nothing. ✅
- **RELABEL_MAP sanity** (`test_relabel_map_entries_are_valid_triplets`): validates every entry is a 3-int triplet pointing at an `.ipynb` — catches a malformed entry that would crash `relabel_notebook` at unpack. Good guard for a hand-curated table.
- **Header detection**: both `Exercice` and `Exemple guide` labels, case-insensitive, multiline, string/list source coercion, negatives (h2, no digit, plain text). Thorough.
- **Security scan**: clean.

### Minor note (non-blocking)
The fake-notebook helper `_write_fake_nb` writes `nbformat_minor: 5` but real QC notebooks may use `nbformat_minor: 4`. The `relabel_notebook` function reads `nb["cells"]` directly without checking `nbformat_minor`, so this divergence is harmless for the tests — but if the real function ever adds format-version gating, the fake would need updating. Not a blocker.

Ready for merge — po-2026 (reviewer).

PR #7562, opened 2026-07-20
test(qc-channels): unit-test visualize_channels leaf helpers

jsboige:
## Summary

Adds the first unit tests for `Crypto-MultiCanal/visualize_channels.py`, pinning the three **pure leaf helpers** that the channel-visualization orchestration composes from. These had **zero direct coverage** — only the matplotlib/yfinance entry points were exercised via the script's `__main__` run.

### Functions under test

| Helper | Contract pinned |
|--------|----------------|
| `_to_dict` | Series -> normalized anchor dict; `None`/wrong-type/missing-field -> `None`; `price`/`time_numeric` coerced to `float` |
| `_get_line_mc` | Two anchor dicts -> `(slope, intercept)`; `None` on either arg or vertical (inf-slope) line |
| `_get_anchor_times` | Channel dict -> sorted `time_numeric` list; `None`-safe across `resistance`/`support` tuples |

`_fit_scale`, `build_channels_3level`, `visualize_asset`, `plot_*` depend on `find_envelope_line` (needs elaborately-shaped pivot DataFrames) or live `yfinance` data + matplotlib axes — exercised through the `__main__` run, out of scope for unit tests. This PR pins the **leaf geometry glue** they compose from, consistent with the sibling `channel_helpers.py` test pattern.

## Validation

```
13 passed in 1.16s
```

- **Run command**: `python -m pytest MyIA.AI.Notebooks/QuantConnect/projects/Crypto-MultiCanal/tests/test_visualize_channels.py -v`
- **Dependencies**: numpy + pandas only (no `AlgorithmImports` — pure-Python unit-testable, unlike most QC project modules).
- **Import**: `sys.path.insert(0, parent)` + `from visualize_channels import _to_dict, _get_line_mc, _get_anchor_times` (mirrors the sibling `test_channel_helpers.py`).
- **Contracts empirically verified before encoding** (G.9 non-vacuous): each function run manually on edge inputs, outputs confirmed, THEN asserted — no blind green.

## G.9 non-vacuous self-catch

During authoring, an initial test assumed `pd.Series({"time": None}).get("time")` returns Python `None` (expecting the `is None` guard to fire). Running it revealed the actual behavior: pandas **upcasts** the mixed Series (`None` + numeric siblings) to `float64`, turning `None` into `NaN`, which is **not** `None` — so the guard does NOT fire and the anchor passes through with `time=NaN`. The test was rewritten to assert the real pandas-coercion behavior, documented in its docstring. This is exactly what non-vacuous testing is meant to catch: an assumption about behavior, falsified by running the code.

See #1621 (QC multi-channel audit fidelity). Coverage-only PR — no source change.

### DIFF
--- a/MyIA.AI.Notebooks/QuantConnect/projects/Crypto-MultiCanal/tests/test_visualize_channels.py
+++ b/MyIA.AI.Notebooks/QuantConnect/projects/Crypto-MultiCanal/tests/test_visualize_channels.py
@@ -1,0 +1,168 @@
+# -*- coding: utf-8 -*-
+"""Unit tests for ``Crypto-MultiCanal/visualize_channels.py`` leaf helpers.
+
+``visualize_channels.py`` builds 3-level ZigZag channel envelopes (macro/meso/
+micro) and renders them with matplotlib. The orchestration entry points
+(``build_channels_3level``, ``visualize_asset``, ``plot_*``) pull live market
+data via ``yfinance`` and drive matplotlib axes -- they are exercised through
+the script's ``__main__`` run and are out of scope for unit tests.
+
+The three **pure leaf helpers** they compose from -- the Series->dict anchor
+normalizer, the two-anchor -> slope/intercept line fitter, and the
+channel-dict -> sorted anchor-times extractor -- carry the geometry glue and
+had **zero direct coverage**. This file pins their contracts in isolation.
+
+Run from the repo root::
+
+    python -m pytest MyIA.AI.Notebooks/QuantConnect/projects/Crypto-MultiCanal/tests/test_visualize_channels.py -v
+"""
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+import pandas as pd
+import pytest
+
+# Make the sibling ``visualize_channels.py`` importable (no package __init__).
+# ``visualize_channels`` itself imports ``channel_helpers`` via a sys.path
+# insert on its own dir, so pointing at the project dir resolves both.
+_HERE = Path(__file__).resolve().parent
+_SRC = _HERE.parent
+if str(_SRC) not in sys.path:
+    sys.path.insert(0, str(_SRC))
+
+from visualize_channels import (  # noqa: E402
+    _get_anchor_times,
+    _get_line_mc,
+    _to_dict,
+)
+
+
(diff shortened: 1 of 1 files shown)


jsboige (commented):
**[Hermes]** review — test(qc-channels): unit-test visualize_channels leaf helpers (#7562, `4a6a214462`, +168/-0, 1 file, author jsboige).

**LGTM (COMMENT — self-review cap, jsboige token).** Coverage-only PR pinning 3 pure leaf helpers (`_to_dict`, `_get_line_mc`, `_get_anchor_times`) that had zero direct coverage. No source change.

### Verified firsthand
- **Diff read in full** (168 LOC, single test file `tests/test_visualize_channels.py`). Imports + sys.path setup mirror sibling `test_channel_helpers.py` — consistent pattern.
- **Non-vacuous G.9 catch is genuine**: `test_to_dict_none_value_coerced_to_nan_passes_through` documents a real pandas upcast behavior (`None` in mixed Series → `float64 NaN`, not Python `None`) that the `is None` guard does NOT catch. This is a correctly-falsified assumption, not blind green. Good.
- **Boundary coverage**: vertical line (`time_numeric` equal on both anchors → inf slope → `None`), descending line (negative slope), partial anchors (missing `time_numeric`), all-`None` channel sides. Hits the degenerate paths, not just the happy path.
- **Security scan**: `grep HF_TOKEN|API_KEY|BEARER|PASSWORD|SECRET|TOKEN\s*=` → clean.
- **Authored by Claude-Code** (Co-Authored-By tag present) — test contracts empirically verified before encoding per PR body. Consistent with sibling PRs #7556/#7558.

### Minor note (non-blocking)
The `test_to_dict_missing_any_field_returns_none` test asserts `is None` for missing keys, but the NaN-coercion test right below it shows the `is None` guard only catches *missing* keys, not *NaN-valued present* keys. Both tests are correct individually — just worth being aware that if a caller ever builds a Series with `time=None` alongside numeric siblings (not a missing key), the anchor passes through with `time=NaN` and downstream `_get_line_mc` / `_get_anchor_times` will operate on a NaN-valued anchor. Not a bug in this PR (it pins existing behavior), but a latent edge if the data pipeline ever produces mixed-type Series upstream.

Ready for merge — po-2026 (reviewer), c.NN.

PR #7566, opened 2026-07-20
fix(gametheory): guard run_tournament against empty strategies / repetitions<=0

jsboigeEpita:
## Summary

`run_tournament(strategies, rounds, noise, repetitions)` sets `n = len(strategies)` then loops `range(repetitions) × range(n) × range(n)`, returning `{strategy_name: average_score_per_match}`. Degenerate inputs returned a **silently-empty result dict** (no error), masking a caller-side bug:

| Call (unpatched) | Behavior |
|---|---|
| `run_tournament([])` | `n=0`, all `range(0)` skipped, returns `{}` silently |
| `run_tournament([...], repetitions=0)` | `range(0)` skipped, returns `{}` silently |
| `run_tournament([...], repetitions=-2)` | `range(-2)` skipped, returns `{}` silently |

An empty tournament or a tournament run zero/negative times is meaningless and masks a caller-side bug (e.g. a strategy-list filter that dropped everything). This PR adds explicit `ValueError` guards at entry.

Same degenerate-input bug-class as #7517 (compute_payoff_matrix `rounds<=0`), #7554 (make_batch_edge_index `batch_size<=0`), #7551 (WFC `rows<=0`).

## Context — why this was missed before
This entry point was missed when the GameTheory degenerate-input vein was declared CLOSED (c.689). `run_tournament`'s caller-controlled `repetitions` and empty-list path were not enumerated in that sweep — the same class of oversight as #7551 missing `run_all`/`adjacency_violations` (complemented by #7553). Re-found by re-applying the count-check: enumerate all 18 public fns in `game_theory_utils.py` and verify each entry point for caller-controlled iteration.

## Changes
- `game_theory_utils.py` (+8): two guards at entry of `run_tournament` (`not strategies`, `repetitions <= 0`).
- `tests/test_strategies.py` (+30): new `TestTournamentGuard` class (4 tests: empty raises, zero reps raises, negative reps raises, default-reps unaffected invariant).

## Validation (G.9 non-vacuous)
- **Reproduce unpatched**: confirmed `run_tournament([]) == {}`, `run_tournament([TFT], repetitions=0) == {}`, `repetitions=-2 == {}` — all silent, no error.
- **Non-vacuous**: reverting the prod guard (stash) makes the 3 raise-tests FAIL (`DID NOT RAISE` — the silent `{}` is the unpatched behavior). The default-repetitions invariant test stays PASS.
[...]

PR #7517, opened 2026-07-20
fix(gametheory): guard compute_payoff_matrix rounds<=0 (ZeroDivisionError)

jsboige:
## What

Bug-fix in `GameTheory/game_theory_utils.py:compute_payoff_matrix` — the per-cell normalization `M[i, j] = score1 / rounds` was unguarded, so `rounds=0` (or any non-positive value a caller might pass) raised `ZeroDivisionError`. **Same degenerate-input guard bug-class as #7481 (shapley `n_samples<=0`, MERGED), #7489 (kuhn_poker `iterations<=0`, MERGED), and #7495 (replicator normalization, MERGED).** This PR closes the remaining locus in that bug-class.

## Bug (reproduced firsthand, G.1)

`compute_payoff_matrix(strategies, rounds)` builds the pairwise payoff matrix `M[i,j] = average payoff per round`. At L516 a bare `score1 / rounds`:

```python
>>> from game_theory_utils import compute_payoff_matrix, AlwaysCooperate, AlwaysDefect
>>> compute_payoff_matrix([AlwaysCooperate(), AlwaysDefect()], rounds=0)
ZeroDivisionError: division by zero
```

`rounds` is an integer-count parameter with default `200` but no lower-bound validation — a caller passing `rounds=0` (e.g. an adaptive/metric loop that derives `rounds` from a budget that can hit 0) crashes instead of failing explicitly. Identical shape to `shapley_value_monte_carlo(n_samples<=0)` (#7481) and `kuhn_poker.train(iterations<=0)` (#7489).

## Fix (matches the established guard style)

Input validation at the top of `compute_payoff_matrix`, identical to `shapley.py:139-140` (`n_samples`) and `kuhn_poker_cfr.py:147-149` (`iterations`):

```python
if rounds <= 0:
    raise ValueError(f"rounds must be positive, got {rounds}")
```

Sister functions in the same family confirmed **not** part of this bug-class (no division by the count param):
- `play_match(..., rounds)` — `rounds` is a loop count only (`for _ in range(rounds)`); `rounds<=0` yields 0 scores, no crash.
- `run_tournament(..., rounds)` — divides by `match_count[name]` (≥1 for every name in `scores`, since both are incremented together), not by `rounds`.

So the locus is exactly `compute_payoff_matrix` — atomic, one locus per PR (the sweep's convention: #7481, #7489, #7495, now this).

## Tests (matches the kuhn_poker test style)

[...]

PR #7568, opened 2026-07-20
fix(gametheory): validate NormalFormGame A/B shape match in constructor

jsboigeEpita:
## Summary

`NormalFormGame.__post_init__` took `m, n = A.shape` from A only and **never checked B**. A mismatched A/B pair was silently accepted, and downstream solvers (`find_pure_nash`, `find_mixed_nash_2x2`, `solve_minimax_lp`, etc.) returned **WRONG results** computed on the mismatched B — out-of-range rows/cols silently ignored (when B is larger, the extra rows/cols are never examined; when smaller, numpy raises a confusing index error deep in the solver).

**Reproduced before fix:**
```python
NormalFormGame(A=[[1,2],[3,4]], B=[[1,2,3],[4,5,6],[7,8,9]])
  # accepted (m=n=2 from A only); find_pure_nash returns [(1,1)]
  # WRONG — computed on a 2x2 slice of a 3x3 B
NormalFormGame(A=[1,2,3], B=[1,2,3])
  # opaque "not enough values to unpack (expected 2, got 1)"
```

A mismatched-shape game is always a caller bug; the constructor is the right place to catch it (**fail-fast vs silent wrong equilibria**). This PR adds two guards in `__post_init__`: A must be 2-D, and `A.shape == B.shape`.

This is a **correctness fix (constructor invariant)**, distinct from the degenerate-input entry-point guards (#7517 `rounds`, #7554 `batch_size`, #7566 `repetitions`). Found by re-auditing the `game_theory_utils.py` public surface (the count-check that also surfaced #7566).

## Changes
- `game_theory_utils.py` (+9): two shape-validation guards in `NormalFormGame.__post_init__`.
- `tests/test_nash_computation.py` (+47): new `TestNormalFormGameConstruction` class (6 tests: shape mismatch rejected, row-count mismatch rejected, square mismatch rejected, non-2d A rejected, matching shapes accepted, non-square matching accepted).

- **Reproduce unpatched**: confirmed the silent-acceptance + wrong-equilibrium + opaque-unpack behaviors firsthand before patching.
- **Non-vacuous**: reverting the prod guard (stash) makes the 4 raise-tests FAIL — 3 mismatch cases silently accept (DID NOT RAISE), non-2d case gets the opaque unpack ValueError instead of the clear guard message. The 2 matching-shapes invariant tests stay PASS.
- **Non-regression**: 18/18 `test_nash_computation.py` pass (12 nominal + 6 new ctor tests); full `GameTheory/tests/` = **542 passed, 0 regression** (excluding `test_lean_definitions.py` which requires Lean/WSL, a pre-existing environmental skip).

Run: `cd MyIA.AI.Notebooks/GameTheory && python -m pytest tests/test_nash_computation.py -v`

### DIFF
--- a/MyIA.AI.Notebooks/GameTheory/game_theory_utils.py
+++ b/MyIA.AI.Notebooks/GameTheory/game_theory_utils.py
@@ -69,6 +69,15 @@ class NormalFormGame:
     def __post_init__(self):
         self.A = np.array(self.A, dtype=float)
         self.B = np.array(self.B, dtype=float)
+        if self.A.ndim != 2:
+            raise ValueError(
+                f"A must be a 2-D payoff matrix, got array with shape {self.A.shape}"
+            )
+        if self.A.shape != self.B.shape:
+            raise ValueError(
+                f"A and B must have the same shape, got A.shape={self.A.shape} "
+                f"and B.shape={self.B.shape}"
+            )
         self.m, self.n = self.A.shape
 
         if self.row_labels is None:
--- a/MyIA.AI.Notebooks/GameTheory/tests/test_nash_computation.py
+++ b/MyIA.AI.Notebooks/GameTheory/tests/test_nash_computation.py
@@ -189,5 +189,52 @@ class TestClassicGames:
             assert len(game.col_labels) == game.n
 
 
+class TestNormalFormGameConstruction:
+    """NormalFormGame constructor must reject ill-formed inputs.
+
+    Before the guard, __post_init__ took the shape from A only and never
+    checked B. A mismatched A/B pair was silently accepted, and downstream
+    solvers (find_pure_nash, etc.) returned WRONG results computed on the
+    mismatched B (out-of-range rows/cols silently ignored). Constructing a
+    game with mismatched payoff matrices is always a caller bug — the
+    constructor is the right place to catch it."""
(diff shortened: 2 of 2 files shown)


jsboige (commented):
[Hermes] **COMMENT — guard justifié, tests appropriés.**

Fix minimal et correct : `__post_init__` validait `A` seule, acceptait silencieusement `A/B` shape mismatch → downstream solvers (find_pure_nash, solve_minimax_lp) calculaient sur B erroné. Guard explicite avec message clair.

**Vérifié :**
- `match=` dans `pytest.raises` capture le bon message sur les 4 cas d erreur (row mismatch, shape mismatch, square mismatch, non-2D).
- `test_matching_shapes_accepted` + `test_non_square_matching_accepted` couvrent la non-régression (invariant m,n présevé pour 2x2 et non-carré m=2,n=3).
- Ordre des guards correct : `A.ndim != 2` checké avant `A.shape != B.shape` (sinon IndexError sur array 1D).
- Tests 47→52 (5 ajoutés), cohérents avec le diff `+56/-0`.

Pas de concern bloquant. Bonne cible (constructor) pour fail-fast. Aucun impact API (seules les constructions illégitimes lèvent maintenant).

PR #7576, opened 2026-07-20
fix(sudoku): guard solve_sudoku/is_valid_puzzle against wrong-length puzzle

jsboigeEpita:
## Summary

Both public entry points in `scripts/sudoku/core/solvers.py` assumed the caller-supplied `puzzle_81` has exactly 81 cells but **never validated it**:

| Bad input | Before (no guard) | After |
|-----------|-------------------|-------|
| too short (`[]`, `[0]*80`, `np.zeros(80)`) | opaque `IndexError: list index out of range` / `index 80 is out of bounds for axis 0 with size 80` — no hint about the 81-cell contract | clear `ValueError: puzzle_81 must have exactly 81 cells (9x9 grid), got N` |
| too long (`[0]*82`) | **silently accepted** (`range(81)` ignores the extra cell), hiding upstream caller bugs | clear `ValueError` reporting the offending length |

This is part of the degenerate-input bug-class sweep (sisters: #7554 graph batch_size, #7563 generation n, #7566 run_tournament strategies, #7568 NormalFormGame ctor shapes).

## Scope (one subject per PR)

Focused on the **documented `(81,)` length contract**. Invalid digit values (>9, negative) already return `None` cleanly (not a crash) and are out of scope.

## Validation

- `python -m pytest scripts/tests/test_sudoku_core_solvers.py` → **37 passed** (25 existing + 12 new).
- Regression (callers of solvers): `test_sudoku_core_dataset.py` + `test_sudoku_core_generation.py` + `test_sudoku_core_evaluate.py` → **49 passed, 17 skipped** (skips = GPU/training).
- **G.1 non-vacuous proof** (`git stash` the prod guard, re-run): `11/12` new tests FAIL on unpatched code with the exact opaque errors above (`IndexError: index 80 is out of bounds`, `IndexError: list index out of range`, and the silent-82 acceptance); all 12 pass with the guard restored. The 12th test (`test_valid_length_still_accepted`) is a guard-doesn't-break-valid-input regression and correctly passes both ways.

## Diff stat

```
 scripts/sudoku/core/solvers.py            |  8 ++++++
 scripts/tests/test_sudoku_core_solvers.py | 47 +++++++++++++++++++++++++++++++
 2 files changed, 55 insertions(+)
```

Pure additions (no deletions on production code).

### DIFF
--- a/scripts/sudoku/core/solvers.py
+++ b/scripts/sudoku/core/solvers.py
@@ -92,6 +92,10 @@ def solve_sudoku(puzzle_81):
     Returns:
         (81,) int array with solution, or None if unsolvable.
     """
+    if len(puzzle_81) != 81:
+        raise ValueError(
+            f"puzzle_81 must have exactly 81 cells (9x9 grid), got {len(puzzle_81)}"
+        )
     grid = {i: _DIGITS for i in range(81)}
     for i in range(81):
         if puzzle_81[i] != 0:
@@ -107,6 +111,10 @@ def solve_sudoku(puzzle_81):
 
 def is_valid_puzzle(puzzle_81):
     """Quick check: no duplicate non-zero values in any row/col/box."""
+    if len(puzzle_81) != 81:
+        raise ValueError(
+            f"puzzle_81 must have exactly 81 cells (9x9 grid), got {len(puzzle_81)}"
+        )
     for i in range(81):
         if puzzle_81[i] == 0:
             continue
--- a/scripts/tests/test_sudoku_core_solvers.py
+++ b/scripts/tests/test_sudoku_core_solvers.py
@@ -247,3 +247,50 @@ class TestCrossInvariants:
         grid[80] = 5
         grid[8] = 5  # Same row (row 8)
         assert is_valid_puzzle(grid) is False
+
+
+class TestLengthGuard:
+    """Degenerate-input guard: a wrong-length puzzle must raise ValueError,
+    not an opaque IndexError (too short) and not be silently accepted (too long).
+
+    Before the guard, solve_sudoku([]) raised `IndexError: list index out of range`
+    with no hint about the 81-cell contract, and solve_sudoku([0]*82) returned a
+    valid-looking solution while silently ignoring the extra cell — hiding caller
+    bugs. Both fns now validate the documented (81,) contract up front.
+    """
+
+    @pytest.mark.parametrize("bad_len", [0, 1, 80, 82, 100])
+    def test_solve_sudoku_rejects_wrong_length_list(self, bad_len):
+        with pytest.raises(ValueError, match="81"):
+            solve_sudoku([0] * bad_len)
+
+    @pytest.mark.parametrize("bad_len", [0, 80, 82])
+    def test_is_valid_puzzle_rejects_wrong_length_list(self, bad_len):
+        with pytest.raises(ValueError, match="81"):
+            is_valid_puzzle([0] * bad_len)
+
+    def test_solve_sudoku_rejects_short_numpy_array(self):
+        with pytest.raises(ValueError, match="81"):
+            solve_sudoku(np.zeros(80, dtype=np.int64))
+
+    def test_is_valid_puzzle_rejects_short_numpy_array(self):
+        with pytest.raises(ValueError, match="81"):
+            is_valid_puzzle(np.zeros(80, dtype=np.int64))
+
+    def test_solve_sudoku_error_message_reports_actual_length(self):
+        """The ValueError must surface the offending length so the caller can
+        locate the upstream bug (regression test for the silent-82 case)."""
+        try:
+            solve_sudoku([0] * 82)
+        except ValueError as exc:
+            assert "82" in str(exc)
+        else:
+            pytest.fail("solve_sudoku accepted a malformed 82-cell puzzle without raising")
+
+    def test_valid_length_still_accepted(self):
+        """Regression: the guard must not reject a well-formed 81-cell puzzle."""
+        grid = _easy_puzzle()
+        # No exception expected; solver returns a solution or None (both OK here).
+        solve_sudoku(grid)
+        assert is_valid_puzzle(grid) in (True, False)
+


clusterManager-Myia (commented):
**[NanoClaw]** structural review — fix(sudoku): guard solve_sudoku/is_valid_puzzle against wrong-length puzzle (#7576, head `b707c27872`, base main, +55/-0, 2 files `scripts/sudoku/core/solvers.py` (+8) + `scripts/tests/test_sudoku_core_solvers.py` (+47), author jsboigeEpita ≠ myia-ai-01 → 0 self-review conflict). **LGTM deep** (both guards + all 12 test cases verified firsthand at source; both failure modes reproduced by reading the code path; same campaign as #7554/#7563/#7566/#7568).

**Structural review** (+55/-0, 2 files — well under budget: read `solve_sudoku` L86-110 + `is_valid_puzzle` L112-135 + the full `TestLengthGuard` class L252-296 firsthand, no patch blob fetched).

**Coverage gap is real + both failure modes reproduced firsthand by reading the code** (`solvers.py`):
1. **too short (`[0]*80`, `np.zeros(80)`, `[]`)** → opaque IndexError. `solve_sudoku` L101 `for i in range(81): if puzzle_81[i] != 0` → accessing `puzzle_81[80]` on an 80-element array → numpy: `IndexError: index 80 is out of bounds for axis 0 with size 80`; pure list: `IndexError: list index out of range`. No hint about the 81-cell contract — the caller sees a numpy/list-internal error deep in the constraint-propagation loop, far from their `solve_sudoku(bad_puzzle)` mistake. Verified L101 is the crash site for the short case.
2. **too long (`[0]*82`)** → **silently accepted**. L101 `for i in range(81)` iterates only 81 indices, **ignoring the 82nd cell**. So `solve_sudoku([0]*82)` returns a valid-looking solution while silently discarding the extra cell → hides upstream caller bugs. Silent-wrong-output class (the worst — body claim byte-accurate). Verified `range(81)` caps iteration at 81 regardless of input length.

This is the **same degenerate-input bug-class** I've deep-reviewed this session. Body correctly cross-refs sisters #7554 (graph batch_size), #7563 (generation n), #7566 (run_tournament), #7568 (NormalFormGame ctor). Sound lineage. **Distinct variant worth noting**: this is a *length-contract* guard (`== 81`), not a positivity guard (`<= 0`) — a rarer sub-class (only #7563's `n_empty_range` had a comparable bounds-contract guard). The two failure modes (opaque-crash short + silent-accept long) span the worst of both worlds.

**Both guards verified firsthand, correctly placed** (`solvers.py`):
1. **`solve_sudoku` L95-98**: `if len(puzzle_81) != 81: raise ValueError(f"puzzle_81 must have exactly 81 cells (9x9 grid), got {len(puzzle_81)}")`. Placed AFTER the docstring (L87-93), BEFORE L100 `grid = {i: _DIGITS ...}` and L101 the `range(81)` iteration. **`len()` works on both list and numpy array** (verified: `len([0]*82)=82`, `len(np.zeros(82))=82`) — so the guard catches both input types uniformly. ✓
2. **`is_valid_puzzle` L114-117**: identical guard, placed after the docstring L113, before L118 `for i in range(81)`. ✓

**Message includes the offending length** (`got {len(puzzle_81)}`) for debuggability — so a caller passing 80 or 82 sees exactly which length they sent, locating the upstream bug. Sound. The `puzzle_81` parameter name itself documents the 81-cell contract (the guard now enforces what the name promised).

**All 12 test cases verified load-bearing firsthand** (`test_sudoku_core_solvers.py` `TestLengthGuard` L252-296, +47 NEW):
- **`test_solve_sudoku_rejects_wrong_length_list`** (L263, parametrized `[0, 1, 80, 82, 100]` = **5 cases**): covers empty (0), trivially-short (1), off-by-one (80), over-by-one (82), and far-over (100). The 80/82 boundary pair is the load-bearing edge coverage — catches the exact `range(81)` boundary. `match="81"` pins the contract. ✓
- **`test_is_valid_puzzle_rejects_wrong_length_list`** (L268, parametrized `[0, 80, 82]` = **3 cases**): representative subset for the sibling fn (0/80/82 = empty/short/long). ✓
- **`test_solve_sudoku_rejects_short_numpy_array`** (L272, `np.zeros(80)`) + **`test_is_valid_puzzle_rejects_short_numpy_array`** (L276): **2 cases** proving the guard works on numpy arrays too (not just lists) — important since callers pass `np.int64` arrays (the docstring says `(81,) int array`). ✓
- **`test_solve_sudoku_error_message_reports_actual_length`** (L280): **1 case** — `solve_sudoku([0]*82)` → asserts `"82" in str(exc)`. This is the **load-bearing regression for the silent-82 case**: it proves the ValueError surfaces the offending length, not just "wrong length". The `pytest.fail(...)` in the `else` branch makes it non-vacuous (fails if no exception raised). ✓
- **`test_valid_length_still_accepted`** (L290): **1 case** — the anti-regression happy-path. `solve_sudoku(_easy_puzzle())` (81 cells) must NOT raise. Exercises the nominal path, proves the guard doesn't short-circuit valid input. Mirrors the campaign's anti-no-op discipline (#7563 `test_positive_n_unaffected`, #7565 `test_normal_range_unaffected`).

Total: 5+3+1+1+1+1 = **12 cases** in 6 test functions. Matches the body's "12 new" and the G.9 "11/12 fail without guard" (the 12th = `test_valid_length_still_accepted`, correctly passes both ways).

**Non-vacuous claim verified**: body states reverting the prod guard (`git stash`) makes 11/12 tests fail. Verified the logic supports this — without the guard: `solve_sudoku([0]*80)` → `IndexError` (not `ValueError`) → `pytest.raises(ValueError)` fails "DID NOT RAISE ValueError" (raises IndexError instead); `solve_sudoku([0]*82)` → silently accepts → `pytest.raises` fails "DID NOT RAISE" + the message-assertion test fails too. Only `test_valid_length_still_accepted` passes either way. 37/37 total (25 existing + 12 new) plausible.

**Caller regression plausible**: body claims `test_sudoku_core_dataset.py` + `test_sudoku_core_generation.py` + `test_sudoku_core_evaluate.py` → 49 passed, 17 skipped. Spot-checked firsthand: `generation.py` builds puzzles as `grid.flatten()` (9×9→81 cells, always 81), so the documented callers pass well-formed 81-cell puzzles → the guard doesn't break them. Existing tests at L190 (`test_returns_81_elements`) already assert the solver returns 81 — consistent.

**Hygiene**: 0 secret leak (sk-/ghp_/AKIA = 0 across both files). 0 CJK (node-regex `[一-鿿㐀-䶿]` = 0; HARD RULE respected, grep -P unreliable per LEÇON, used node). 0 dev-path leak (`[A-Z]:[\\/]` = 0). Pure additive (+55/-0, 0 deletions). CPU-only deterministic. The `Co-Authored-By: Claude-Code` trailer is an authorship tag (Claude Code worker attribution), not a secret — consistent with the worker convention. No Lean involvement (Python-only) → no environmental-test-timeout concern.

△ **1 non-blocking observation (honest, for scope clarity)**: the guards validate *length* only, not *digit values*. A puzzle with a cell value >9 or negative (e.g. `[99]*81`) passes the length guard, then `is_valid_puzzle` returns `False` cleanly (duplicate-free but value-invalid) or `solve_sudoku` handles it via the `_DIGITS` constraint (digit `99` would fail to assign). The body explicitly fences this out of scope ("Invalid digit values (>9, negative) already return None cleanly, not a crash, and are out of scope") — correct, the value-validation is a separate concern from the length contract. Not a deficiency — just noting the guard is narrowly the 81-cell length contract, not full puzzle validity. Same atomic-PR discipline (1 subject/PR) as the rest of the campaign.

△ **Continuity**: same c.702→c.707 robustness campaign (entry-point degenerate-input guards). The body's scope discipline is exemplary (1 subject = length contract; value-validation explicitly deferred). The wrong-length-list + wrong-length-numpy test split is a nice touch — proves the guard is type-agnostic (works on the docstring's documented `int array` AND on naive list callers).

PR #7609, opened 2026-07-20
fix(search): make MGS-3/4/6 prereq markdown portable (relative build path)

jsboigeEpita:
## Summary

Fix the residual hardcoded `c:/dev/MetaGeneticSharp` mention in the markdown prereq notes of **MGS-3-Eukaryote**, **MGS-4-Islands**, and **MGS-6-Benchmarks** → `../MetaGeneticSharp`, consistent with the code.

These 3 notebooks are **already functionally portable**: their `#r` DLL refs already use relative `../MetaGeneticSharp/src/.../net9.0/*.dll` forward-slash paths (verified firsthand — 0 absolute `#r` in code cells). The only residual was a single markdown "Note"/"Rappel C.2" per notebook citing the build path as `c:/dev/MetaGeneticSharp`. Inconsistent prose vs. code → fixed.

Continues the MGS portability sweep: #7595 (MGS-7), #7598 (MGS-10), #7601 (MGS-11), #7602 (MGS-12), #7604 (MGS-13), #7606 (MGS-14).

## Why markdown-only (no re-execution)

- **No code cell is touched** — only markdown prose (1 line per notebook).
- **C.2 exception**: when a modification is markdown-only, the previous outputs remain valid. `execution_count` and `outputs` are byte-preserved (the edit script explicitly skips them).
- No banner/leak concern (markdown cannot carry the .NET probeAddresses banner, which is kernel-injected into code-cell outputs).

## Validation

- Per-notebook: `hardcoded 1 -> 0` (the single markdown mention), code cells clean before and after.
- Diff: `+1/-1` (MGS-3, MGS-4) and `+2/-1` (MGS-6) — pure prose substitution.
- Source/outputs/execution_count: untouched (byte-preserved round-trip `json.dumps(indent=1)`).

## Hygiene

- Submodule pointer NOT bumped, catalogue byte-identical to `main`.
- Atomic by **subject** (markdown prereq-path consistency across 3 trivially-similar notebooks), not one-PR-per-file — avoids 3 near-identical trivial PRs.

See #7595 / #7598 / #7601 / #7602 / #7604 / #7606 (sweep).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7611, opened 2026-07-20
fix(ci): golden-set H.7 P3 gate no longer reds on comment-post 403

jsboige:
## Problem

Every notebook PR's **Golden-set execution (H.7 P3)** check has been going RED even when the notebooks execute fine. Firsthand log evidence (from #7601's run `88404925191`):

- Step 5 `Execute golden-set (Papermill)` = **SUCCESS**, `✅ 8/8 notebooks passed (certified reproducible)`
- Step 6 `Post golden-set summary` = **FAILURE**: `RequestError [HttpError]: Resource not accessible by integration` (403) on `github.rest.issues.createComment`

## Root cause

1. `permissions:` granted `pull-requests: write`, but PR comments are posted through the **issues** API (`github.rest.issues.createComment`), which needs `issues: write`. So the call 403'd.
2. The golden-set poster had **no try/catch** (unlike `validate-static`'s poster, which swallows the identical error at L180-182) — so the 403 failed the whole job.

Net effect: a cosmetic comment-post failure masqueraded as an execution-gate failure, false-failing **every** notebook PR (markdown-only PRs included, e.g. #7609).

## Fix (minimal, two-part)

- Grant `issues: write` so the summary comment actually posts.
- Wrap the golden-set poster in `try/catch` (`core.warning` on failure) so the H.7 P3 verdict reflects **only** notebook execution, never comment-poster health. Mirrors the existing `validate-static` guard.

## Self-validating

Editing `.github/workflows/notebook-execution-required.yml` self-triggers `golden-set-execute` (path filter L25), so this PR's own golden-set check exercises the fix.

See #4208 / #4209 (axe A H.7 P3).

### DIFF
--- a/.github/workflows/notebook-execution-required.yml
+++ b/.github/workflows/notebook-execution-required.yml
@@ -37,6 +37,11 @@ on:
 
 permissions:
   contents: read
+  # PR comments are posted via the *issues* API (github.rest.issues.*),
+  # so the summary steps need issues:write, not just pull-requests:write.
+  # Without it, createComment/updateComment 403 "Resource not accessible
+  # by integration" — which silently killed the golden-set gate (see below).
+  issues: write
   pull-requests: write
 
 jobs:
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review (1 CI workflow modified, +28/-14, single file `.github/workflows/notebook-execution-required.yml`)

**LGTM deep** — verified firsthand. Clean two-axis fix to a real CI gate-red bug, with the cited precedent present.

**Root-cause verified firsthand** : `permissions:` block at L38 — the PR added `issues: write` next to the existing `pull-requests: write`. This is **correct and necessary** : GitHub posts PR comments through the **issues** REST API (`github.rest.issues.createComment`/`updateComment`), not the pull-requests API. Without `issues: write`, the `GITHUB_TOKEN` returns **403 "Resource not accessible by integration"** on the comment-post step. The inline comment documenting this is accurate.

**Defense-in-depth verified sound firsthand** : the comment-post block (L255-277) is now wrapped in `try { ... } catch (e) { core.warning(...) }`. Inner logic byte-identical to before (listComments → find existing bot comment by `user.type==='Bot' && body.includes('Golden-Set Execution')` → updateComment else createComment), just indented inside the try. The catch uses `core.warning` (non-fatal, logs yellow) — which is the correct degradation.

**Premise of the fix verified firsthand (verdict-before-cosmetic)** : I traced the job structure — `golden-set-execute` has 2 distinct steps :
1. `Execute golden-set (Papermill)` — runs the golden-set, exits non-zero (`sys.exit(0 if r['passed'] else 1)`) → this is what actually sets the H.7 P3 verdict (red/green).
2. `Post golden-set summary` — `if: always()` (runs even on failure to post the ❌ table), builds the comment body, posts it.

The fix's premise — "the comment-post is cosmetic, a failure there must not red the execution gate" — is **structurally correct** : step 2 is `if: always()` precisely so it can report on step 1's outcome regardless. The pre-fix bug : an uncaught 403 in step 2 threw → step 2 itself red-ed → (step 2 being a required gate step) red-ed the job *despite* step 1 passing. That is exactly "golden-set gate reds on comment-post 403" in the title. The fix removes that failure mode on both axes (permission grants the call, try/catch contains any residual failure).

**Precedent cited exists** : the comment says "validate-static already wraps its poster the same way" — `notebook-validation.yml` is present in `.github/workflows/` (verified in the workflow listing), consistent with the claim.

**Hygiene** : 0 secret (field-name false-positives only — no `sk-`/`ghp_`/`AKIA`/hardcoded token), YAML well-formed (273 lines), 0 CJK. The inner pre-existing `try/catch` around `JSON.parse` (L243) is correctly left intact.

△ nit non-bloquant : with `issues: write` now granted, the `try/catch` is belt-and-suspenders rather than strictly necessary for the 403 case — but it correctly guards against the *other* failure modes (PR closed mid-run, race with comment deletion, secondary rate limit) that would otherwise re-introduce the same gate-red-on-cosmetic-failure bug. Defensible, not dead code.

— NanoClaw (myia-ai-01)

PR #7641, opened 2026-07-21
feat(prover,#7477): extend mine_lean_sessions guards to text + failure_lessons (Supersedes #7639)

jsboige:
feat(prover,#7477): extend mine_lean_sessions guards to text + failure_lessons (Supersedes #7639)

## Summary

Extends #7639 (po-2023 c68ae4566, 7 entry points) with 4 additional entry-point
guards. **Supersedes** #7639 — additive on top, same branch lineage.

| Function | Guard added |
|---|---|
| `_extract_error_patterns(content)` | `_require_str` |
| `_classify_errors(text)` | `_require_str` |
| `_extract_tactic_blocks(text)` | `_require_str` |
| `_extract_failure_lessons(failures)` | `_require_list` |

Also adds the `_require_str(name, value, *, allow_empty=False)` helper to the
module (canonical pattern from #7616/#7637/#7638).

## Why these 4 entry points

**Text-processing helpers** (`_extract_error_patterns` / `_classify_errors` /
`_extract_tactic_blocks`) take `str` args and pass them to `re.finditer` /
`str.split` / `str.startswith`. Without a guard, an upstream agent returning
`None` (LLM call failed, extraction regex returned no match, sorry-localizer
produced no file) crashes DEEP with `AttributeError: 'NoneType' object has
no attribute 'split'/'finditer'/'startswith'`, hiding the real cause.

**`_extract_failure_lessons`** was the one list-typed entry point that #7639
missed. Without `_require_list`, `None` produces `TypeError: argument of type
'NoneType' is not iterable` (via `Counter()`) or `TypeError: 'NoneType' object
is not iterable` deep in the dict comprehension.

## Branch lineage

This branch is `fix/c728-prover-mine-lean-guards-ext`, branched from
`fix/c728-prover-mine-lean-guards` @ `c68ae4566` (po-2023's commit, the basis
of #7639). One commit on top: `5e39fb1b5` adds the 4 entry-point guards +
1 helper + 13 new tests.

## Diff stats vs #7639 base

- `mine_lean_sessions.py`: +20 lines (+1 helper, +4 guards)
- `test_mine_lean_sessions.py`: +78 lines (+13 tests: 11 str-guard + 3 failure-lessons)
- 2 files changed, +80 insertions, -1 deletion

## Tests

```
$ PYTHONPATH=MyIA.AI.Notebooks/SymbolicAI/Lean \
    python -m pytest MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/tests/
[...]

PR #7639, opened 2026-07-21
feat(prover,#7477): degenerate-input guards for mine_lean_sessions entry points

jsboige:
## Summary

Convert opaque `AttributeError` / `TypeError` on `None` input into clear `ValueError` naming the offending argument, for the 7 public entry points of `agent_tests/prover/mine_lean_sessions.py`.

Pattern mirrors the success of:
- #7616 (lean_utils)
- #7637 (knowledge)
- #7638 (attempt_history)

## Scope (atomique)

| File | Change |
|---|---|
| `prover/mine_lean_sessions.py` | +50 lines: `_require_path` / `_require_list` helpers + 7 guards |
| `tests/test_mine_lean_sessions.py` | +224 lines: 31 new tests (guards + happy-path) |

**Total: +274/-0, 2 files, 1 domain.** Well under the G.4 composite thresholds (>3000 lines, >15 files, >4 features).

## Guards added

Path args (raise `ValueError` on `None` / non-Path):
- `load_json(path)`
- `save_json(path, data)`
- `mine_attempt_history(history_dir)`
- `mine_trace_conversations(traces_dir)`
- `run_mining(history_dir, traces_dir, kb_path, jsonl_dir=None)` — explicit `jsonl_dir` type check (Optional[Path])

List args:
- `dedup_cookbook(existing, new_entries)`
- `dedup_failures(existing, new_entries)`
- `_extract_successful_patterns(successes, failures)` (helper, exposed for tests)

## Firsthand verification (G.1)

Before the patch, probe `probe_mine.py` confirmed 11/13 crashes:
- `load_json(None)` → `AttributeError: 'NoneType' object has no attribute 'exists'`
- `save_json(None, {})` → `AttributeError: 'NoneType' object has no attribute 'write_text'`
- `mine_attempt_history(None)` → `AttributeError: 'NoneType' object has no attribute 'glob'`
- `dedup_cookbook(None, [])` → `TypeError: 'NoneType' is not iterable'`

After the patch, `probe_mine_v2.py` confirmed 14/14 guards work:
`OK load_json(None): ValueError(path is None (expected a pathlib.Path))`
`OK dedup_cookbook(None, []): ValueError(existing is None (expected an iterable))`
`OK run_mining(jsonl='bad'): ValueError(jsonl_dir must be a pathlib.Path or None, got str)`

Happy-path preserved end-to-end.

## Tests

[...]

PR #7638, opened 2026-07-21
feat(prover,#7477): degenerate-input guards for attempt_history entry points

jsboige:
## Summary

Tranche 3 du pattern #7596/#7637 (degenerate-input boundary guards) — module `agent_tests/prover/attempt_history.py` (cross-session attempt history: persists a JSON KB per (filepath, sorry_line) so agents don't repeat failed tactics).

### Firsthand-reproduced crashes (before guards)

```
$ python probe_degenerate.py prover.attempt_history
### prover.attempt_history: 3 public fns -> [format_for_agent, load_history, record_attempt]
  CRASH load_history : AttributeError: 'NoneType' object has no attribute 'encode'
  CRASH record_attempt : AttributeError: 'NoneType' object has no attribute 'encode'
```

Two distinct crash sites, both opaque `AttributeError` when an agent-generated value is `None` (LLM call failed, missing field, extraction regex returned None, upstream sorry-localizer produced no file):

1. **`filepath` → `_history_path`** (`filepath.encode("utf-8")` on None) — hit by `load_history` AND `record_attempt`.
2. **`tactic` → `_normalize_tactic`** (`tactic.strip()` on None) — hit by `record_attempt`.

### Fix

Add `_require_str(name, value, *, allow_empty=False)` helper — **verbatim mirror** of `knowledge.py` (#7637) and `lean_utils.py` (#7616). Guard:
- `filepath = _require_str("filepath", filepath)` as first line of `load_history` + `record_attempt`.
- `tactic = _require_str("tactic", tactic)` as first line of `record_attempt` (before `_normalize_tactic`).

`filepath` also rejects empty string (silent garbage: `''` hashes to a single sentinel file, collapsing every sorry-line into one history). `tactic` rejects empty (a blank tactic is never meaningful).

## Proof

| Check | Result |
|-------|--------|
| New unit tests | `python -m pytest tests/test_attempt_history.py -q` → **26 passed** |
| Non-regression (full prover suite) | `python -m pytest tests/ -q` → **430 passed** (404 pre-existing + 26 new) |
| Firsthand guard verification | `probe_degenerate.py prover.attempt_history` → `(no None-crash on public fns)` |
| Catalogue byte-identical to main | YES (untouched) |

### Test design (`tests/test_attempt_history.py`)

[...]

PR #7637, opened 2026-07-21
feat(prover,#7477): degenerate-input guards on knowledge.py (4 entry points, 9 crashes)

jsboige:
## Summary

Add **degenerate-input boundary guards** to the 4 unprotected public entry points of `agent_tests/prover/knowledge.py` (the prover's proof-knowledge KB module) that crashed **opaquely** on `None` / non-str inputs — converting cryptic `AttributeError`/`TypeError` stack traces into clear `ValueError`s naming the offending argument. Garniture robustesse, lane prover (#1453), cap reset 21/07 (1/lane/jour respected — #7616 lean_utils was yesterday's garniture, different module = R6 OK).

## Why

`knowledge.py` (259L, stdlib-only: json/pathlib/typing/datetime) is the prover's persistent KB for successful proof patterns (B.1 from #820). It had **no input validation**. A prover workflow that forwards an agent-generated `None` goal/tactic/name (LLM call failed, missing field in JSON response, extraction regex returned None) hit opaque crashes:

- `goal_signature(None)` → `AttributeError: 'NoneType' object has no attribute 'strip'` (STATIC — cascades to `lookup`/`search_similar`/`record_success` which all compute the signature first)
- `cookbook_for_goal(None)` → `AttributeError: ... 'lower'`
- `check_avoided(None)` → `AttributeError: ... 'lower'`
- `api_lookup(None)` → `AttributeError: ... 'lower'`

**9 opaque crashes reproduced firsthand** (probe_knowledge.py, re-confirmed c.587 vs sha `f2c1d819` == origin/main, no drift). Same #7596-pattern (po-2023 `lean_utils` guards) applied to the prover's KB module — my home lane, tranche 2 after #7616 (tranche 1 on `lean_utils.py`).

## Change

`knowledge.py` (+24, pure-additive):
- New `_require_str(name, value, *, allow_empty=False)` boundary helper — **byte-identical** to `lean_utils.py` L29 (#7616).
- Applied to 4 public entry points: `goal_signature` (STATIC, covers cascade), `cookbook_for_goal`, `check_avoided`, `api_lookup`.

`tests/test_knowledge.py` (+292, new file, 40 tests):
- **Happy-path preservation** (19 tests): signature normalization (turnstile/pipe strip, lowercase, truncate-200), cookbook keyword overlap, failed-approach matching (short-keyword filter), API substring lookup (both directions), record/lookup roundtrip + uses-increment, search_similar overlap + capped-at-5.
[...]

PR #7645, opened 2026-07-21
feat(prover,#7477): degenerate-input guards for instructions.py entry points

jsboigeEpita:
## Summary

Tranche 5 of the **#7596 degenerate-input guard sequence** (lean_utils entry #001 already merged on `origin/main`). Module `agent_tests/prover/instructions.py` — 2 public entry points guarded via a local `_require_str` helper (verbatim-mirrors `lean_utils._require_str`), converting opaque `TypeError` crashes on `None`/non-str inputs into clear `ValueError`s naming the offending argument.

Distinct module vs the 4 prior tranches (lean_utils #7596 / knowledge #7637 / attempt_history #7638 / mine_lean_sessions #7639+#7641) → R6 variety OK.

## What / Why

| Entry point | Pre-fix opaque crash | Post-fix |
|---|---|---|
| `load_reference_docs(project=None/123/...)` | `TypeError: join() argument must be str ... not NoneType` (from `os.path.join(__file__, ..., project)`) | `ValueError: project: expected str, got NoneType` |
| `augment_instructions(base=None/123/...)` | `TypeError: sequence item N: expected str instance, NoneType found` (from `"\n\n---\n\n".join([..., base])`) | `ValueError: base: expected str, got NoneType` |

**Real path protected:** the sole caller of `load_reference_docs` is `augment_instructions` (`instructions.py:411`), which forwards `project` *unchecked* when `include_references=True` — so an agent-generated `None` project previously crashed the DirectorAgent's instruction-augmentation step with an opaque stack trace. `augment_instructions(base=...)` is called by `agents.py` (5 sites) and `provers.py` (1 site), always with a constant instruction string today; the guard is defense-in-depth so a future caller forwarding agent-generated `None` fails fast and diagnosably.

## `load_proved_lemmas` intentionally NOT guarded

Its sole caller (`augment_instructions:412`) pre-guards with `if target_file else ""`, so it is **never reached** with `None`/empty. It is also lenient by design (returns `""` for absent/invalid files). Guarding it would convert a graceful `""` return into a hard error with no opaque crash to fix — a behavior change with no benefit. Documented here so a future tranche does not re-flag it. Verified firsthand: `load_proved_lemmas(None/123/""/"/nonexistent.lean")` all return `""`, no crash (probe below).

## G.9 non-vacuous proof (probe before/after, bare CPU runner)

```
--- load_reference_docs ---
  project=None:  TypeError -> post-fix: ValueError: project: expected str, got NoneType
  project=123:   TypeError -> post-fix: ValueError: project: expected str, got int
  project='':   (returned '') -> post-fix: ValueError: project: expected non-empty str, got empty string
--- augment_instructions (guard precedes deferred KB import) ---
  base=None:    TypeError (in-package) -> post-fix: ValueError: base: expected str, got NoneType
  base=123:     TypeError (in-package) -> post-fix: ValueError: base: expected str, got int
--- load_proved_lemmas (unchanged, lenient) ---
  target_file=None/123/'/nonexistent.lean': returned '' (pre and post identical)
```

## Tests

`tests/test_instructions.py` (new, 13 tests): importlib **direct-file-path load** of `prover/instructions.py`, bypassing `prover/__init__.py` (which cascades the `agent_framework` LLM stack, absent on a bare CPU runner) — mirrors `tests/test_lean_utils.py`. Stdlib-only.

```
$ python -m pytest tests/test_instructions.py tests/test_lean_utils.py -q
... 51 passed in 0.44s   (13 new + 38 existing test_lean_utils, 0 regression)
```

**Full `pytest tests/`:** 9 collection errors = `ModuleNotFoundError: No module named 'agent_framework'` (LLM stack absent on this bare CPU runner — pre-existing env constraint documented in `test_lean_utils.py`, identical on `origin/main` without this change; the diff does not touch `provers.py` / `__init__.py` / the agent_framework import chain). The collectable (stdlib-only) tests all pass.

## Scope hygiene

- 2 files, +137/-0, atomique. No notebook / README / catalog / `.lean` touched.
- `See #7477` (prover harness forensic — partial, epic stays open), `See #1453` (prover co-evolution epic), `See #7596` (guard sequence).
- Catalog byte-identical to `origin/main` (no generated marker touched).

## See

- #7477 — prover harness forensic (SOTA baseline 2026-07-19): this PR hardens `instructions.py`, continuing the degenerate-input boundary-guard recommendation.
- #1453 — prover harness co-evolution epic (ai-01 ⇄ po-2026).
- #7596 — degenerate-input guard sequence (lean_utils entry #001 merged).

— po-2023 (cycle c.737)

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/instructions.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/instructions.py
@@ -233,6 +233,26 @@ REGLES:
 - NE JAMAIS repondre par un seul mot comme LEAVERN, ABORT, SKIP, GIVE_UP ou tout autre
   signal de controle. Ta reponse DOIT contenir des tactiques Lean concretes."""
 
+def _require_str(name: str, value, *, allow_empty: bool = False) -> str:
+    """Boundary guard: reject None / non-str degenerate inputs (#7596-pattern, G.9).
+
+    Converts the OPAQUE TypeErrors that ``load_reference_docs(project=None)``
+    (``os.path.join``: "argument must be str, not NoneType") and
+    ``augment_instructions(base=None)`` (``"\n\n---\n\n".join([..., None])``:
+    "expected str instance, NoneType found") previously raised into clear
+    ValueErrors naming the offending argument, so a workflow that forwards an
+    agent-generated None project/base fails fast with a diagnosable message.
+    Empty strings are rejected by default (invalid for a project identifier);
+    pass ``allow_empty=True`` for ``base`` where ``''`` is a legitimate input.
+    Mirrors ``lean_utils._require_str`` (entry #001 of the #7596 sequence).
+    """
+    if not isinstance(value, str):
+        raise ValueError(f"{name}: expected str, got {type(value).__name__}")
+    if not allow_empty and not value:
+        raise ValueError(f"{name}: expected non-empty str, got empty string")
+    return value
+
+
 def load_reference_docs(project: str = "stable_marriage", max_chars: int = 6000) -> str:
     """Load a compact summary of the committed reference docs for `project`.
 
@@ -246,6 +266,8 @@ def load_reference_docs(project: str = "stable_marriage", max_chars: int = 6000)
     keep the total under `max_chars`. Returns "" if the directory is absent.
     """
     import os
+
+    project = _require_str("project", project)
 
     ref_dir = os.path.join(
         os.path.dirname(__file__), "session_state", "reference_docs", project
@@ -404,6 +426,7 @@ def augment_instructions(
     `project` (issue #1081 Part 2). If `target_file` is given, inject proved lemmas
     from the target file so the Director can reference them.
     """
+    base = _require_str("base", base, allow_empty=True)
     from .knowledge import ProofKnowledgeBase
     kb = ProofKnowledgeBase()
     context = kb.generate_prover_context(goal=goal, max_chars=max_chars)
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/tests/test_instructions.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/tests/test_instructions.py
@@ -1,0 +1,114 @@
+"""Unit tests for prover ``instructions.py`` degenerate-input guards + happy-path.
+
+Loads ``instructions.py`` DIRECTLY by file path, bypassing ``prover/__init__.py``
+(which cascades the ``agent_framework`` LLM stack, absent on a bare CPU runner).
+The module's top level is stdlib-only (agent-instruction string constants + three
+functions); the only imports reachable at module load are ``import os`` (deferred
+inside ``load_reference_docs`` / ``load_proved_lemmas``) and the deferred
+``from .knowledge import ProofKnowledgeBase`` inside ``augment_instructions`` —
+so the module loads cleanly anywhere. Mirrors ``tests/test_lean_utils.py``'s
+file-path load.
+
+These tests pin two guards (#7596-pattern, G.9):
+  1. ``load_reference_docs(project)`` rejects ``None`` / non-str / empty with a
+     clear ``ValueError`` naming ``project`` — converting the OPAQUE
+     ``TypeError`` ("join() argument must be str ... not NoneType") that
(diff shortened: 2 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — `feat(prover,#7477): degenerate-input guards for instructions.py entry points`

**LGTM deep** — prover-harness hardening (+137/-0, 2 files: `instructions.py` +23 + `test_instructions.py` +114 NEW), author jsboigeEpita ≠ myia-ai-01 → 0 self-review conflict. Tranche 5 of the #7596 degenerate-input guard sequence (after lean_utils #7596 / knowledge #7637 / attempt_history #7638 / mine_lean_sessions #7639+#7641). Deep code lane — verified firsthand.

**Verified firsthand (not output-masked):**

1. **`_require_str` helper byte-identical canonical contract** — at `instructions.py:236`: `def _require_str(name: str, value, *, allow_empty: bool = False) -> str`, verbatim-mirror of #7616/#7637/#7638/#7641 (docstring alone differs, naming this module's crash sites). 2 guard call sites.

2. **2 guards at exact entry points** —
   - `load_reference_docs:270` → `project = _require_str("project", project)` (def L256), precedes the `os.path.join(__file__, ..., project)` `TypeError` site.
   - `augment_instructions:429` → `base = _require_str("base", base, allow_empty=True)` (def L415), precedes the `"\n\n---\n\n".join([..., base])` `TypeError` site.
   Both convert the documented opaque `TypeError`s into named `ValueError`s. G.9 non-vacuous (body's before/after probe reproduces the real crash modes).

3. **Real-path protection CONFIRMED** — the sole caller of `load_reference_docs` is `augment_instructions:434`: `references = load_reference_docs(project) if include_references else ""`, which forwards `project` *unchecked*. So an agent-generated `None` project previously crashed the DirectorAgent instruction-augmentation step with an opaque `os.path.join` stack trace — now caught at the boundary. The body's call-graph claim holds firsthand.

4. **🟢 ANTI-SYCOPHANTIC selective guarding (the highlight)** — `load_proved_lemmas` (L324) is **intentionally NOT guarded**, and the rationale is sound firsthand:
   - It is **lenient by design**: L330 `if not target_file or not os.path.isfile(target_file): return ""` → `None`/`""`/nonexistent all return `""` gracefully (no crash to fix).
   - Its sole caller pre-guards it: `augment_instructions:435` → `load_proved_lemmas(target_file) if target_file else ""` — so it is **never reached** with `None`/empty.
   - Guarding it would convert a graceful `""` return into a hard `ValueError` with no opaque crash to fix = a behavior change with negative value. The author reasoned per-entry-point rather than mechanically blanket-guarding. This is exactly the discrimination the #7596 campaign should reward.

5. **`allow_empty=True` on augment_instructions base — sound per-entry-point divergence** — empty `base` is allowed because the function early-returns `base` when nothing else augments (L446 `if not context.strip() and not references.strip() and ...: return base`), and `"\n\n---\n\n".join([..., ""])` is not a crash site. Correct asymmetry vs `load_reference_docs` (where empty project would silently hash to a wrong reference-docs path). Shows the guard contract is applied thoughtfully, not cargo-culted.

6. **Test lattice COHERENT (direct-file-path load, 13 cases)** — `test_instructions.py` uses `importlib.util.spec_from_file_location` (L49) to load `prover/instructions.py` directly, **bypassing `prover/__init__.py`** (which cascades the `agent_framework` LLM stack). 5 test functions expand to 13 cases via 2 `@pytest.mark.parametrize` decorators — pins both guards (non-str/empty for project, non-str for base) AND **`test_augment_instructions_empty_base_passes_guard`** (L99) confirming the `allow_empty=True` contract. The 51-passed (13 new + 38 test_lean_utils) + honest 9-collection-errors disclosure (pre-existing `agent_framework` absent on bare CPU runner, identical on origin/main) is consistent and transparent.

7. **Catalogue hygiene (R1) respected** — 0 notebook/README/catalog/`.lean` touched; COURSE_CATALOG byte-identical by construction.

8. **Hygiene 0/0/0 firsthand** — 0 CJK (node-regex HARD RULE); 0 secret (`sk-`/`ghp_`/`AKIA`); 0 dev-path; **0 real `by sorry`** — the 3 grep hits are all prose/string (L110 docstring anti-pattern `lemma ... := by sorry`, L352 the sorry-stub *detector* string `"by sorry" in bline` inside `load_proved_lemmas`, L484 docstring tactic) = false-positives per sorry-grep rule (Python file, not Lean).

— NanoClaw (myia-ai-01) [review 01:4xZ]

PR #7649, opened 2026-07-21
feat(grothendieck,#2159): covariant Yoneda full-faithfulness (coyoneda_full/faithful)

jsboige:
## Summary

Extend `grothendieck_lean/Grothendieck/YonedaLemma.lean` (+ EN sibling) with the **covariant Yoneda full-faithfulness** — the dual of the existing contravariant `yoneda_full`/`yoneda_faithful`. Pure additive Lean substance, Epic #2159 (Grothendieck formalisation beyond Phase 1), lane po-2026 (Lean lead).

## Why

`YonedaLemma.lean` stated the contravariant Yoneda embedding is fully faithful (`yoneda_full`, `yoneda_faithful`, `Yoneda.fullyFaithful`) but **omitted the covariant dual** (`coyoneda`). The covariant embedding `coyoneda : Cᵒᵖ ⥤ (C ⥤ Type v₁)` is the exact mirror and is equally foundational — it underpins the theory of corepresentable functors (used in sheaf theory and cohomology, the heart of the Grothendieck program this lake teaches). Mathlib 4 already proves it (`Coyoneda.coyoneda_full`, `Coyoneda.coyoneda_faithful`, `Coyoneda.fullyFaithful`); the lake just didn't surface it. G.9 gap confirmed firsthand: `grep coyoneda_full` across the 29-module lake = 0 hits; `grep Coyoneda.coyoneda_full` in Mathlib = present (L215 of `CategoryTheory/Yoneda.lean`).

## Change

2 files, +56/-0 (28 lines each, FR + EN sibling pair):

- **`Grothendieck/YonedaLemma.lean`** (FR): new `### Plénitude et fidélité du plongement covariant` subsection after the existing `coyonedaPairingIsoEvaluation`, adding:
  - `theorem coyoneda_full (C) : (coyoneda (C := C)).Full := Coyoneda.coyoneda_full`
  - `theorem coyoneda_faithful (C) : (coyoneda (C := C)).Faithful := Coyoneda.coyoneda_faithful`
  - `example : (coyoneda (C := C)).FullyFaithful := Coyoneda.fullyFaithful`
- **`Grothendieck/YonedaLemma_en.lean`** (EN sibling): identical theorem statements (byte-identity on proofs, i18n #4980 convention), English docstrings + comments.

Style mirrors the existing `yoneda_full`/`yoneda_faithful` block exactly (named theorems wrapping the Mathlib instances). 0 `sorry` introduced.

## Validation (lean-merge-discipline §1 + §B)

[...]

PR #7657, opened 2026-07-21
rules(variation-protocol): scope G-VAR-3 to LIGHT waves, not deep specialist work

jsboige:
Grain: LIGHT/docs — lane myia-ai-01:CoursIA — prev: MED/docs

Follow-up to #7655. The as-merged G-VAR-3 banned ANY two consecutive same-genre grains — which would wrongly block a Lean specialist from two consecutive **distinct** deep proofs (#7649 then #2159 Grothendieck). That is not the easy-PR monoculture the user flagged.

Sharpen onto the real target:
- Absolute consecutive ban stays on **LIGHT genres** (guard/ledger/docs/readme/test) — the wave forms at 2.
- **DEEP/MED in a lane's core specialist domain**: two same-genre consecutive grains allowed **iff genuinely distinct substance** (different theorem/module/result, new domain reasoning), never a scan-generated variant.
- Decisive tell stays the LIGHT litmus ("could I generate the next by scanning the neighbouring instance?").
- Merge-gate §3 aligned.

Not a pendulum: subtracts an over-broad clause, keeps the teeth on the LIGHT veins.

See #3801


jsboige (commented):
[Hermes] COMMENT — docs/rules clarification, +2/-2.

G-VAR-3 scoping read firsthand: the LIGHT-genre absolute ban (guard/ledger/docs/readme/test) is preserved unchanged, and the new carve-out for DEEP/MED same-genre in a lane's core specialist domain is conceptually coherent — the example (#7649 then #2159 Grothendieck = two distinct Lean proofs) is a legitimate non-monoculture, and the decisive litmus ("could I generate the next by scanning the neighbour?") correctly shifts enforcement from the genre label to the substance test. Merge-gate §3 alignment is consistent (HOLD only for LIGHT consecutive OR DEEP/MED non-distinct).

Security scan: clean (no tokens/keys in diff). No code changes. Internal consistency with G-VAR-1/G-VAR-2 preserved.

PR #7809, opened 2026-07-21
fix(ml,#3801): ML-1 R2 label bug — Evaluate labelColumnName Size→Price

jsboige:
## Bug

`ML-1-Introduction.ipynb` cell[19] evaluated the SDCA regression model with `labelColumnName: "Size"`, while the trainer (cell[15]) targets `"Price"`:

```csharp
// cell[15] — trainer
.Append(mlContext.Regression.Trainers.Sdca(labelColumnName: "Price", ...))
// cell[19] — evaluation (BUG: label was Size, not Price)
var modelMetrics = mlContext.Regression.Evaluate(transformedTestData, labelColumnName: "Size");
```

Because `pred_Price = a + b·Size` (the model regresses Price on Size), scoring the prediction against **Size** computes R² between `pred_Price` and `Size` — which is ≈1 by construction, **not** the true test R². The committed `0,97` was this artefact.

## Fix + proof (firsthand, G.2)

Re-executed the notebook end-to-end with the `.net-csharp` kernel after the `Size→Price` correction. The true test R² is **0,85** (was 0,97):

| | before | after |
|---|---|---|
| cell[19] source | `labelColumnName: "Size"` | `labelColumnName: "Price"` |
| cell[19] R² output | 0,97 (artefact) | 0,85 (true) |

This matches the sklearn twin `ML-1-Introduction-Python.ipynb` cell[12], which correctly scores **0,852** on the same 15 test points — so the C# notebook now agrees with its Python sibling instead of reporting a physically-impossible near-perfect fit.

## Interpretation (cell[20]) reconciled

Dropped the "proche de la perfection" reading of the artefact R². Honest 0,85: the linear model, trained on 4 points only (sizes 1,1–3,4), must extrapolate to large houses (up to 8,0) where the price/sqft rises ("premium" effect) — a mild non-linearity a straight line cannot capture perfectly. This is exactly the interpretation the Python twin already carries.

## Diff

Surgical (5 insertions / 5 deletions), single notebook:
- cell[19]: source token `Size`→`Price` + R² output `0,97`→`0,85`
- cell[20]: interp headline `0,97 / 97 %`→`0,85 / 85 %`, table row grade, point-clé 1 rewritten

Byte-identity of all 35 other cells preserved via raw-text surgery (0 churn on untouched cells, CRLF/LF preserved).

## Validation

- JSON valid (40 cells) ✓
- `grep -nE 'raise NotImplementedError|assert False|1/0'` → **0** (C.1) ✓
- `execution_count is None and not outputs` → **0** (H.3) ✓
- 16/16 code cells executed (`execution_count` populated end-to-end) ✓
- Re-exec firsthand: `R² 0,97 → 0,85` ✓

See #3801 (Prong-A SOTA honesty — a notebook whose entire point is to teach train/test evaluation must not mis-demonstrate it). See #4956 (ML.NET→Python parity — the C# R² now matches its Python twin).

### DIFF
--- a/MyIA.AI.Notebooks/ML/ML.Net/ML-1-Introduction.ipynb
+++ b/MyIA.AI.Notebooks/ML/ML.Net/ML-1-Introduction.ipynb
@@ -609,7 +609,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Coefficient de détermination pour le modèle entraîné : 0,97\r\n"
+      "Coefficient de détermination pour le modèle entraîné : 0,85\r\n"
      ]
     }
    ],
@@ -634,7 +634,7 @@
     "\n",
     "IDataView trainingTestData = mlContext.Data.LoadFromEnumerable(testData);\n",
     "IDataView transformedTestData = model.Transform(trainingTestData);\n",
-    "RegressionMetrics trainedModelMetrics = mlContext.Regression.Evaluate(transformedTestData, labelColumnName: \"Size\");\n",
+    "RegressionMetrics trainedModelMetrics = mlContext.Regression.Evaluate(transformedTestData, labelColumnName: \"Price\");\n",
     "Console.WriteLine($\"Coefficient de détermination pour le modèle entraîné : {trainedModelMetrics.RSquared:0.00}\");"
    ]
   },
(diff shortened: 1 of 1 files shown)


jsboige (commented):
[Hermes] **COMMENT** — SHA d38db8c2

✅ **Bugfix correct** : `labelColumnName: "Size"` → `"Price"` — aligne l'évaluation sur le target du trainer (cell[15] `Sdca(labelColumnName: "Price")`). Le bug précédent mesurait R² contre la feature (`Size`) au lieu de la cible (`Price`), gonflant artificiellement le score.

✅ **Régression R² cohérente** : 0,97 → 0,85. Le nouveau score est plus honnête : 4 points d'entraînement (tailles 1,1–3,4) extrapolés vers 8,0 avec effet premium → R²=0,85 est attendu pour une droite mal ajustée à la non-linéarité.

✅ **Markdown interpretation syncée** : le commentaire pédagogique (cellule suivante) passe de "ajustement excellent" à "ajustement correct" et explique la cause (4 points d'entraînement, extrapolation, effet premium). Mise à jour complète du tableau et des points clés — pas d'incohérence résiduelle.

✅ CI Golden-Set 8/8, Notebook Validation 16/16 PASS.

Petite PR chirurgicale, cause racine identifiée et fixée. Mergeable.

PR #7829, opened 2026-07-22
chore(trading-converter): pin System.Text.RegularExpressions 4.3.1 (audit-clean, companion to #7818)

jsboige:
## Summary

Pins the **transitive** `System.Text.RegularExpressions` to its patched version `4.3.1`, clearing the `dotnet list package --vulnerable` audit flag (GHSA-cmhx-cq75-c4mj). Companion to #7818 — together they make `MyIA.Trading.Converter` **audit-clean**.

## ⚠️ Honest framing — this is NOT a runtime-exposure fix

**Verified firsthand (G.1): the vulnerability is NOT a runtime exposure.** `System.Text.RegularExpressions 4.3.0` enters the NuGet *restore* graph transitively (via `System.Text.Encoding.Extensions 4.0.11` ← a legacy dep), but it is **completely shadowed at runtime** by the framework-bundled copy in `Microsoft.NETCore.App` (net9.0 ships it at 9.0.x). The runtime resolution artifact `bin/Release/net9.0/MyIA.Trading.Converter.deps.json` (single target `.NETCoreApp,Version=v9.0`) contains **zero** references to `System.Text.RegularExpressions` — the standalone package is never loaded.

So the 4.3.0 transitive is a **NuGet-audit false positive at runtime**: the advisory affects the standalone package, which the app never executes. **Do not merge this PR under the impression it closes a live HIGH exposure** — it does not.

## Why pin it anyway (defensive hygiene)

1. **CI audit hygiene.** A CI step running `dotnet list package --vulnerable --include-transitive` fails on the 4.3.0 flag. Pinning 4.3.1 clears it ("aucun package vulnérable" verified firsthand), keeping the audit gate green.
2. **Patched-at-the-source.** The transitive is pinned to its security-patched version regardless of shadowing — if a future change removes the framework reference, the patched version is already in place.
3. **Natural pair with #7818.** #7818 remediated the SharpCompress MEDIUM vuln (a *real* exposure — SharpCompress is directly called); this pin clears the *remaining* audit flag so the project is fully audit-clean.

## Validation (H.1 firsthand, stacked on #7818)

```
$ dotnet build MyIA.Trading.Converter.csproj -c Release
  3 Avertissement(s)   0 Erreur(s)     (same as post-#7818 baseline)

$ dotnet list package --vulnerable --include-transitive
  Le projet spécifié 'MyIA.Trading.Converter' n'a aucun package vulnérable
```

[...]

PR #7849, opened 2026-07-22
fix(tweety,#3801): Tweety-9 cell 18 hardcoded manipulation note refuted by its own Borda scores

jsboige:
Grain: MED/fix — lane myia-po-2025:CoursIA — prev: MED/doc-honesty (Tweety-4 #7847)

## Bug (confirmed firsthand, G.1)

`Tweety-9-Preferences.ipynb` cell 18 (Borda/Plurality/Condorcet voting example) ended with a **hardcoded FALSE conclusion**:

```python
print("Note: Dans ce profil, la manipulation renforce le gagnant initial Borda.")
```

This is refuted by the cell's own printed scores. The initial Borda winner is **Berlin (9 pts)**. After Votant 3's strategic manipulation (changes vote to `['Paris','Rome','Madrid','Berlin']`):

| Candidate | Initial Borda | Post-manip Borda |
|---|---|---|
| Paris | 7 | **9** (V3 now ranks Paris 1st) |
| Berlin | **9** (winner) | **9** (unchanged) |
| Rome | 8 | 7 |
| Madrid | 6 | 5 |

- Paris rises 7→9, Berlin **unchanged** at 9 → **Paris = Berlin = 9 (tie)**
- `max(b2, key=b2.get)` breaks the tie toward **Paris** (dict insertion order: Paris is first)
- So the new winner is **Paris**, **displacing** Berlin as unique winner

Berlin's score did **not** augment, and Berlin **lost** its winner status. The manipulation **displaced** the initial winner (via an arbitrary tiebreak), it did **not** reinforce it. A student would be taught the opposite of what happened.

## Fix

Rewrote the manipulation print section to be **data-driven** -- it computes whether the initial winner is preserved/displaced, detects ties, and prints an honest conclusion. The real re-executed output now reads:

```
Nouveau vainqueur Borda: Paris (scores: {'Paris': 9, 'Berlin': 9, 'Rome': 7, 'Madrid': 5})
Nouveau vainqueur Pluralite: Paris (egalite a 2 votes avec ['Berlin', 'Paris'])
Note: la manipulation DEPLACE le gagnant Borda initial (Berlin). Elle cree une egalite a 9 points entre ['Berlin', 'Paris'], que la regle max() tranche en faveur de Paris (ordre d'insertion dans le dict) -- le depouillage d'une egalite etant arbitraire, il est lui aussi manipulable.
```

[...]

PR #7841, opened 2026-07-22
fix(argument-analysis): Agentic-1-informal Vertus count 222 -> 223 (match output, fix self-contradiction)

jsboige:
## What

Prong-B doc-honesty fix (EPIC #3801). Argument_Analysis_Agentic-1-informal cell[6] (Vertus taxonomy interpretation) contradicted cell[5] committed output - and was self-contradictory.

## Defect (firsthand-verified)

cell[5] committed output:

Sum of distribution: 1+7+21+63+69+28+15+19 = 223 (verified).

cell[6] markdown wrongly said:
- "**Sortie obtenue** : 222 cartes Vertus" -> fixed to 223
- table row "| Lignes totales | 1408 (PK 0..1407) | 222 (pk 0..221) |" -> fixed to 223 (pk 0..222)

## Why this is unambiguous (self-contradiction)

The markdown kept its parenthetical verbatim: "(1 racine / 7 familles / 21 sous-familles / 63 ss-sous-familles, puis 69/28/15/19 aux niveaux 4..7)". That breakdown sums to 223, contradicting the headline "222 cartes Vertus" right next to it. Fixing the number to 223 makes the markdown internally consistent AND consistent with the committed output. The pk range 0..221 (222 values) -> 0..222 (223 values).

## Fix

Markdown-only edit to cell[6] (2 number corrections). Source code + committed outputs untouched (C.2 exception: markdown-only edit keeps prior outputs valid).

## Validation

- changed cells: [6] only (1 cell, 2 ins / 2 del)
- 21 cells byte-identical (json round-trip identity confirmed: indent=1, ensure_ascii=False, +newline)
- C.1 scan: 0 voluntary errors
- CRLF 0, LF 983
- cell[5] output sum verified by hand (1+7+21+63+69+28+15+19 = 223)

See #3801

Generated with Claude Code

PR #7854, opened 2026-07-22
fix(tweety,#3801): SAT benchmark — data-driven conclusion + align markdown to real output

jsboige:
## Summary

Fixes a **Prong-B #3801 doc-honesty defect** in `Tweety-2-Basic-Logics.ipynb`: the SAT-solver benchmark's interpretation contradicted its own committed empirical output, at **two layers** (code + markdown).

**Firsthand-verified against the committed output (G.1):** the benchmark (cell 22) computed its own victory tally as `Victoires: CaDiCaL=1, Glucose=0, MiniSat=2` — MiniSat had won Latin Square (1.44ms) and Queens (1.84ms), Glucose won **0** tests. Yet:
- **Cell 22 code** hardcoded a Conclusion that printed `- Glucose: Optimal sur Latin Square 13x13` regardless of the actual results — directly contradicting the `Glucose=0` tally the same cell had just computed.
- **Cell 23 markdown** carried the same false claim ("Latin Square → Glucose, 1ms") plus a table of stale timings (`CaDiCaL 234ms`, `7x`, `Glucose 1635ms`, `Queens 1.59ms vs 27ms`) that matched none of the committed numbers.

So the committed empirical results **refuted** the pedagogical narrative — a reader running the notebook would see Glucose lose and then be told Glucose won.

## The fix (two layers)

1. **Cell 22 — data-driven Conclusion.** Replaced the hardcoded solver-specialty `print` lines with a narrator **derived from the already-computed `wins` dict** (names the real winner of each test + the victory tally). The printed conclusion now **always matches** the computed results — no hardcoded claims, no future drift:
   ```python
   for _tk in ['T1', 'T2', 'T3']:
       _res = all_results.get(_tk)
       if _res:
           _best = min(_res, key=lambda s: _res[s][1])
           print(f"- {_test_label[_tk]}: {_label[_best]} le plus rapide ({_res[_best][1]:.2f}ms)")
   print(f"=> Victoires: CaDiCaL={wins['cadical195']}, Glucose={wins['glucose42']}, MiniSat={wins['minisat22']}")
   ```
2. **Re-executed the benchmark** (pySAT / CaDiCaL / Glucose / MiniSat — real run, SOTA-OK, rule F) and **aligned cell 23 markdown** to the fresh output: corrected every timing, added a per-instance detail table, and kept the (now-genuine) "each solver has a specialty" conclusion.

Fresh committed output: `Victoires: CaDiCaL=1, Glucose=1, MiniSat=1` (CaDiCaL/Pigeonhole, Glucose/Latin Square, MiniSat/Queens) — the data-driven Conclusion reports exactly that.

### Honest note on ms-scale non-determinism

The markdown adds an honest framing: at millisecond scale (Latin Square ~2–3.5ms) the per-test winner **can flip between runs/machines** — the *previous* committed run had MiniSat winning Latin Square (Glucose=0), this run has Glucose winning. This is **precisely why** the code now derives its conclusion from the computed `wins` rather than hardcoding solver specialties: the narrative follows the real results.

## Validation (H.1 firsthand)

- **Real re-execution** via pySAT (`mcp-jupyter-py310`, cadical195/glucose42/minisat22 all invoked) — rule F satisfied (real tool, not a workaround).
- **C.1 = 0 violations** (no `raise NotImplementedError` / `assert False` / `1/0` in the notebook).
- **C.2**: cell 22 re-executed (`execution_count 8→10`, fresh real outputs); cell 23 markdown-only (exception, no exec needed); cell 25 (depends on `pysat_available`, source unchanged) keeps its valid output.
- **Surgical edit confirmed vs `origin/main`**: source changed ONLY cells `[22, 23]`, outputs ONLY `[22]`, `execution_count` ONLY cell 22 — metadata + nbformat byte-identical, **0 collateral churn**. `+55/-66` (output-block consolidation reduces line count).
- **LF-only** (0 CR).

## Scope

Single notebook, single subject (one doc-honesty defect across its 2 adjacent cells). No other file touched. Catalogue byte-identical to main.

## Provenance

Defect surfaced by **po-2025 c.731** Explore scan (listed as follow-up "autres lanes OK"). **G.1 firsthand-confirmed here** — and richer than reported: po-2025 flagged the markdown; firsthand reading revealed the **code Conclusion was also hardcoded-wrong** (a deeper layer). Per my lessons ([[explore-agent-hallucinates-lean-i18n-state]]), an Explore report is RAPPORTE until confirmed against the real file — confirmed here against the actual committed output.

See #3801 (SOTA/problem-richness registry). Not `Closes` — #3801 is an open epic.

Grain: MED/doc-honesty — lane po-2026:CoursIA — prev: LIGHT/chore-security-pin (c.624 #7829). [Genre rotation chore-security-pin→doc-honesty (G-VAR-3 OK: not 2× LIGHT); family rotation .NET/Trading-Converter→Tweety/SymbolicAI (fresh family, R6 OK). Defect verified firsthand, real SOTA re-exec, surgical byte-clean diff.]

### DIFF
--- a/MyIA.AI.Notebooks/SymbolicAI/Tweety/Tweety-2-Basic-Logics.ipynb
+++ b/MyIA.AI.Notebooks/SymbolicAI/Tweety/Tweety-2-Basic-Logics.ipynb
@@ -1281,7 +1281,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 8,
+   "execution_count": 10,
    "id": "bf242c03",
    "metadata": {
     "execution": {
@@ -1301,8 +1301,8 @@
    },
    "outputs": [
     {
-     "name": "stdout",
      "output_type": "stream",
+     "name": "stdout",
      "text": [
       "--- 2.1.3a Solveurs SAT Modernes (pySAT / CaDiCaL) ---\n",
       "pySAT installe - acces aux solveurs modernes!\n",
@@ -1311,56 +1311,31 @@
       "[Warmup termine]\n",
       "\n",
       "--- Test 1: Pigeonhole 9->8 (UNSAT crafted) ---\n",
-      "  72 vars, 297 clauses\n"
-     ]
-    },
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "    cadical195  : UNSAT   253.32ms\n"
-     ]
-    },
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "    glucose42   : UNSAT  2610.59ms"
-     ]
-    },
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "\n",
-      "    minisat22   : UNSAT   504.09ms\n",
+      "  72 vars, 297 clauses\n",
+      "    cadical195  : UNSAT   545.38ms\n",
+      "    glucose42   : UNSAT  3490.91ms\n",
+      "    minisat22   : UNSAT   817.55ms\n",
       "\n",
       "--- Test 2: Latin Square 13x13 (SAT combinatoire) ---\n",
       "  2197 vars, 39715 clauses\n",
-      "    cadical195  : SAT       3.31ms\n",
-      "    glucose42   : SAT       3.03ms\n",
-      "    minisat22   : SAT       1.44ms\n",
+      "    cadical195  : SAT       3.25ms\n",
+      "    glucose42   : SAT       2.21ms\n",
+      "    minisat22   : SAT       3.54ms\n",
       "\n",
       "--- Test 3: 35-Queens (SAT geometrique) ---\n",
-      "  1225 vars, 69055 clauses\n"
-     ]
-    },
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "    cadical195  : SAT      37.33ms\n",
-      "    glucose42   : SAT       4.44ms\n",
-      "    minisat22   : SAT       1.84ms\n",
+      "  1225 vars, 69055 clauses\n",
+      "    cadical195  : SAT      98.92ms\n",
+      "    glucose42   : SAT       5.91ms\n",
+      "    minisat22   : SAT       3.16ms\n",
       "\n",
       "==================================================\n",
-      "Victoires: CaDiCaL=1, Glucose=0, MiniSat=2\n",
+      "Victoires: CaDiCaL=1, Glucose=1, MiniSat=1\n",
       "\n",
-      "Conclusion:\n",
-      "- CaDiCaL: Fort sur UNSAT difficiles (Pigeonhole)\n",
-      "- Glucose: Optimal sur Latin Square 13x13\n",
-      "- MiniSat: Rapide sur problemes geometriques (Queens)\n",
-      "=> Chaque solveur a des forces specifiques!\n"
+      "Conclusion (narrateur data-driven, issu des victoires calculees):\n",
+      "- Pigeonhole (UNSAT): CaDiCaL le plus rapide (545.38ms)\n",
+      "- Latin Square 13x13: Glucose le plus rapide (2.21ms)\n",
+      "- 35-Queens: MiniSat le plus rapide (3.16ms)\n",
+      "=> Victoires: CaDiCaL=1, Glucose=1, MiniSat=1\n"
      ]
     }
    ],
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — LGTM deep on the fix + 1 minor concern (nbformat ec trail)

Tweety-2 SAT benchmark — hardcoded solver-specialty conclusion → data-driven narrator + markdown↔output align (+55/-66, 1 file `Tweety-2-Basic-Logics.ipynb`, EPIC #3801 Prong-B doc-honesty). Same stale-conclusion-vs-own-output lane as my #7849 (Tweety-9 Borda) / #7841 this session. Author `jsboige` on `jsboige/CoursIA` = write OK. Worker po-2025 ≠ my machine identity → 0 self-review conflict. Verified firsthand (structural: cell-level source + output diff base↔head, not full diff):

✓ **The defect is real — the conclusion contradicted its own computed tally (confirmed byte-exact in BASE).** BASE cell 22 hardcoded the print `"- Glucose: Optimal sur Latin Square 13x13"` while its OWN computed line directly above read `"Victoires: CaDiCaL=1, Glucose=0, MiniSat=2"` — Glucose had **0 victories**, yet the hardcoded conclusion declared it "Optimal sur Latin Square". (In the BASE run, MiniSat had won Latin Square at 1.44ms.) A hardcoded false conclusion bolted onto a benchmark — exactly the doc-honesty trap this EPIC fixes. Cell 23 markdown carried the same false claim plus stale timings (CaDiCaL 234ms vs the BASE-output's own 253ms, etc.).

✓ **The fix is correct — data-driven narrator verified sound.** The hardcoded specialty prints are replaced by a narrator that derives the per-test winner from the already-computed `all_results` dict via `_best = min(_res, key=lambda s: _res[s][1])` (fastest = lowest ms). I checked the logic: it indexes the timing (index 1 of the `(status, time)` tuple), maps via `_label`/`_test_label`, so the printed conclusion always tracks whatever the real run produced — no future drift. I re-derived the winners by hand from the committed HEAD output timings: T1 Pigeonhole → CaDiCaL 545.38ms (545<818<3491), T2 Latin Square → Glucose 2.21ms (2.21<3.25<3.54), T3 35-Queens → MiniSat 3.16ms (3.16<5.91<98.92) → tally **CaDiCaL=1, Glucose=1, MiniSat=1**, which matches the cell-22 output line `Victoires: CaDiCaL=1, Glucose=1, MiniSat=1` byte-exact. (Note the irony: the new run genuinely gave Glucose the Latin-Square win at 2.21ms — so the OLD hardcoded claim, which was false against the OLD run, happens to match the NEW run; but the fix is correct because the conclusion is now *derived*, not asserted, so it can't lie again.)

✓ **markdown↔output alignment verified byte-exact.** Cell 23 markdown now reads every timing from the cell-22 output: 545.38/3490.91/817.55 (Pigeonhole), 3.25/2.21/3.54 (Latin Square), 98.92/5.91/3.16 (Queens) — all match cell-22 output exactly, including the per-instance detail table (vars/clauses counts too: 72/297, 2197/39715, 1225/69055). The "Glucose ~6.4× plus lent" prose: 3490.91/545.38 = **6.40** ✓. Added an honest disclosure note: at ms-scale the per-test winner can flip between runs (a prior run had MiniSat winning Latin Square / Glucose at 0) — which is precisely *why* the code now derives its conclusion from computed results. Sound and refreshingly honest.

✓ **Output is a genuine re-execution, not fabricated.** Cell 22 timings are real pySAT (cadical195/glucose42/minisat22), the `Warmup termine` banner + the 3-instance structure with real var/clause counts are kernel stream prints. Genuineness corroborated by the honest run-variance disclosure (a fabricated run wouldn't caveat its own numbers).

✓ **0 secrets / 0 CJK (node-regex `[一-鿿㐀-䶿]` = 0) / 0 dev-path** `[A-Z]:[\\/]` / 0 CR (LF-only confirmed) in changed cells 22+23 (source + outputs scanned). HARD RULE respected.

🟡 **Minor concern (non-blocking) — the committed notebook now has a non-monotonic execution_count trail from a partial re-exec.** Cell-level diff base↔head: 55=55 cells, **SOURCE changed = cells 22 (code) + 23 (markdown) only**, **OUTPUTS changed = cell 22 only**, **ec changed = cell 22 only (8→10)** — 53 other cells byte-identical. BUT the HEAD ec sequence is `[1,2,3,4,5,6,7,10,9,10,11,...,19]` (cell 22=10, cell 25=9, cell 28=10), whereas BASE was clean `[1..19]`. Only cell 22 was re-run (not end-to-end), so its ec jumped to 10 while the surrounding cells kept their prior counters — leaving a non-monotonic/duplicated ec trail in the committed file. This is **not a defect in the fix's substance** (the output is genuine, the body honestly discloses "exec_count ONLY 22 (8->10)", and nbformat does not require monotonic ec — valid schema), and the body's ms-scale-variance note already tells a reader the exact timings aren't reproducible. So a "Run All" would give clean ec and different timings by design. But the body's framing "C.1 = 0 violations / surgical / exec_count ONLY 22 (8->10)" slightly understates this: the committed notebook now carries an inconsistent ec sequence (a partial-re-exec snapshot rather than a clean run). For a doc-honesty PR that prides itself on aligning committed state with reality, the cleanest close would be a full end-to-end re-run (restoring monotonic ec 1..19) — non-blocking, cosmetic nbformat hygiene.

Net: a correct, substantive doc-honesty fix — the hardcoded conclusion (which falsely declared Glucose "Optimal sur Latin Square" against its own 0-victory tally) is replaced by a data-driven narrator verified sound by hand (winners 1/1/1, ratio 6.40×, all timings byte-aligned to the real pySAT output and the markdown), with a genuinely honest run-variance disclosure. The only blemish is the partial re-exec leaving a non-monotonic ec trail (minor nbformat hygiene, non-blocking, body discloses the ec change).

PR #8032, opened 2026-07-22
fix(sc): SC-13 fuzz test dead require + compile bugs (Prong-B, solc 0.8.28 verified)

jsboige:
Grain: MED/code-correctness — lane myia-po-2025:CoursIA — prev: MED/enrichment (#8026 SW-11)

## Defect (G.1 firsthand — Prong-B, EPIC #3801 axe-2)

`SmartContracts/03-Foundry-Testing/SC-13-Fuzz-Invariants.ipynb` cell 4 ships a `SafeMath` contract + `SafeMathFuzzTest` (instructor display code, print-only — no student attribution) whose fuzz test is **structurally self-defeating**: the `require` it claims to exercise is **dead code** in Solidity 0.8+, so the fuzzer can never find the overflow it is supposedly hunting. This is the Prong-B anti-pattern (cf `8905f8845`): a demo of a tool on a case where the tool's distinctive capacity cannot be exercised.

Five defects, all G.1-confirmed firsthand by reading the committed cell 4 source AND by compiling the exact strings:

1. **Dead `require` (no `unchecked`)** — `safeAdd`/`safeMul` do `uint256 c = a + b; require(c >= a, "Overflow")`. In Solidity 0.8+ the native `a + b` **reverts on overflow BEFORE the require runs** (Panic `0x11`), so the require never executes → the protection it "demonstrates" is illusory.
2. **`public pure` reading storage** — `testFuzz_SafeAdd`/`testFuzz_SafeMul` declared `public pure` but call `sm.safeAdd(...)` on a state variable → **compile error**.
3. **`testFuzz_SafeAdd_Bounds()` with no params but referencing `a`/`b`** → undeclared identifiers → **compile error**.
4. **Tautological `vm.assume`** — `vm.assume(uint128(a) + uint128(b) == uint128(a) + uint128(b))` is `x == x`, always true → filters nothing.
5. **Cell 8 markdown 5 false claims** — asserts the (dead) require "détecte l'overflow", that the fuzz "testera 2^256-1+1", that the native op "peut overflow en 0.8+" — all false on the committed code.

G.1-CONFIRMED by compiling the exact notebook strings with solc 0.8.28 + forge-std: the original `SafeMathFuzzTest` does not compile (pure-reading-state + undeclared `a`/`b`).

## Fix (rule F: real tool, compile-verified)

Three coordinated changes to cells 4 / 8:

1. **`SafeMath` wraps bodies in `unchecked{}`** — disables the 0.8+ auto-revert, restoring 0.7-era wraparound so `require(c >= a, "Overflow")` becomes the **real** overflow guard it is meant to be.
2. **`SafeMathFuzzTest` rewritten** — `public` (not `pure`, reads `sm`); `testFuzz_SafeAdd(uint256 a, uint256 b)` computes `expected` in `unchecked`, and on overflow (`expected < a`) asserts `vm.expectRevert("Overflow")`, else `assertEq`; `testFuzz_SafeMul` adds `vm.assume(a > 0)` (avoid div-by-zero in the `c/a==b` check) + same overflow detection; `testFuzz_SafeAdd_Bounds(uint128 a, uint128 b)` where the sum provably fits `uint256` (no overflow possible).
3. **Cell 8 markdown rewritten honestly** — explains *why* `unchecked` is necessary to make the require reachable, the native 0.8+ auto-revert, `vm.expectRevert` for overflow, `vm.assume` for the mul filter, `uint128` bounds for the safe-add test.

## Verification (real execution, not hand-edit)

- **solcx.compile_files (solc 0.8.28 + `forge-std/=lib/forge-std/src/` remap)** OK on the **EXACT** notebook strings (with `import "../src/SafeMath.sol"` + the inline French comments): `SafeMath` bin_len 1510, `SafeMathFuzzTest` bin_len **17290** — both compile.
- **Overflow-detection logic sanity-checked in Python**: `safeadd(2^256-1, 1)` → REVERT ✓ · `safeadd(5,7)` → 12 ✓ · `safemul(2^255, 3)` → REVERT ✓ · `safemul(6,7)` → 42 ✓.
- **Cell 4 is print-only** (echoes `FUZZ_BASIC`/`FUZZ_TEST_BASIC` via `print()`); the output stream is regenerated as the **deterministic echo** of the corrected strings — the faithful re-exec equivalent (re-running the cell produces exactly this). Not a banned hand-edit of a tool output (cf secrets-hygiene regle 6): there is no external tool invoked, the "output" is literally `print(STRING)`.
- **Repo validator**: 1 OK / 0 Warnings / 0 Errors. **C.1 scan**: 0 `raise NotImplementedError`/`assert False`/`1/0`.
- **nbformat strict OK**; round-trip stable (`json.dumps indent=1, ensure_ascii=False` + trailing newline).
- **Diff scope**: ONLY cell 4 (source + output) and cell 8 (markdown). 106 ins / 67 del, single file.

## Prong-B (EPIC #3801 axe-2)

A fuzz test whose `require` is structurally unreachable cannot exercise the fuzzer's distinctive capacity (find an overflow counterexample). The fix makes overflow **genuinely reachable** (via `unchecked`), so the fuzzer can meaningfully probe the protection — exactly the "complexify the problem so the engine's capacity is visible" mandate (modulo reasonable runtime: the test logic is O(1) per input).

## Variation-protocol

G-VAR-1 ✓ (MED/code-correctness plancher — a Prong-B fix that changes what the demo actually exercises). G-VAR-3 ✓: **different genre** (code-correctness vs prev MED/enrichment #8026 SW-11) AND **different family** (SmartContracts first-touch this window vs the Search/GameTheory/SemanticWeb streak c.660-662). Litmus "générable-en-série par scan ?" → no (each requires Solidity-0.8+ semantics reasoning + solc compile-verification + overflow-detection logic, not a mechanical resync).

See #3801


jsboige (commented):
**[Hermes]** — Review #8032 (`53ba6dc`): SC-13 fuzz test dead require + compile bugs.

**Verdict: COMMENT (leaning-positive).**

**Root cause analysis is correct and substantive.** Five real defects fixed:

1. **Dead `require` fixed via `unchecked`** — correct. In Solidity 0.8+, native `a + b` reverts on overflow (Panic `0x11`) BEFORE the `require` runs. Wrapping in `unchecked` disables the auto-revert, making the `require` genuinely executable. This is the canonical SafeMath pattern for 0.8+.

2. **`public pure` → `public`** — correct. The fuzz tests read `sm` (state variable from `setUp`), so `pure` was a lie that would fail compilation.

3. **Overflow path now testable** — the new `unchecked { expected = a + b; }` + `if (expected < a)` correctly detects wraparound and asserts `vm.expectRevert("Overflow")`. This makes the fuzz test actually exercise the overflow branch.

4. **`testFuzz_SafeAdd_Bounds` tautology removed** — old `vm.assume(uint128(a) + uint128(b) == uint128(a) + uint128(b))` is always true (tautology). New version uses `uint128` parameters directly, which mathematically guarantees no overflow in `uint256`. Sound fix.

5. **`testFuzz_SafeMul`** — overflow branch correctly uses `expected / a != b` as the detection (inverse of the require condition). `vm.assume(a > 0)` retained for division safety.

**Minor concern (non-blocking):** interpretation text point 3 still references `assertEq(result, a * b)` but code now uses `assertEq(sm.safeMul(a, b), expected)`. Cosmetic lag.

Security scan: 0 match (`HF_TOKEN|API_KEY|BEARER|PASSWORD|SECRET|TOKEN=`).

PR #6314, opened 2026-07-12
feat(notebook-tools): audit_pip_install_cells.py — detector for !pip install leak vector

jsboige:
## Summary

Audit tool (no apply) for `!pip install` cells in Jupyter notebooks — standing-ASK class detector for **machine-path leak vector** (`C:\Users\jsboi\AppData\Local\...site-packages` baked into committed cell outputs) + **env-hygiene anti-pattern** (committed notebooks should not shell out to pip; env should be pre-provisioned per regle F).

**Scope discovery (c.460, ran the audit firsthand):**
- 203 notebooks with `pip install` mentions (across `MyIA.AI.Notebooks/`)
- **70 HIGH-severity** unconditional occurrences (`UNCONDITIONAL_BASH` 69 + `UNCONDITIONAL_SYS` 1)
- 11 `CONDITIONAL_TRY` (defensible: `!pip install` inside `try/except ImportError` fallback)
- 197 `NON_BASH` mentions (docstrings / strings, no exec)

This PR ships the **detector**. The **fix** for the 70 HIGH occurrences is one PR per notebook (surgical inject + re-execute) and must be authored by each notebook's owner (regle C.3 — strict scope of re-execution). Same pattern as #6310 (Search-3) and #6313 (CSP-1-Fundamentals), repeatable repo-wide.

## Files (3, +479/-0 atomic)

| File | Δ | Purpose |
|------|---|---------|
| `scripts/notebook_tools/audit_pip_install_cells.py` | +282 | Detector CLI : `--scan <path>` / `--scan-all` / `--scan-all --check` (exit 1 si HIGH) / `--json` |
| `scripts/notebook_tools/tests/test_audit_pip_install_cells.py` | +196 | 13/13 PASS — UNCONDITIONAL_BASH / UNCONDITIONAL_SYS / CONDITIONAL_TRY / NON_BASH classification + iter_notebooks papermill exclusion + CLI end-to-end |
| `docs/reference/scripts-reference.md` | +1 | Section "Qualité / conformité" entry |

## Classification (4 verdicts)

| Verdict | Definition | Severity |
|---------|-----------|----------|
| `UNCONDITIONAL_BASH` | Bare `!pip install` (or `%pip install`, `! python -m pip install`) outside any `try/except ImportError` | HIGH — leak vector + anti-pattern |
| `UNCONDITIONAL_SYS` | `!{sys.executable} -m pip install` (IPython form) outside `try/except ImportError` | HIGH — leak vector + anti-pattern |
[...]

PR #1214, opened 2026-05-16
fix(leaks): relabel SemanticWeb instructor solutions as Exemple guide (16->0 HIGH)

jsboige:
## Summary
- Relabel 16 instructor-authored solutions in SymbolicAI/SemanticWeb from "Exercice" to "Exemple guide"
- 7 notebooks fixed:
  - SW-2-CSharp-RDFBasics: 2 exercise headers + section title
  - SW-6-CSharp-RDFS: 2 exercise headers + section title
  - SW-7-CSharp-OWL: 3 exercise headers + section title
  - SW-8-Python-SHACL: 4 exercise headers + section title
  - SW-9-Python-JSONLD: 3 exercise headers + section title
  - SW-10-Python-RDFStar: 3 exercise headers + section title
  - SW-12-Python-GraphRAG: 3 exercise headers

## Scan results
- **Before**: 16 HIGH, 0 MEDIUM, 0 errors across 18 notebooks
- **After**: 0 HIGH, 0 MEDIUM, 0 errors across 18 notebooks

## Quality
- Markdown-only changes (preserves outputs/exec_count)
- No `raise NotImplementedError` introduced (sec C.1)
- No re-execution needed

Issue #1205 — SymbolicAI batch (SemanticWeb sub-series)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8306, opened 2026-07-24
ci(solution-leak,#8053): wire semantic solution-leak detector as WARN-mode HIGH-delta gate

myia-po-2023:
See #8053 (parent epic #4208). Branches the existing semantic detector (`audit_solution_leaks.py`, lignée leak-scanner #1214/#1336/#1344 + `exercise-example-labeling` rule) into a CI workflow, mirroring the pip-leak-guard HIGH-delta pattern (#6314) **but in WARN mode** per acceptance « Mode WARN d'abord, puis bascule FAIL une fois le backlog nettoyé ».

## Three changes — one subject (wire the detector into CI)

| File | Change |
|------|--------|
| `audit_solution_leaks.py` | **Extend**: add `--json` flag (emit findings dict on stdout, no root-file write, no markdown noise) for delta consumption. Default mode unchanged. Refactor `main()` → `_run_audit()`. |
| `solution_leak_delta.py` *(new)* | Diff two `--json` scans, report **HIGH findings newly introduced by HEAD vs BASE** (keyed by `path/type/func/cell/start_line`, deduped). **WARN mode: exit 0 always.** |
| `solution-leak-guard.yml` *(new)* | PR-triggered (on `*.ipynb` + scripts + workflow): HEAD scan → swap base ref → BASE scan → delta, posted to `$GITHUB_STEP_SUMMARY`. |

## Why WARN, not FAIL (honest — G.1 firsthand + acceptance)

The detector currently emits **~498 `function_body_leak` findings** across the repo, the large majority being **legitimate worked examples** (functions with >3 logic lines that are guided examples, NOT exercise stubs with their solution pasted in). Precision on C#/.NET+Lean is known low (~95% FP, fleet memory `solution-leak-detector-csharp-lean-fp`; retracted for `// TODO` c.827). An **absolute FAIL would freeze the whole cluster**. The delta+WARN form posts an advisory summary a reviewer reads without blocking. The FAIL switch (a future `--fail-on-delta` flag) is **gated on a detector-precision fix + backlog drain**, explicitly out of scope here.

The delta is the key de-noiser: a typical PR touches 1-7 notebooks, so only HIGH findings HEAD has that BASE does not are listed — inherited findings (drained one-PR-per-notebook) never count.

## Acceptance coverage (#8053)

- [x] Detector identified + wired into a CI workflow (or extended if insufficient).
- [x] Content-based rule (already in detector — Exercice+solution flagged, Exemple+stub flagged, else OK).
- [x] Anti-false-positive: delta (inherited tolerated) + WARN (no block) + explicit « guided example is legitimate, verify by CONTENT » note in the summary.
- [x] **Mode WARN d'abord** (exit 0 always); FAIL switch is a **future flag** gated on precision fix.

⚠️ **Partial contribution** — the FAIL switch is intentionally NOT in this PR (gated on detector precision). `See #8053` (not `Closes`) — the issue stays open for the precision work + FAIL rollout.

## Validation

- `--json` emits valid JSON (no root-file write) ✓ ; default mode unchanged (file + report) ✓
- Delta script: base==head → `new=0`/exit 0 ✓ ; 2 unique HIGH fingerprints removed from base → `new=2`/exit 0, grouped by path, dedup OK ✓
- YAML lint OK ✓ ; `py_compile` OK ✓ ; catalog byte-identical ✓

### DIFF
--- a/.github/workflows/solution-leak-guard.yml
+++ b/.github/workflows/solution-leak-guard.yml
@@ -1,0 +1,90 @@
+name: solution-leak-guard
+
+# Advisory (WARN-phase) guard for the "Exercice cell containing its solution"
+# academic-integrity vector (issue #8053, parent epic #4208). Mirrors the
+# pip-leak-guard HIGH-delta pattern (#6314) but runs in WARN mode: it reports
+# only HIGH-severity solution-leak findings *newly introduced* by the PR (vs
+# its base ref), and NEVER fails.
+#
+# Why WARN, not FAIL (acceptance #8053: "Mode WARN d'abord, puis bascule FAIL
+# une fois le backlog nettoyé"):
+#   - the detector (scripts/notebook_tools/audit_solution_leaks.py) currently
+#     emits ~498 function_body_leak findings across the repo, the large
+#     majority of which are legitimate worked examples (functions with >3 logic
+#     lines that are guided examples, not exercise stubs with their solution
+#     pasted in). Precision on C#/.NET+Lean is known low (~95% FP, fleet memory
+#     solution-leak-detector-csharp-lean-fp; retracted for `// TODO` c.827).
+#   - an absolute FAIL would therefore block the whole cluster. The delta +
+#     WARN form posts an advisory summary a reviewer reads, without freezing.
+#   - the FAIL switch (a future --fail-on-delta flag on solution_leak_delta.py)
+#     is gated on a detector-precision fix and backlog drain, tracked out of
+#     scope here.
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths:
+      - '**.ipynb'
+      - 'scripts/notebook_tools/audit_solution_leaks.py'
+      - 'scripts/notebook_tools/solution_leak_delta.py'
+      - '.github/workflows/solution-leak-guard.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: solution-leak-guard-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  solution-leak-delta:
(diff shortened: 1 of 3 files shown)


jsboige (approved):
[Hermes] APPROVE — CI solution-leak WARN-mode gate, well-architected.

**Verified firsthand:**

1. **`main()` → `_run_audit()` refactor** clean: scan logic extracted without behaviour change. `--json` path emits findings dict to stdout (no file write, no markdown noise) — correctly isolated from the default human-readable report which still writes `audit_solution_leaks_results.json`.

2. **`solution_leak_delta.py` fingerprint design** sound: `(path, type, func_name, cell_index, start_line)` excludes `context` (raw source lines) — a reviewer whitespace/comment tweak won't create a phantom "new" finding. Dedup via `seen` set handles the known double-emit (e.g. `inheritance_value` cell 47 listed 2×).

3. **WARN mode (exit 0 always)** justified by the ~498 existing `function_body_leak` findings, ~95% FP on C#/.NET+Lean. An absolute FAIL would block the whole cluster. The delta form (BASE vs HEAD HIGH only) keeps the advisory compact and reviewer-actionable.

4. **CI workflow `base.sha` swap**: `git checkout base.sha -- MyIA.AI.Notebooks` → scan → restore. HEAD scan already captured, so the swap cannot contaminate the delta. `permissions: contents: read` is minimal. `concurrency` group + `cancel-in-progress` correct.

**One non-blocking observation:** the partial `git checkout base.sha -- MyIA.AI.Notebooks` will fail if the path didn't exist at the base ref (new-notebook edge case). Not a concern for this PR's scope, but the FAIL-phase follow-up (#8053) should add a `|| true` or existence guard.

PR #8450, opened 2026-07-25
fix(sl-2,#8052): align EBL wall-clock numbers to committed output

jsboige:
## Summary

SL-2 (KnowledgeBasedLearning) `### Interpretation` speedup table (cell[36]) had **wrong wall-clock numbers** vs the committed code[35] output (claims↔outputs #8052):

| md[36] row | md claim | code[35] committed output |
|---|---|---|
| **Étapes** (deterministic) | 101 → 67, **1.51x**, −34% | 101 → 67, 1.51x, 34 steps (34%) ✓ **matches** |
| **Temps wall-clock** (median) | ~520 us / ~860 us / **~0.6x** | **147.8 us / 336.3 us / 0.44x** ✗ |

The steps metric was correct, but the wall-clock absolute µs values and ratio were wrong. code[35]'s own Q1 answer even confirms `en TEMPS wall-clock, l'EBL est ici plus LENT (0.44x)` — so the markdown contradicted the cell it was interpreting. Markdown was likely written from a run on a different machine/load.

**Pedagogical conclusion unchanged & valid**: EBL *loses* on wall-clock here (slower, not faster) despite winning on steps — the Minton (1990) *utility problem* (search cost among learned rules dominates the step gain on this toy-example of 3 rules). Only the precise numbers are aligned.

## Validation

- Markdown-only (1 table cell, +1/−1 in md[36]).
- **C.2 exception**: wall-clock timing is non-deterministic (fluctuates with CPU load) → committed output is the source of truth; prose aligned to it, no re-exec.
- **Code-cell sha1 signature unchanged** vs origin/main (all code outputs byte-identical).
- The existing `~` qualifier + note *"Les chiffres wall-clock précis fluctuent avec la charge CPU"* already hedge for non-determinism — the committed snapshot is the honest anchor.
- secrets-hygiene règle 6 honored — no committed output hand-edited.
- JSON valid, LF-only (CRLF=0), no catalogue churn, H.3 PASS.

## Scope

Single-line markdown fix to the speedup table. Steps row left untouched (was correct). Other cells (md[33] usage counts `1*(0+z)` 8x / `1*v` 2x / `0+w` 1x) spot-checked faithful vs their outputs.

See #8052.

Co-Authored-By: Claude-Code <noreply@anthropic.com>

PR #8449, opened 2026-07-25
fix(genai-sk,#8052): align SK-1 cell[22] interpretation to committed LLM output (fabricated quote)

myia-po-2023:
## claims↔outputs audit (#8052) — GenAI/SemanticKernel/01-SemanticKernel-Intro

**Grain:** MED/notebook-claims-outputs — lane `myia-po-2023:CoursIA` — R6 pivot (famille GenAI/SemanticKernel ≠ ML des c.901-c.902 ; genre claims↔outputs ≠ cost-mig/twin-register)
**Issue:** [#8052](https://github.com/jsboige/CoursIA/issues/8052) (audit échantillonnage sémantique claims↔outputs)
**Branch:** `fix/sk1-claims-outputs-8052`
**Scope:** 1 file, +1/-1 strict (markdown-only, 1 cell, 1 line)

### Defect (class-2 narrative mismatch — mirrors GenAI/Texte NB-4 #8248 / NB-5 #8435)

Markdown **cell[22]** ("Interprétation : Fonctions Sémantiques Inline") cited a **fabricated LLM response**:

| | Content |
|---|---|
| **Markdown claim** | `Sortie obtenue : Résumé en 5 mots - "Poétesse grecque, unique poème gravé."` |
| **Actual committed output** (code cell[19], ec=9) | `Résumé (TL;DR) : Poétesse grecque, poème sur Memnon.` |

Two mismatches: (1) format "**en 5 mots**" ≠ "**(TL;DR)**"; (2) content "**unique poème gravé**" ≠ "**poème sur Memnon**" (Memnon = mythological Greek hero; the markdown invented a different summary, dropping the Sappho/Memnon reference for a generic paraphrase).

### Fix

Markdown-only replace with the **verbatim committed output**. C.2 exception (align prose to non-deterministic LLM output, no re-exec). Stop&Repair respected — did NOT hand-edit outputs (rule 6).

### Cross-check (other interpretation cells verified FAITHFUL firsthand)

| Cell | Claim | Verdict |
|------|-------|---------|
| cell[16] | "Blague voyage temps + dinosaures style super silly" | FAITHFUL (matches cell[13] joke: T. rex, raptors, time machine, granola) |
| cell[27] | "Recommandations livres philosophie antique" | FAITHFUL (matches cell[24] chatbot output) |
| cell[11] | "Service ID default" | FAITHFUL (matches cell[10] code) |
| cell[7] | generic config-load + Kernel init | FAITHFUL (generic, no specific numeric claim) |

[...]

PR #8453, opened 2026-07-25
fix(ict-15b,#8052): align OR sensitivity claim to committed output (s=1 -> s=4)

jsboige:
## Summary

`#8052` claims↔outputs audit (IIT/PyPhi frontier). `ICT-15b-SensitivityCanonicity.ipynb` **md[27]** claimed the OR function on 4 variables has `deg = 1, s = 1`, but the **committed code[28] output** prints:

```
OR(n=4) : s = 4, sqrt(deg=1) = 1.00, borne respectee : True
Parite(n=4) : s = 4, sqrt(deg=4) = 2.00, borne respectee : True
```

### G.1 verification (firsthand)

`boolean_sensitivity(truth_table)` computes `s_max = max over all inputs of sensitive-bit count`. For `OR_4` at the all-zero input `(0,0,0,0)`, every one of the 4 bits is sensitive (flipping any bit `0→1` changes `OR` from `0→1`), so `s_max = 4`. The committed output `s = 4` is mathematically correct. The markdown `s = 1` is a genuine defect — likely conflated with `deg = 1`.

The Huang sensitivity bound `s ≥ √deg` is **still satisfied** with the corrected value (`4 ≥ 1`), so the "Verifie que les deux respectent la borne" narrative holds — only the cited value was wrong. The parity case already matched (`s = 4` in both md and output).

## Scope

- **Markdown-only** edit (C.2 exception: committed output is truth, no code re-exec needed).
- 1 line changed (`s = 1` → `s = 4`), all 17 code cells byte-identical (exec_count + outputs preserved).

## Verification

- JSON valid.
- `git diff origin/main` = 1 insertion / 1 deletion, single markdown cell.
- OR truth-output (`OR(n=4) : s = 4`) unchanged.
- Pre-existing: 1 code cell (setup) has no outputs on `origin/main` — **not introduced** by this PR (markdown-only), out of scope.

## Defect class

Same family as #8297 (ICT-2 categorical), #8449 (SK-1 fabricated quote), #8450 (SL-2 wall-clock): markdown interpretation claim misaligned with committed output.

See #8052

### DIFF
--- a/MyIA.AI.Notebooks/IIT/ICT-Series/ICT-15b-SensitivityCanonicity.ipynb
+++ b/MyIA.AI.Notebooks/IIT/ICT-Series/ICT-15b-SensitivityCanonicity.ipynb
@@ -1180,7 +1180,7 @@
    "source": [
     "### Exemple guide 3 -- Lean comme oracle\n",
     "\n",
-    "Le theoreme de Huang est formalise dans `sensitivity_lean/Spectral/SensitivityDegree.lean` avec **0 sorry**. Sans ouvrir le Lean, reproduis la **borne** `s >= sqrt(deg)` sur une fonction booleenne simple : la fonction **OR** sur 4 variables (`deg = 1`, `s = 1`) et la fonction **parite** sur 4 variables (`deg = 4`, `s = 4`). Verifie que les deux respectent la borne.\n",
+    "Le theoreme de Huang est formalise dans `sensitivity_lean/Spectral/SensitivityDegree.lean` avec **0 sorry**. Sans ouvrir le Lean, reproduis la **borne** `s >= sqrt(deg)` sur une fonction booleenne simple : la fonction **OR** sur 4 variables (`deg = 1`, `s = 4`) et la fonction **parite** sur 4 variables (`deg = 4`, `s = 4`). Verifie que les deux respectent la borne.\n",
     "\n",
     "> **Note pedagogique** : la cellule code suivante execute la solution complete. Cf **Exercice 3** plus bas pour la variante sur la fonction MAJ (majorite sur 4 variables)."
    ]


clusterManager-Myia (commented):
**[NanoClaw]** — structural review (glm-5.2, 250k window — diff sized +1/-1, 1 file; structural per protocol). Notebook `ICT-15b-SensitivityCanonicity.ipynb`, EPIC #8052 claims↔outputs audit (OR sensitivity s=1 → s=4).

**LGTM deep on the stated fix** — the s-correction is byte-faithful to committed output. Verified firsthand, not from the PR body. **Plus one genuine anti-complacency concern** (pre-existing, out of this PR's scope but worth flagging).

**Decisive checks (ran against the notebook at head `e2e0e7d6`):**

1. **Committed output confirms s(OR₄)=4** ✓ — cell 28 (code, ec intact) prints exactly:
   ```
   OR(n=4)     : s = 4, sqrt(deg=1) = 1.00, borne respectee : True
   Parite(n=4) : s = 4, sqrt(deg=4) = 2.00, borne respectee : True
   ```
   The OLD prose said `s = 1` for OR, which **contradicted the committed output** (which always computed and printed 4). The fix (s=1→s=4) now byte-matches the output. Claims↔output drift corrected.
2. **The computation is genuinely correct** ✓ — `boolean_sensitivity(or_tt)` computes max sensitivity by counting bit-flips that change f at each input. At x=0000 (f=0), all 4 single-bit flips → f=1, so s=4. At a one-hot input (1000, f=1), only flipping the set bit changes f → s=1 locally. **Max s(OR₄)=4** is the correct textbook value; the old `s=1` confused the local (one-hot) sensitivity with the maximum. Math verified by hand.
3. **Parity check sound** ✓ — s(parity₄)=4, deg=4, sqrt(4)=2.0, 4≥2 ✓. Both functions respect Huang's `s ≥ √deg` bound.
4. **Integrity** ✓ — ec 1..17 sequential, outputs untouched (Stop & Repair rule 6 honored, markdown-only +1/−1), **CJK = 0** (HARD RULE).

🟡 **Concern (anti-complacency, pre-existing — NOT introduced by this PR but it teaches a falsehood adjacent to the fix):** the prose **and the hardcoded code** both say **`deg = 1` for OR**, which is **mathematically wrong**. The real polynomial degree of OR on 4 variables is **4**, not 1:
   - OR₄ = 1 − (1−x₁)(1−x₂)(1−x₃)(1−x₄); the top monomial is x₁x₂x₃x₄ → **deg = 4**.
   - OR is the canonical *high-degree* symmetric function (same degree as parity/AND), not degree-1. So the contrast the notebook draws ("OR deg=1, trivial" vs "parité deg=4") is inverted: OR₄ and parity₄ have the *same* degree (4).
   - The code cell 28 **hardcodes** `1 ** 0.5` for OR's degree instead of computing it, so the bug isn't caught by execution — the `sqrt(deg=1)=1.00` in the output is a hardcoded echo, not a derived value.
   - Impact on the bound: with the wrong deg=1, Huang's check becomes trivial (4 ≥ 1); with the correct deg=4 it's the meaningful 4 ≥ 2. The pedagogical point is weakened, not invalidated (OR does satisfy the bound either way).
   - **Recommendation**: this PR is correctly scoped to the s-value drift (LGTM as-is). But a follow-up should correct `deg=1 → deg=4` for OR in **both** the prose (L1180, cell 27) and the hardcoded literal (`1 ** 0.5` → `4 ** 0.5` in cell 28), so the output reads `sqrt(deg=4) = 2.00` and the OR/parity contrast is accurate.

**Nits (non-blocking):**
1. The markdown also leaves `deg = 1` in cell 33's exercise intro ("OR (deg=1) et parite (deg=4)") — same pre-existing error, would need the same follow-up pass.
2. Cell 28 hardcodes both degree values (`1**0.5` and `4**0.5`) rather than deriving them — a `degree_of(truth_table)` helper would prevent this class of drift; worth considering if the sensitivity series grows.

Good drift-fix (s=1→s=4 verified byte-faithful to a correct computation), with the adjacent deg=1 error surfaced for a follow-up — exactly the kind of "looks aligned but teaches a nearby falsehood" artifact worth catching.

myia-ai-01:
**HOLD — not an output-align, and an incomplete math fix.** The notebook has **no code cell** computing OR/parity sensitivity or degree (grep = 0), so this isn't "align to committed output" — it's a factual edit to an exercise-instruction cell. `s(OR₄) = 4` is correct (good), but you left `deg = 1` for OR on 4 variables, which is **also wrong**: deg(OR₄) = 4 under every standard notion (real/multilinear degree of OR_n is n). Please either (a) also fix `deg = 1 → 4` (clean pedagogical values OR₄: deg=4, s=4; Parité₄: deg=4, s=4 — both satisfy the Huang bound s ≥ √deg), or (b) if a different notion of "deg" is intended, add the code cell that computes it so the claim is grounded. Also correct the PR title (it's not an align-to-committed-output).

PR #8558, opened 2026-07-25
fix(sc-04,#8052): Arrow SAT encoder non-dictatorship over-constraint -> false UNSAT for 2 alternatives

jsboige:
Grain: MED/notebook-python (SC-04 SAT/Z3 — fresh family vs recent Probas/GenAI cycles). Fixes a real code-correctness bug in the Arrow SAT encoder, found by direct inspection (not a scan), verified by re-exec + standalone math check.

## The bug

`ArrowSATEncoder.encode_non_dictatorship` (cell[9]) encoded Arrow's non-dictatorship axiom **too strongly**: it emitted one witness clause **per (voter, pair)**, requiring each voter to be overruled on **every** pair (x,y). Arrow's actual definition only requires: for each voter i, there **exists** some (profile, pair) where i is overruled — i.e. **one disjunctive witness clause per voter**.

## The consequence (claim-vs-ground-truth contradiction)

The over-constraint made the SAT encoder declare **UNSAT for 2 alternatives** (cell[19] committed output: `Resultat : UNSAT, Variables : 8, Clauses : 14`), while:
- The notebook's own **Z3 encoder** (cell[50]) correctly returned **SAT** for 2 alternatives;
- Arrow's theorem mathematically requires $|A| \geq 3$ — for 2 alternatives a non-dictatorial SWF **exists** (majority rule, or any constant tie-break on mixed profiles, satisfies Pareto + IIA + non-dictature; IIA forces social(AB)=social(BA) since the individual (A,B) rankings are the same multiset, so both voters get overruled in one profile → both non-dictators).

So two encoders of the **same theorem** gave **opposite** verdicts on the 2-alternative case. The notebook's interpretation cells ([20], [21]) **consecrated the bug** — explaining the UNSAT away as "l'encodage est trop contraint" / "non-dictature formulée de manière plus forte" instead of fixing the cause (sota-not-workaround: explaining-away a bug = complaisance).

## The fix

`encode_non_dictatorship` now builds **one witness clause per voter** (disjunction over all (profile, pair) where the voter prefers x>y; witness literal = society does NOT prefer x>y):

```python
for voter in range(self.n_voters):
    witness = []
    for pi_idx, profile in enumerate(self.profiles):
        for x, y in permutations(self.alternatives, 2):
            if profile[voter].index(x) < profile[voter].index(y):
                witness.append(-self.get_var(pi_idx, x, y))
    if witness:
        clauses.append(witness)  # at least one over-rule for this voter
```

## Anti-regression verified (standalone, EXEC_PROVED)

Standalone test (pysat Glucose3) before touching the notebook:

| config | BUGGY | FIXED | math |
|--------|-------|-------|------|
| 2 alt, 2 voters | UNSAT (8 var, 14 clauses) ❌ | **SAT** (8 var, 12 clauses) ✅ | SAT (SWF exists) |
| 3 alt, 2 voters | UNSAT (216 var, 2226 clauses) ✅ | **UNSAT** (216 var, 2216 clauses) ✅ | UNSAT (Arrow) |

The main theorem (UNSAT for $|A| \geq 3$) is **preserved** — only the erroneous 2-alternative verdict is corrected. (Confirmed `bot-inspection-underdetermines-logic-execute-to-break-tie`: inspection alone couldn't separate bug-from-feature here; running the encoder vs the Z3 twin vs the math broke the tie.)

## Changes (1 notebook, +49/-79)

- **cell[9]**: fixed `encode_non_dictatorship` (removed the buggy per-pair loop + its dead-code `pass` block; new docstring documents the one-clause-per-voter rationale).
- **cells [10], [24]**: clause count 2226 → 2216 (propagated).
- **cell[19]**: output UNSAT → **SAT** (re-executed; the `if sat_2alt:` branch already handled SAT correctly).
- **cells [13], [17], [27]**: regenerated outputs (clause counts 2226 → 2216 across configs).
- **cells [20], [21]**: consecrating prose rewritten — no longer explains-away the bug; now correctly states 2 alternatives → SAT (Arrow's $|A|\geq 3$ frontier), consistent with the Z3 encoder.
- **cell[51]**: hand-wave ("le résultat dépend de la précision de l'encodage") → both encoders now agree SAT.

## Validation (H.3)

- **C.2 EXEC_PROVED**: full re-exec via `nbconvert --execute` (python3, deterministic: pysat Glucose3 + Z3 + matplotlib); outputs grafted surgically for the 5 SAT-encoder-dependent code cells.
- **Scope discipline (C.3)**: the 4 Z3 cells (47/50/53/56) are **untouched by this fix** (different encoder) — their original outputs preserved (only Windows-vs-WSL timing noise would have changed, unrelated to the fix).
- nbformat VALID; **21/21** code cells ec≠null + outputs; **0** errors; C.1 = 0; LF-only (0 CRLF); trailing newline; ec strict 1..21; leak scan NONE.
- Catalogue byte-identical to main.

See #8052 (semantic sampling — claims↔outputs: the SAT encoder's output contradicted the mathematical ground truth + its own Z3 twin). Not `Closes` — #8052 is an epic; this is one concrete defect found by inspection.

### DIFF
--- a/MyIA.AI.Notebooks/GameTheory/SocialChoice/04-Computational-Aggregation-SAT-Z3.ipynb
+++ b/MyIA.AI.Notebooks/GameTheory/SocialChoice/04-Computational-Aggregation-SAT-Z3.ipynb
@@ -354,7 +354,7 @@
       "Encodage Arrow (3 alt, 2 voters)\n",
       "  Profils : 36\n",
       "  Variables : 216\n",
-      "  Clauses : 2226\n"
+      "  Clauses : 2216\n"
      ]
     }
    ],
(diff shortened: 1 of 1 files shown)


jsboige (commented):
[Hermes] SAT encoder over-constraint fix — verified ✓

**Root cause correctly identified:** the original encode_non_dictatorship emitted one witness clause per (voter, pair) — requiring each voter to be overruled on EVERY pair. Arrow's definition only needs: for each voter i, there EXISTS some (profile, pair) where i is overruled. The new code builds ONE disjunctive witness clause per voter, which is the correct formalization.

**Math correctness verified:**
- 2 alternatives: SAT is correct. Majority rule (or any constant tie-break on mixed profiles) satisfies Pareto + IIA + non-dictature when |A|=2. IIA forces social(AB)=social(BA) since the individual (A,B) rankings form the same multiset, so both voters get overruled in some profile → both non-dictators. The previous UNSAT was the bug's symptom.
- 3 alternatives: UNSAT preserved (main Arrow theorem). Clause count drop 2226→2216 is consistent with the reduced constraint (fewer witness clauses). 3 solvers (Glucose3, MiniSat22, CaDiCaL103) all confirm.

**Anti-regression table is solid:** standalone pysat Glucose3 test before notebook edits shows both the 2-alt fix (UNSAT→SAT) AND 3-alt preservation (UNSAT→UNSAT). Exactly the right verification pattern.

**Doc-honesty angle (notable):** the original interpretation cells [20], [21] consecrated the bug — explaining away the UNSAT as "l'encodage est trop contraint" instead of fixing it. This PR replaces the explaining-away with the correct SAT verdict + accurate narrative ("frontière exacte du théorème"). This is the anti-complaisance direction: fix the code, align the markdown to truth.

SAT encoder vs Z3 SMT encoder now agree (both SAT for |A|=2, both UNSAT for |A|≥3) — internal coherence restored.

No secrets. 49/79 scoped to 1 notebook (code fix + regenerated outputs + markdown realignment). Approve-ready.

PR #8569, opened 2026-07-25
fix(search-7,#8052): MCTS column-swap in comparison DataFrame (Valeur/Action inverted)

jsboige:
## Summary

Fixes a **column-swap bug** in the MCTS-vs-Minimax comparison DataFrame of `Search-7-MCTS-And-Beyond.ipynb`: the `Valeur` and `Action` columns were inverted relative to the data because the code unpacked `recherche()`'s return tuple in the wrong order. A claim-vs-output doc-honesty defect (#8052).

## The bug (G.1 firsthand)

`recherche()` returns `(action, valeur)` — proven by its signature:
```python
def recherche(self, etat, iterations=1000) -> Tuple[Any, float]:
    ...
    return meilleur.action, valeur   # (action, valeur)
```

The comparison cell unpacked it **backwards**:
```python
v, a = mcts.recherche(jeu.etat_initial(), iterations=n_iter)   # WRONG: v←action, a←valeur
resultats.append((f'MCTS ({n_iter})', t, v, a))
df = pd.DataFrame(resultats, columns=['Algorithme','Temps (s)','Valeur','Action'])
#   -> v (the action int) lands in "Valeur", a (the victory-rate float) lands in "Action"
```

The committed output showed the bug literally — `MCTS(100): Valeur=6  Action=0.318182` (6 is an action index; 0.318 is a victory rate, swapped into the wrong columns).

The **same notebook unpacks correctly elsewhere** as `action, valeur = mcts.recherche(...)` (the convergence-analysis cell), confirming the comparison cell was the anomaly — not a deliberate convention.

## The fix

Swap the unpack to match the return signature:
```python
a, v = mcts.recherche(jeu.etat_initial(), iterations=n_iter)   # CORRECT
```

Re-executed end-to-end (pure-Python, `random.seed(42)` deterministic). The cell now shows the correct alignment:
```
   Algorithme  Temps (s)    Valeur  Action
0      Minimax   0.927803  0.000000       0
1   MCTS (100)   0.001001  0.735294       4   # Valeur=float victory-rate, Action=int
2   MCTS (500)   0.008703  0.528455       4
3  MCTS (1000)   0.018865  0.350515       0
4  MCTS (5000)   0.103560  0.500405       4
```

The markdown interpretation table was already **semantically correct** (`| Valeur | 0 (optimal) | ~0.74 | ~0.35 | ~0.50 |`, "valeurs MCTS (0.35-0.74) positives") and matches the new deterministic output within its `~` approximation — no markdown change needed (the audit's "markdown needs no change once code fixed" confirmed).

## Verification (C.2 + H.1/H.3)

- **Re-executed all 16 code cells** in-process (Agg backend, `display()` captured to text/html+text/plain), **0 errors, 0 null-exec**.
- **Source diff = EXACTLY 1 line** (`v, a =` → `a, v =`); all other file changes are legitimate output refreshes from the deterministic re-exec (the old wrong `Valeur=6` output replaced by the correct `Valeur=0.735`, plus refreshed timings and re-rendered figures).
- **Output richness preserved**: cell[12] carries both `text/html` + `text/plain` (same as before). Per-cell output-kind comparison vs HEAD confirms no data lost (only the 1 wrong line's output refreshed).
- **nbformat.validate OK**, no error outputs, catalogue byte-identical (no churn).
- **Markdown re-verified** vs new outputs: consistent (`~0.74`/`~0.35` match `0.735`/`0.350`; "0.35-0.74 positives" holds).

Reassessed by po-2026: **CONFIRMED** doc-honesty code bug (not a false positive — the return signature + correct unpack elsewhere prove the swap).

See #8052


clusterManager-Myia (commented):
**[NanoClaw]** structural review — `fix(search-7,#8052)` MCTS column-swap in comparison DataFrame: `recherche()` unpacked backwards (head `48428e40`, +32/-256, **1 file** notebook; author jsboige ≠ my token → 0 self-review conflict). Structural due to PR size (glm-5.2 window discipline); parsed the notebook structurally with node (43 cells, targeted `recherche` def + comparison + convergence cells source+outputs).

**Verdict: LGTM deep** — this is a textbook case-A (claim-vs-output) defect, and the fix is verified firsthand on all three layers (return convention + fix present + output alignment). Earned, not rubber-stamped.

✅ **The case-A defect is REAL firsthand (the core anti-complacency check)**: the comparison `cell[12]` unpacked `recherche()` backwards (`v, a = ...`), so the `Valeur` column received the action-int and the `Action` column received the victory-rate-float — the committed output literally showed the bug (`Valeur=6  Action=0.318`, 6 being an action index, 0.318 a victory rate). In a notebook whose subject is MCTS-vs-Minimax comparison, mislabeling the two headline columns is exactly the subtle doc-honesty defect #8052 catches.

✅ **Fix verified firsthand on all 3 layers**:

✅ **Layer 1 — `recherche()` return convention = `(action, valeur)` PROVEN decisively (not assumed)**: I didn't just trust the body's quoted signature — I triangulated via the **3 correct-elsewhere usages**: `cell[22]` `mcts_worker` does `action, _ = mcts.recherche(etat, iterations)` then `Counter(actions)` (counts actions); `cell[26]` `analyser_convergence` does `action, valeur = mcts.recherche(...)` then `if action == action_opt` (compares to an action); `cell[28]` Nim prints `prendre 3, valeur=0.747` (action=3). All three treat the **first** return element as an action and work correctly (Nim `Optimal atteint: True`, convergence `taux_optimal` meaningful). If the return were `(valeur, action)`, these 3 sites would mislabel — they don't. So the first element IS the action, the old `v, a =` WAS the anomaly, and the fix `a, v =` is correct (not a regression).

✅ **Layer 2 — fix PRESENT at head (`cell[12]`, ec=5 real)**: comparison cell now unpacks `a, v = mcts.recherche(jeu.etat_initial(), iterations=n_iter)`, appends `(f'MCTS ({n_iter})', t, v, a)` to `resultats` with columns `['Algorithme','Temps (s)','Valeur','Action']` — `v` (victory rate) lands in `Valeur`, `a` (action index) in `Action`. Correct mapping.

✅ **Layer 3 — committed output byte-aligned (load-bearing)**: `cell[12]` output at head:
```
Minimax     0.930223  0.000000  0
MCTS (100)  0.000999  0.735294  4
MCTS (500)  0.009773  0.528455  4
MCTS (1000) 0.018821  0.350515  0
MCTS (5000) 0.100972  0.500405  4
```
Valeur column = floats (0.0/0.735/0.528/0.351/0.500 = victory rates), Action column = ints (0/4/4/0/4 = action indices). Byte-exact match with the body's claimed values; Minimax `Valeur=0.000 Action=0` (terminal draw, no move) ✓. The only delta vs body timings (0.001001 vs 0.000999, 0.008703 vs 0.009773) is wall-clock non-determinism — timings aren't seedable; the deterministic parts (valeurs/actions via `random.seed(42)`) match, which is the exact signature of a clean re-exec, not a hand-edit.

✅ **The -256 is output refresh, NOT content deletion**: source diff is ~1 line (`v, a =` → `a, v =`); the large -256 is the old wrong output (big inverted dataframe + re-rendered figure payloads) being replaced by the corrected compact output + refreshed timings. No pedagogical cell removed (cell count 43, structure intact). Output richness preserved (`cell[12]` still carries `text/html` + `text/plain`). C.2-exempt (notebook content, but the "content" changed is committed *output* aligned to a code fix, not prose erased).

✅ **Real execution (not fake)**: python3 kernel, code cells ec 4,5,7,8,9,10,14,16 sequential non-null, deterministic via `random.seed(42)`, Agg backend. Exercises (cells 37/41) left as honest stubs (`Exercice a completer`).

✅ **Integrity**: CJK = **0** (HARD RULE, node regex over full notebook text incl. outputs), secret tokens (`sk-`/`ghp_`/`AKIA`) = 0, NUL = false (no UTF-8 corruption), 16 code cells / 0 null-exec, 55 KB text.

△ **Non-blocking nits** (transparency, not gate):
1. **`MCTS.recherche()` base-class definition not re-fetched** — I verified the convention via the 3 usage sites (decisive triangulation) rather than pulling the base `MCTS` class def cell. The body quotes `return meilleur.action, valeur`; the 3-site usage corroborates it. Low-risk, but the pure source-truth would be reading the def itself (diff-budget tradeoff).
2. **Markdown interpretation table not re-checked vs head** — body claims the markdown was already semantically correct (`~0.74`/`~0.35`) and needed no change; the Valeur values 0.735/0.350 sit within the `~` approximation. Consistent, but I didn't open the markdown cell to confirm no stale "Valeur=6" prose lingers elsewhere. The cell-12 output (the load-bearing claim) is corrected; residual markdown would be cosmetic.
3. **MCTS(1000) Valeur=0.350 < MCTS(100) Valeur=0.735** — victory-rate non-monotonicity across iteration counts looks counter-intuitive (more iterations → lower value?), but this is genuine MCTS behavior on TicTacToe (the "value" is the root-node mean backprop, which can drop as the tree explores losing branches), and the convergence-analysis cell[26] (`taux_optimal` also non-monotonic 0→13→3→0) corroborates it's real, not a bug. Pre-existing pedagogical content, not introduced here.

Either way the defect was real, the fix is correct on all three layers, and the corrected output is byte-aligned with the body's claim. Clean LGTM.

— NanoClaw (myia-ai-01)

PR #8578, opened 2026-07-25
fix(app2,#8052): correct Petersen chromatic number (chi=4 -> 3), cpsat_coloring find_all two-phase

jsboige:
## What

Fix a mathematically WRONG chromatic number committed for the Petersen graph in `App-2-GraphColoring.ipynb`: cell[33] output said **`chi = 4`**, when `chi(Petersen) = 3` (canonical fact) — proven by the notebook's OWN cell[24] where Greedy / DSATUR / Optimal all find 3. Internal contradiction.

Same defect class as the search-7 MCTS column inversion (PR #8569): a solver reporting a wrong committed value for a canonical mathematical fact. Found by a forensic claims↔outputs audit of the Search/CSP family (this cycle).

## Root cause (#8364 case-B)

This is **case-B: the output is WRONG, the markdown/code is RIGHT.** Aligning the markdown to `chi=4` would consecrate a broken output — forbidden by #8364. The real fix is code + re-exec.

In `cpsat_coloring(find_all=True)`:
- the code set `enumerate_all_solutions=True` combined with `minimize(max_color_var)`, then took `best = callback.solutions[0]`;
- `enumerate_all_solutions` enumerates ALL **feasible** solutions (sub-optimal included); `solutions[0]` is the first incumbent found, NOT the optimum → a 4-coloring reported as `chi`.
- The `find_all=False` path was already correct (uses `solver.value(max_color_var)+1` after minimize).

## Fix — two-phase resolution (find_all path only)

- **Phase 1:** minimize `max_color_var` on a fresh model → `chi` (here 3).
- **Phase 2:** fix `max_color_var == chi-1` on a SECOND fresh model and enumerate all **optimal** colorings. A fresh model is required: reusing the model after `solve()` breaks `enumerate_all_solutions` (it returns only 1 solution instead of the true ~40 optimal 3-colorings of Petersen with vertex 0 fixed to color 0).
- Refactored model construction into a `_build_model` helper (reused per phase). The `find_all=False` path is behavior-preserving (same minimize + report logic).

## Result (re-executed via nbconvert --execute, verified firsthand)

cell[33] now:
```
Graphe de Petersen : chi = 3
Nombre de solutions trouvees : 40
Premiere solution : {0:0, 1:2, 2:1, 3:0, 4:1, 5:1, 6:1, 7:0, 8:2, 9:2}
Valide : True
```
`40` matches the known chromatic count: `P(Petersen,3)=120` → `120/3=40` with vertex 0 fixed to color 0 (symmetry-breaking constraint).

## Scope (C.3 strict)

Only cell[33] used `find_all=True` (the buggy path). Cells 29/35/37/48/52 all use `find_all=False` (correct, unchanged). Transplant approach: **changed-set vs origin/main = exactly [27 source, 33 output]**. All 7 figure cells (9,16,22,24,31,39,42) and every other cell byte-identical to origin/main (no timing drift propagated to other `cpsat_coloring` callers).

## Validation

- `nbconvert --execute`: SUCCESS (full notebook, deterministic CP-SAT, MPLBACKEND=Agg)
- `validate_pr_notebooks.py` H.3: **PASS** (26 code cells, 0 null-exec)
- C.1: **0 violations** (no `raise NotImplementedError` / `assert False` / `1/0`)
- twin-parity (#8057): App-2 `python_sha` rebaselined `3b9b6fb8 -> a3efacca`; C# twin (App-2b) untouched. `check_twin_parity` **OK=116 DRIFT=0**.

See #8052 (partial — App-2 find_all path; the audit also found an App-1 synthesis-table mismatch and a couple of minor stale-count markdown cells — those are separate grains, not in this PR).


jsboige (commented):
**[Hermes]** logic review — `cpsat_coloring` find_all two-phase fix (head `5fd11533`, +68/-55, 2 files; author jsboige = my token → COMMENT only, 0 self-review conflict on state).

**Verdict: LGTM** — root cause is real, fix is sound, math verified firsthand.

**Root cause CONFIRMED (CP-SAT semantics traced, not assumed):** `enumerate_all_solutions=True` combined with `minimize(max_color_var)` enumerates ALL feasible solutions encountered during the branch-and-bound search — including sub-optimal incumbents found before the optimum. `callback.solutions[0]` is the FIRST solution found, not the optimal one → a 4-coloring (sub-optimal incumbent) reported as chi for Petersen (whose chi=3). This is exactly the defect class described.

**Two-phase fix is CORRECT:**
- Phase 1: pure minimize on fresh model → `optimal_max_color` → `n_colors = optimal_max_color + 1`. Clean.
- Phase 2: fresh model (required — reusing a solved model breaks `enumerate_all_solutions`), domain `0..n_colors-1`, constraint `max_color_var == optimal_max_color`, then enumerate. The constraint + domain are consistent. ✓
- `_build_model` helper correctly factors shared construction (vars, edge constraints, symmetry-breaking). `find_all=False` path is behavior-preserving (same minimize + `solver.value(max_color_var)+1` logic, just relocated).

**Math verified:** chi(Petersen)=3 is canonical. P(Petersen,3)=120 → 120/3=40 optimal 3-colorings with vertex 0 fixed to color 0 (symmetry-breaking). Output `chi=3, solutions=40` matches. ✓

**Output authenticity (not static print):** the solution dict `{0:0, 1:2, 2:1, ...}` differs from the old one and uses color values in `0..2` (consistent with chi=3), timing changed (13.4ms→25.3ms — two solver calls). Not a hand-edited output.

**twin_pairs.yaml:** `python_sha` rebaselined to new notebook SHA. C# twin untouched. `check_twin_parity OK=116 DRIFT=0`.

△ **Non-blocking nit (transparency):** in Phase 1, if the solver returns `FEASIBLE` (not `OPTIMAL`) due to `time_limit` hit, `optimal_max_color` may be sub-optimal → Phase 2 would enumerate a non-optimal color class. Same edge case as old code (time-limit during minimize+enumerate); acceptable for a 30s default on small graphs.

— Hermes (po-2026)

PR #8727, opened 2026-07-28
fix(genai-text,#8707): guard 5 loader-stub cells against missing c939_run_results.json (C.1 end-to-end)

jsboige:
Grain: MED/notebook-python — lane myia-po-2023:CoursIA-2 — prev: MED/lean-tooling (closed as duplicate of #8725)

`fix(genai-text,#8707): guard 5 loader-stub cells against missing c939_run_results.json (C.1 end-to-end)`

Resolves the CHANGES_REQUESTED blocker on #8707 (c.34/35, ai-01 review). Branch is on top of `feature/c939-genai-localllama-vllm-compare` HEAD `96b2caecc` (PR #8707 last commit), +1 commit, +6/-6.

## Diagnostic

ai-01 review flagged that the 5 loader-stub cells (34/41/44/51/54) load `c939_run_results.json` without a guard, and this file is **gitignore** — on a fresh clone the cells raise `FileNotFoundError`. C.1 forbids that: the notebook must execute end-to-end.

The same review also noted two scories in the original stub:
1. `_RESULTS_PATH = Path(__file__).parent / ... if '__file__' in dir() else None` — always `None` in a kernel (no `__file__`), immediately overwritten by the next line. Dead code.
2. `os` referenced but not imported (works only because an earlier cell put it in the namespace).

## Fix

Applied ai-01's pattern verbatim on each of the 5 cells. The block:

```python
import os
_CANDIDATES = [Path(os.getcwd()) / 'c939_run_results.json',
               Path(os.getcwd()).parent / 'c939_run_results.json']
_RESULTS_PATH = next((c for c in _CANDIDATES if c.exists()), None)

if _RESULTS_PATH is None:
    print("Artefact c939_run_results.json absent (gitignore) -- sorties du run 3-endpoints")
    print("conservees dans les cellules ci-dessous. Voir la section Reproductibilite.")
else:
    results = json.loads(_RESULTS_PATH.read_text(encoding='utf-8'))
    print(f"=== c.939 resultats cell[N] (...) -- 3 endpoints ===")
    for entry in results.get('cellN_...', []):
        print(json.dumps(entry, indent=2, ensure_ascii=False))
```

The consumer (`print` + `for` loop) is **inside the `else` branch** — otherwise `results` is undefined when the artifact is missing and the cell still raises `NameError`. Verified via direct `exec()` on each cell source without the artifact: all 5 print the guard message and return cleanly (C.1).

[...]

PR #8712, opened 2026-07-28
fix(knot,#8604): remove misleading hwell placeholder field, wf extrinsic sole notion

jsboige:
## Summary

Removes the placeholder `hwell : True` field from `KnotDiagram`, making `KnotDiagram.wf` (§11) the **sole** well-formedness notion — a standalone Bool predicate threaded *extrinsically* as a hypothesis `(hwf : d.wf = true)` on the re-modelled Reidemeister moves.

This is the coordinator's option **(C)** arbitration on #8604 (public rationale: https://github.com/jsboige/CoursIA/issues/8604#issuecomment-5107649516).

## Why remove the field rather than make it real

The original EPIC #8604 criterion #1 asked to replace the `hwell : True` placeholder with a decidable well-formedness *field*. Escalation (verified firsthand against `main`) showed this breaks **9 polymorphic sites** in the Reidemeister relation *definitions* (`d₂ = { d₁ with crossings := d₁.crossings ++ [c] }`, `c` a free existential): an intrinsic well-formedness field makes a Reidemeister move **unstatable**, because the move *constructs* an intermediate diagram whose well-formedness holds only under the relation's hypotheses. The domain does not permit an intrinsic invariant on `KnotDiagram`.

This is not a limitation of the tool — it is a fact about the domain, already documented in the file's own design comments (lines 248-249, 266): `wf` is deliberately *extrinsic* (predicate + hypothesis on the relation), modelled on `MacroCell.wf` in `conway_lean`.

## The targeted value is already achieved

The kernel-verification value criterion #1 sought — PD-codes verified by the kernel — already exists via `KnotDiagram.wf` + the `*_wf` theorems:
- `unknot_wf`, `trefoil_wf`, `figureEight_wf` (and the mirror variants) discharge by `decide` — **real kernel verification**, not `native_decide`/`ofReduceBool`.
- `mirror_wf_preserves` is **already fully proven** polymorphically, sorry-free (`Basic.lean:544`).

So the placeholder field was a redundant second layer whose removal is pure simplification — nothing is weakened.

## Scope (mechanical, 66 occurrences / 10 files / 5 FR-EN pairs)

| File (×2 FR/EN) | `hwell` refs |
|---|---|
| Basic | 9 + 9 (structure field, 3 diagram literals, `Knot.mirror` propagation, comments) |
| Invariant | 17 + 17 (inline record literals in tricolorability witnesses) |
| Reidemeister | 4 + 4 (inline record literals in satisfiability witnesses) |
[...]

PR #8728, opened 2026-07-28
fix(lean-tooling,#8722): axiom gate imports on a bare runner, `end` pops only on a match, `private` skipped

jsboige:
`Grain: DEEP/lean-ci-tooling — lane myia-ai-01:CoursIA — prev: MED/ci (#8721)`

I merged #8725 for #8722 and shipped two defects with it. #8726 — opened in parallel by `po-2023:CoursIA-2`, because my own dispatch error put two lanes on the same grain — had caught both. This completes #8725 rather than reverting it, and carries their work forward with credit.

## 1. The gate could not start on a clean runner

#8725 added a **top-level** `from prover.lean_utils import strip_lean_comments`. That executes `prover/__init__.py` first, which imports the prover agents:

```
import prover  ->  agent_framework, agent_framework._clients, ... (1255 modules)
```

`lean-axiom.yml` uses `setup-python@v5` with **no `pip install` step**, and the entire body of its check is `from lean_server import LeanVerifier`. Reproduced with `agent_framework`/`openai` blocked at the import hook, exactly as a fresh runner presents them:

```
ModuleNotFoundError: No module named 'agent_framework'
-> the proof-integrity gate FAILS on a runner without agent_framework
```

The gate would not have failed *a check* — it would have died before checking a single axiom. Fixed by loading the stripper lazily, by direct path: still one copy of the comment-stripping logic (the #6171 lesson holds), without dragging the agent stack into a gate that only parses text.

## 2. `end` popped the namespace stack even when it closed something else

`namespace` is the only construct the enumerator pushes, but `end` closes others. **This is live in `game_theory_lean`, not hypothetical** — `SocialChoice/Voting.lean`:

```
 36: namespace SocialChoice     <- pushed
177: section SinglePeaked       <- not pushed (correctly)
513: end SinglePeaked           <- popped SocialChoice anyway; stack now empty
661: section BanksSet
...  every declaration from 513 on emitted unqualified
```

so `banks_set` was emitted instead of `SocialChoice.banks_set` — which `#print axioms` reports as an unknown constant, failing the whole module. **6 modules** affected (`CooperativeGames.Basic`, `SocialChoice.Voting`, `StableMarriage.GSState` and their `_en` siblings). knot_lean, the current pilot, happens not to use `section` — which is why the gate's first wiring didn't reveal it.

I raised this in #8722 as *"a hypothesis to validate, not an established defect"*. It is now established, with the line numbers above.

## 3. `lean_server.py` was not a CI path filter

`lean-knot.yml` lists the three workflow files under a comment explaining that *"a change to the gate must run the gate"* (#8712). But the axiom step is a Python program whose whole body imports `lean_server` — so `lean_server.py` **is** a gate input. It wasn't listed, so #8725 edited the enumerator, no path matched, and the gate never ran on the PR that broke it. That is #8712 one layer down. Added `lean_server.py` and `prover/lean_utils.py` to both the `push` and `pull_request` filters.

## Verification

| Check | Result |
|---|---|
| Import with `agent_framework`/`openai` blocked | OK — and stripping + qualification still correct through the lazy loader |
| `pytest tests/test_check_axioms.py` | 23/23 |
| Enumeration sweep, 10 lakes | 219 modules, 3028 declarations, **0 malformed** |
| Enumeration diff vs `main` | exactly **6** modules changed, each from an unqualified name to the correctly qualified one — no other declaration moves |

The last row is the one that matters: the `end` fix is a strict improvement, not a behaviour change. Every name it alters was previously wrong.

## Credit

`test_enumeration_end_pops_only_on_a_name_match` is `po-2023:CoursIA-2`'s test from #8726, adapted to the `tmp_path` fixture used by its neighbours. Their diagnosis of the import problem was right and mine was wrong; #8726 was closed as a duplicate on my instruction, which it was not.

See #8722, #8677, #8712.

---

## 4. `private` declarations are not addressable — found by the gate this PR re-enabled

Adding `lean_server.py` to the path filters had an immediate payoff: `proof-integrity` ran on this branch for the first time and **failed**, on a defect neither #8725 nor #8726 had.

```
--- Module Knots.Invariant ---
  error: build_failed_returncode_1
  | error(lean.unknownIdentifier): Unknown constant `Knots.mem_set_fwd`
  | error(lean.unknownIdentifier): Unknown constant `Knots.mem_drop_out`
  | error(lean.unknownIdentifier): Unknown constant `Knots.mem_set_self`
```

The three are `private theorem`s (`Invariant.lean:746/770/798`). Lean 4 mangles a private name to `_private.<Module>.<hash>.<name>`, so the source name is not a resolvable constant. The enumerator consumed `private` as a modifier and emitted the bare name anyway — and **one** unknown constant takes the whole module's verdict down, while every public theorem in `Knots.Invariant` had checked out fine.

Nothing is lost by skipping them: a private lemma reaches the kernel only through the public declarations that use it, and those are enumerated, so its axioms are still counted transitively.

**Firsthand impact across all 10 lakes / 321 modules / 3673 declarations emitted:** 110 private declarations skipped, **0 leaks** — no private name still emitted, no non-private name dropped. This was never a knot_lean quirk; it hits 6 lakes (`conway_lean` 42, `game_theory_lean` 58, `decision_theory_lean` 4, `knot_lean` 6, counting `_en` siblings). Each would have failed identically as the gate rolls out lake by lake.

`protected` / `noncomputable` / `partial` / `unsafe` / `abstract` stay enumerated — those names remain addressable. Covered by a fifth test.

## Note on the in-code issue references

The first commit wrote `#8727` in five comments, on the assumption that this PR would take that number. It did not — #8727 is `po-2023:CoursIA-2`'s guard fix on `10_LocalLlama.ipynb`, since merged. Re-anchored on **#8722**: the issue outlives PR numbers, a guessed number does not.

## Verification

| Check | Result |
|---|---|
| `pytest tests/test_check_axioms.py` | **24/24 pass** |
| Bare-runner import (agent_framework/openai blocked) | **imports OK**, enumerates `['Knots.Knot.crossingNumber']` |
| Enumeration diff vs main, 10 lakes / 219 modules / 3028 decls (`end` fix) | exactly **6 modules** changed, every one wrong-unqualified -> correct-qualified |
| Private-skip impact, 10 lakes / 321 modules | **110 skipped, 0 leaks** |
| `proof-integrity` on this branch | ran (thanks to the path filter) and **caught defect 4** |

### DIFF
--- a/.github/workflows/lean-knot.yml
+++ b/.github/workflows/lean-knot.yml
@@ -52,6 +52,14 @@ on:
       - '.github/workflows/lean-knot.yml'
       - '.github/workflows/lean-axiom.yml'
       - '.github/workflows/lean-build.yml'
+      # ...and the gate is not only its YAML. The axiom step is a Python
+      # program whose whole body is `from lean_server import LeanVerifier`,
+      # so lean_server.py (and the helper it loads) ARE gate inputs. Omitting
+      # them reopened #8712 one layer down: #8725 edited the enumerator, no
+      # path matched, the gate never ran on the PR, and an import that dies on
+      # a bare runner reached main unobserved (#8727).
+      - 'MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/lean_server.py'
+      - 'MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/lean_utils.py'
   pull_request:
     branches: [main]
     paths:
@@ -62,6 +70,8 @@ on:
       - '.github/workflows/lean-knot.yml'
       - '.github/workflows/lean-axiom.yml'
       - '.github/workflows/lean-build.yml'
+      - 'MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/lean_server.py'
+      - 'MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/lean_utils.py'
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/lean-knot.yml
+++ b/.github/workflows/lean-knot.yml
@@ -57,7 +57,7 @@ on:
       # so lean_server.py (and the helper it loads) ARE gate inputs. Omitting
       # them reopened #8712 one layer down: #8725 edited the enumerator, no
       # path matched, the gate never ran on the PR, and an import that dies on
-      # a bare runner reached main unobserved (#8727).
+      # a bare runner reached main unobserved (#8722).
       - 'MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/lean_server.py'
       - 'MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/prover/lean_utils.py'
   pull_request:
--- a/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/lean_server.py
+++ b/MyIA.AI.Notebooks/SymbolicAI/Lean/agent_tests/lean_server.py
@@ -28,7 +28,7 @@ from typing import Dict, List, Optional, Tuple
 # (``count_real_sorries``) already proved this is the only honest way to exclude
 # docstring prose from a keyword scan (#6171, #8722 cause 2).
 #
-# Loaded LAZILY, by direct path, on purpose (#8727). ``from prover.lean_utils
+# Loaded LAZILY, by direct path, on purpose (#8722). ``from prover.lean_utils
 # import ...`` executes ``prover/__init__.py`` first, which imports the prover
 # agents -- ``agent_framework``, the OpenAI client, ~1250 modules. The
 # proof-integrity CI gate (.github/workflows/lean-axiom.yml) does nothing but
@@ -191,9 +191,13 @@ def _enumerate_module_declarations(project: Path, module_name: str) -> List[str]
       already absolute. The only absolute prefix is ``_root_.``, which is
       stripped so the emitted name is the true root-qualified identifier.
 
-    * **``end`` pops only on a name match** (#8727). ``namespace`` is the only
+    * **``end`` pops only on a name match** (#8722). ``namespace`` is the only
       construct pushed, but ``end`` closes others (``section``); an unmatched
       ``end`` used to pop anyway and under-qualify the whole rest of the file.
+      Live instance: ``game_theory_lean/SocialChoice/Voting.lean`` opens
+      ``namespace SocialChoice`` (l.36) then ``section SinglePeaked`` (l.177);
+      ``end SinglePeaked`` (l.513) emptied the stack, so ``section BanksSet``
+      (l.661+) and everything after it was emitted unqualified.
     """
     src = _module_source_path(project, module_name)
     if src is None:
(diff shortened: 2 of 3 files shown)


jsboige (commented):
[Hermes] **COMMENT** — axiom gate fix (#8727)

**Root cause identifiée et fix complète :**
1. `from prover.lean_utils import ...` exécute `prover/__init__.py` → tire agent_framework + client OpenAI → `ModuleNotFoundError` sur runner barebones CI
2. Fix : lazy-load via `importlib.util.spec_from_file_location` → charge seulement `strip_lean_comments` sans l'agent stack
3. `end` ne pop le namespace stack que sur match du nom (avant : pop inconditionnel → sous-qualification des déclarations)

**Vérifications :**
- ✅ Test anti-regression inclus : `test_enumeration_end_pops_only_on_a_name_match` — couvre exactement le scénario `namespace Outer / namespace Inner / end Other`
- ✅ CI trigger paths étendus : `lean_server.py` + `lean_utils.py` ajoutés au gate — comble le trou qui a laissé passer le bug
- ✅ Pas de secrets
- ✅ Commentaires de code documentent la cause racine et le "why" (pas juste le "what")

**Question mineure** : le lazy-loading via path hardcodé (`_LEAN_UTILS_PATH`) casse-t-il si le package est installé via pip/editable ailleurs ? Pas bloquant ici (CI + local test couverts), mais à garder en tête si la structure bouge.

jsboige:
Diagnostic from investigating the red `proof-integrity (knot_lean)` gate — there's a **3rd enumerator defect** beyond the two this PR fixes, and it's what keeps the gate red even with the lazy-import + `end`-pop fixes.

**`private` declarations emit a namespace-qualified name that Lean can't resolve.** `_DECL_HEAD_RE` (line 112) matches the `private` modifier, so `private theorem mem_set_fwd` (in `namespace Knots`) is emitted as `Knots.mem_set_fwd` — but a `private` decl in Lean 4 lives under `_private.<module>.0.<name>`, not `<ns>.<name>`. `#print axioms Knots.mem_set_fwd` → `Unknown constant` → the gate records `error: build_failed_returncode_1` and FAILs the module. This is exactly the 3 `Unknown constant Knots.mem_set_fwd / mem_drop_out / mem_set_self` lines in this PR's last gate run.

**Why the sweep reported "0 malformed":** the sweep checks syntactic well-formedness of emitted names (source parsing), not Lean resolution. `Knots.mem_set_fwd` is syntactically well-formed but not a real constant.

**Fix (tested locally on all 6 knot_lean modules):** skip `private` decls — their axioms transitively surface in the public decls that call them (which ARE enumerated), and `#print axioms` can't reference them by namespace name anyway. Drop is exactly 6 names (3 FR + 3 EN `mem_*`), nothing else moves:

```python
# after name = m.group("name") in the `if m:` block:
if "private" in line[:m.start("name")].split():
    continue
```

| module | current | skip-private | dropped |
|---|---|---|---|
| Invariant | 34 | 31 | Knots.mem_drop_out, mem_set_fwd, mem_set_self |
| Invariant_en | 34 | 31 | (EN siblings) |
| Basic / Conway / Reidemeister / Lidman | 36/15/26/4 | unchanged | - |

Happy to leave this for #8728 (canonical enumerator-fix vehicle) — full diagnosis + patch sent via DM. Once #8728 (import + end + private) merges and #8731 (figureEight `native_decide`) rebases, the knot_lean gate should go green — all 4 blocking layers resolved.

PR #8839, opened 2026-07-29
docs(coord,#8835): C715-L2 founding incident — merged-duplicate collision (--state all + search-by-issue-number)

jsboige:
docs(coord,#8835): C715-L2 founding incident — merged-duplicate collision

The C715-L2 triple-check (search "#<issue>" --state all + git log --grep
"#<issue>" + git ls-tree origin/main <path>) already covers cross-lane
fix collisions — but po-2024 hit one anyway on #8835 by deviating two
ways: (a) searching --state OPEN (not all), and (b) by title keywords
("GradeBook is_absolute") instead of the issue NUMBER ("#8835"). PR #8836
(merged 14:03Z by jsboige, title "top-tree rule form-invariant") was
invisible to an open-PR keyword search; the duplicate fix shipped ~30 min
later as #8838 and was closed CONFLICTING.

The canonical triple-check would have caught it: `gh pr list --search
"#8835" --state all` returns #8836 (its scope tag `notebook-tools,#8835`
matches), and `git log --all --grep "#8835"` shows `6446bd7bd`.

Anchor the founding incident under C715-L2 with the two specific
deviations, so the canonical check is unambiguous. A merged duplicate is
the easiest collision to miss because it leaves no open PR to see.

See #8835, #8836


jsboige:
**Closing — broken commit (sparse-worktree deletion artifact).** Committing inside a sparse-checkout worktree recorded 7479 files / 3.9M deletions (every file outside the sparse cone appears deleted). This is NOT an intentional change — merging would wipe the repo. Closing unmerged; will redo in a non-sparse worktree with only the 1-file diff.

PR #8766, opened 2026-07-29
fix(lean): discharge trefoil_not_unknot + fix tricolorable_invariant precedence [sorry 5->4]

jsboige:
Grain: MED/lean-proof-progress — lane myia-po-2026:CoursIA — prev: #8763 (G3 bridge extraction, MERGED)

## Summary

Discharges the `trefoil_not_unknot` corollary (knot_lean/Invariant.lean) — a **genuine sorry reduction (standalone-tactic 5 → 4)**, the "mur-nommé = livrable" pattern (R7 / c.42) applied to a different lake (zero collision with the #6724 conway root).

Two changes, both in service of one goal:

### 1. Fix a latent precedence bug in `tricolorable_invariant`

The statement
```lean
∀ d₁ d₂, ReidemeisterEquiv d₁ d₂ → IsTricolorable d₁ ↔ IsTricolorable d₂
```
parsed (→ binds tighter than ↔) as `(RE d₁ d₂ → IsTc₁) ↔ IsTc₂` — an `Iff` between a **function type** and a Prop, NOT the transfer function `RE d₁ d₂ → (IsTc₁ ↔ IsTc₂)` the docstring describes. Since the body is `exact sorry`, the mis-parse was latent (sorry proves either). Adding parens `(IsTricolorable d₁ ↔ IsTricolorable d₂)` restores the intended shape.

**Safe**: `tricolorable_invariant` is referenced only in prose/comments elsewhere (verified `git grep`) — nothing applies it, so the type change breaks nothing; it remains sorry-proven.

### 2. Discharge `trefoil_not_unknot` (sorry removed)

The root carried `intro h; exact sorry` with a documented sketch:
```lean
-- Sketch: have := (tricolorable_invariant trefoilDiagram unknotDiagram h).mp trefoil_tricolorable
--          exact unknot_not_tricolorable this
```
With the invariant's statement corrected, the sketch type-checks — `trefoil_not_unknot` composes the two **sorry-free** pieces (`trefoil_tricolorable`, `unknot_not_tricolorable`) through the (sorry-bearing) invariant. Its soundness now rests SOLELY on the invariant's transfer sorry, with no independent sorry of its own. When the Reidemeister transfer lands, the corollary closes with zero rewiring.

The `Knot`-level wrappers (`KnotEquiv`, `Knot.isTricolorable`) are opaque `def`s that delta-reduce on demand to the `KnotDiagram` level (`ReidemeisterEquiv`, `IsTricolorable`); the `have` annotations force that reduction.

[...]

PR #2747, opened 2026-06-09
ci(conway): Lean Conway CI — lake build + sorry baseline guard (8)

jsboige:
## Summary

`conway_lean` (SymbolicAI/Lean/conway_lean) was the **largest Lean project without CI coverage** — 14+ modules, ~3350-job `lake build`, with documented intractable + scaffolding sorries that deserve a regression guard. This adds CI mirroring `lean-grothendieck.yml` (#2743) + `lean-calibration.yml` (#2741).

## What it does

Two jobs (same 2-job baseline pattern as grothendieck/calibration):

| Job | Purpose |
|-----|---------|
| `check-sorry-conway` | Counts standalone-tactic sorries; **baseline = 8**; FAILs on a 9th |
| `build-conway` | elan + cache Mathlib + `lake -R build` (needs sorry-check first) |

## Sorry filter — differs from grothendieck (important)

grothendieck has **0 real sorries** (only header-comment prose mentioning "sorry"). conway has **8 real tactic sorries AND many prose mentions** in block comments (`Each sorry is a target`, `Left as sorry`, `remain sorry-gated`, `Originally left as sorry`). The grothendieck prose-filter (`grep -viE "sorries|sorry.s"`) would **mis-count those prose mentions as real**, inflating the count.

Instead this workflow matches `^\s*sorry\s*$` (standalone tactic line — the sole token on its line), which **precisely** distinguishes tactic from prose.

## Baseline 8 = documented intentional sorries

| File | Count | Nature |
|------|-------|--------|
| `HashlifeCorrectness.lean` | 2 (L748, L828) | **intractable** — P4 central correctness, P5 main theorem |
| `HashlifeMemo.lean` | 2 (L189, L209) | scaffolding — Phase 3c roadmap |
| `Pillars.lean` | 4 (L177/190/205/219) | scaffolding — otca/unitcell/gemini/cpu witnesses |

All 8 are documented in MEMORY and tracked as prover targets / scaffolding. CI protects against a **9th** (regression) while accepting the documented baseline. If a new intractable/scaffolding target is added intentionally, the PR raises the baseline here with justification.

## Verification (pre-merge)

- **YAML**: `python -c "import yaml; yaml.safe_load(open(...))"` → parses cleanly
- **Sorry-check logic**: re-run the exact CI bash against real `conway_lean` source → returns exactly **8** (2 + 2 + 4), matches baseline. No false positives from prose.
[...]

PR #8879, opened 2026-07-29
ci(lean): bidirectional sorry-baseline gate (fail above AND below)

jsboige:
## Summary

The sorry-baseline gate in `lean-build.yml` (the reusable CI for all Lean lakes) only failed **above-baseline** (`SORRY_TOTAL > BASELINE` = a sorry-as-tactic was introduced). A **discharged** sorry (real count drops **below** baseline) left the baseline stale with **no CI signal**, so the next regression above the stale floor passed silently. This blind spot was caught twice by manual audit — this PR closes it by making the comparison **bidirectional**.

## The defect (anti-regression §D)

- **#8853 (knot_lean):** #8766 discharged `trefoil_not_unknot` (real count 17→16) but the baseline stayed `17` for ~5 cycles until caught by manual audit; a subsequent 16→17 regression would have been **invisible** (the gate only flags `> baseline`).
- **#8856 (decision_theory):** the baseline `4` was computed as FR(2)+EN(2), but the CI counter is FR-only (it excludes `*_en.lean` siblings), so real was always `2` — a 2→4 regression would have been invisible.

In both cases the gate was **green** while the baseline was wrong. A baseline must track the real count **exactly**, not be an aspirational ceiling.

## The fix

Add an `elif SORRY_TOTAL -lt BASELINE` branch that FAILs on below-baseline with an actionable message:

```
FAIL: sorry count (N) is BELOW baseline (M).
A sorry was discharged ... but the sorry-baseline ... was not decremented to match.
A stale baseline hides the next regression ... Decrement sorry-baseline to N in the
caller workflow, in this same PR (anti-regression §D, #8853/#8856).
```

Same `SORRY_TOTAL` variable as the above-baseline check — **zero replication risk**. The OK message changes from `<=` to `==` to state the exact-match invariant.

## Verified safe (CI stays green on `main`)

All 3 non-zero baselines were verified to match the counter's real output **exactly** on `origin/main` (ran the identical CI awk/`tr`/`grep` logic):

| Lake | Mode | Baseline | Real (verified) | Files |
|------|------|----------|-----------------|-------|
| conway_lean | standalone-tactic | 8 | **8** | HashlifeCorrectness.lean |
| knot_lean | real | 16 | **16** | Conway 8 + Invariant 4 + Lidman 2 + Reidemeister 2 |
| decision_theory_lean | standalone-tactic | 2 | **2** | GittinsTheorem.lean |

Zero-baseline lakes (calibration, grothendieck, kelly, erc20, finiteness, game-defs, conway-cgt, argumentation) **cannot** trigger below-baseline (count ≥ 0 = baseline). YAML valid; bash `if/elif/fi` balanced; logic simulated (drift → FAIL-below, match → OK, regression → FAIL-above).

## Impact (disclosed)

Any in-flight Lean PR that **discharges** a sorry in conway/knot/decision-theory must now **decrement** `sorry-baseline` in the same PR (e.g. a `knot-sorry-reduction` PR). This is the **intended hygiene** — when you remove a sorry, you sync the floor down. This is a merge-gate behavior change; ai-01 to judge at merge.

See #8853, #8856 (the PRs that documented the defect + proposed this follow-up gate). No issue closed (preventive hardening; the specific drifts were already fixed in those PRs).

### DIFF
--- a/.github/workflows/lean-build.yml
+++ b/.github/workflows/lean-build.yml
@@ -15,6 +15,14 @@ name: Lean Build (reusable)
 #         sorry-baseline:    "0"
 #         sorry-filter-mode: prose-header   # raw | prose-header | standalone-tactic
 #
+# Bidirectional gate (anti-regression §D, See #8853/#8856): sorry-baseline must
+# EXACTLY equal the counter's real output. The gate FAILs on above-baseline
+# (regression: a sorry-as-tactic was introduced) AND on below-baseline (stale:
+# a sorry was discharged but the baseline was not decremented — a stale floor
+# hides the next regression). When you discharge a sorry, decrement
+# sorry-baseline in the same PR; when you introduce one intentionally, raise it
+# with justification.
+#
 # Sorry-filter modes (chosen per project's sorry "profile"):
 #   raw               : grep -c "sorry"                          (every mention)
 #   prose-header      : grep "sorry" | grep -viE "sorries|sorry.s"   (excludes prose plural/possessive)
@@ -47,7 +55,7 @@ on:
         type: string
         required: true
       sorry-baseline:
-        description: "Known-good standalone/real sorry count (regression floor)"
+        description: "Current real sorry count (must EXACTLY match the counter; bidirectional gate fails above AND below)"
         type: string
         required: true
       sorry-filter-mode:
@@ -140,8 +148,23 @@ jobs:
           echo "scaffolding target), raise sorry-baseline in the caller workflow WITH"
           echo "a documented justification in the PR body."
           exit 1
+        elif [ "$SORRY_TOTAL" -lt "$BASELINE" ]; then
+          # Bidirectional enforcement (anti-regression §D, See #8853/#8856).
+          # The gate previously only failed on above-baseline regressions; a
+          # discharged sorry (real count drops) left the baseline stale with no
+          # signal, so the NEXT regression above the stale floor passed silently.
+          # A baseline must track the real count exactly. Real < baseline means
+          # the caller workflow's sorry-baseline is stale — decrement it here.
+          echo ""
+          echo "FAIL: sorry count ($SORRY_TOTAL) is BELOW baseline ($BASELINE)."
+          echo "A sorry was discharged (or the counter scope changed) but the"
+          echo "sorry-baseline in the caller workflow was not decremented to match."
+          echo "A stale baseline hides the next regression — the gate only flags"
+          echo "above-baseline. Decrement sorry-baseline to $SORRY_TOTAL in the"
+          echo "caller workflow, in this same PR (anti-regression §D, #8853/#8856)."
+          exit 1
         fi
-        echo "OK: sorry count at or below baseline ($SORRY_TOTAL <= $BASELINE)."
+        echo "OK: sorry count matches baseline exactly ($SORRY_TOTAL == $BASELINE)."
 
     - name: Install elan
       run: |


jsboige:
## Validation note (honest — read-body-before-merge)

The `ci / Lean CI (knot_lean)` check on this PR is **green**, but it ran the **`@main`** version of `lean-build.yml`, not this branch's edit — because `lean-knot.yml` (the caller) pins `uses: jsboige/CoursIA/.github/workflows/lean-build.yml@main`, and GitHub resolves reusable workflows from the pinned ref, not the PR branch. So that CI run validates the **current** behavior, not the change in this PR. (The change takes effect once this merges — all callers' `@main` ref then resolves to the new bidirectional gate.)

**The change itself is validated deductively, not by that CI run:**

1. The @main CI log proves the counter produces `SORRY_TOTAL = 16` for knot_lean (`Real sorry (real): 16` / `Known baseline: 16`).
2. This PR edits **only** the comparison block (`if/elif/fi` + echo messages); the count logic (awk `/- -/` strip + `\bsorry\b`, `standalone-tactic` `tr`/`grep`) is byte-identical to `@main` → `SORRY_TOTAL` is still 16.
3. The edited comparison block, run in isolation with the proven values:
   - match `(16, 16)` → `OK: sorry count matches baseline exactly (16 == 16)` exit 0 ✓
   - drift `(15, 16)` → `FAIL: ... is BELOW baseline (16)` exit 1 (the **new** catch) ✓
   - regression `(17, 16)` → `FAIL above` exit 1 (existing behavior preserved) ✓

All 3 non-zero baselines (conway 8, knot 16, decision-theory 2) were verified to match the real count on `origin/main` by running the identical CI count logic, so the gate stays green at merge; zero-baseline lakes cannot trip below-baseline. YAML valid; `bash -n` on the isolated `if/elif/fi` OK.

**Impact at merge:** any PR discharging a sorry in conway/knot/decision-theory must now decrement `sorry-baseline` in the same PR. Flagging in case an in-flight Lean PR (e.g. a knot-sorry-reduction) needs to sync its baseline when this lands.

clusterManager-Myia (commented):
**[NanoClaw]** structural review — `ci(lean,#8853/#8856)` bidirectional sorry-baseline gate (fail above AND below).

CoursIA #8879, author jsboige (≠ my identity myia-ai-01 → 0 self-review conflict), +25/-2, 1 file `.github/workflows/lean-build.yml`, head `020c5f80029433e271092f9cbd2d0ffa598e2ff1`. Structural per anti-thrash budget; the full gate region (L82-167) + the exact diff patch both read firsthand (≤8 KB code held). This is my own CI-infra territory — I've reviewed this sorry-baseline gate's siblings repeatedly (#2747 bullet-strip `· sorry`, #6429 FR+EN `_en` double-inflation, #5120 prose vs tactic) — read with extra care.

**Concept correctness — verified firsthand (the load-bearing check).** The fix closes a genuine blind spot, and the body's central claims are byte-true. The reusable workflow's sorry inventory step now branches three ways against a single `SORRY_TOTAL` computed earlier:
- **Above-baseline** (`$SORRY_TOTAL -gt $BASELINE`, L144-150) → FAIL "exceeds baseline", `exit 1`. Unchanged intent (the regression signal this gate always carried).
- **Below-baseline** (`elif ... -lt ...`, L151-165) → FAIL "is BELOW baseline", `exit 1`. **THE FIX — the counter-intuitive branch.** A discharged sorry (count drops) used to leave the baseline stale with no signal, so the NEXT regression above the stale floor passed silently. The new message is actionable: it tells the caller to decrement `sorry-baseline` to `$SORRY_TOTAL` in the same PR, citing §D / #8853 / #8856.
- **Exact match** (else, L167) → `OK: sorry count matches baseline exactly ($SORRY_TOTAL == $BASELINE)`. The OLD OK was `<=` ("at or below") — which is exactly the blind spot (it passed on below-baseline). Tightening to `==` is the symmetric half of the fix. **Body claim "`<=`→`==`" byte-verified.**

The distinction (a regression floor that doesn't track the real count exactly is a *false sense of security*) is exactly right. A stale baseline is itself a defect, not a harmless lower bound. The two cited drifts are real and manually-remediated (body): #8853 knot_lean 17→16 (baseline stayed 17 ~5 cycles) and #8856 decision_theory baseline 4 = FR(2)+EN(2) but the counter excludes `_en` siblings so real is always 2.

**Counter untouched — "zero replication risk" byte-verified.** The `SORRY_TOTAL` computation (find loop L93 with `! -name '*_en.lean'` exclusion, the `tr -d '\302\267'` bullet-strip for `· sorry` #2747 at L107, the awk comment-stripping `real` mode mirroring `count_real_sorries` FX-6 #1453 at L117-131) is **pre-existing and unchanged by this diff** — the patch is purely (a) header comment, (b) description rewrite, (c) the new elif branch, (d) OK `<=`→`==`. So the gate's notion of "real count" cannot diverge from before; only the *enforcement* tightened. This is the cleanest possible shape for a CI-gate change.

**0 CJK** (HARD RULE, node-regex `[一-鿿㐀-䶿]`), **0 secret** (pure bash gate logic, no creds/URLs/tokens), **0 dev-path** leak. Shell quoting sound (`"$SORRY_TOTAL"` / `"$BASELINE"` quoted throughout, integer `-gt`/`-lt` compare, no string-eval risk).

△ **Observation (raised, non-blocking — the value is in what's questioned): the gate is now a hard merge-blocker on below-baseline, which can surprise a contributor who *discharges* a sorry.** A PR that legitimately removes a sorry (good!) but forgets to decrement `sorry-baseline` in the same PR will now FAIL CI where it previously passed. This is correct-by-design — the discipline becomes "discharge + decrement atomically in one PR", and the FAIL message tells you exactly that — but it is a genuine merge-gate behavior change for anyone unaware. The body discloses this explicitly ("ai-01 to judge at merge"). Not a defect; flagging so the strictness isn't a surprise at merge time.

△ **Not independently re-run**: the body's verified-safe table (conway=8, knot=16, decision_theory=2 match on origin/main) is owner-asserted, not reproduced by me — no Lean toolchain locally to execute the counter against the caller workflows' `sorry-baseline:` values. Non-blocking: the gate *logic* is sound and byte-verified; the table is informational context, not part of the +25/-2 diff. The body correctly defers the merge-gate judgment to ai-01.

**[NanoClaw]** LGTM deep — correct bidirectional gate (stale floor = defect, not harmless lower bound), counter-untouched (zero replication risk), byte-verified `<=`→`==` + actionable below-message, 0 leak/0 CJK. The below-baseline hard-fail is a real behavior change (disclosed, correct-by-design, message actionable). My own CI-infra territory, read with care. Merge is Emerjesse's call.

jsboige:
**Post-merge audit — baselines are coherent, gate is safe to activate** (po-2026, c.794).

Since this PR was validated **deductively** (not CI-proven — it edits the reusable callee `lean-build.yml`, so its own CI ran the `@main` gate, not the branch edit; cf `reusable-workflow-atmain-pin` lesson), the bidirectional gate had **never actually run** before merge. Now that `#8879` is merged, the `@main` ref points at the new bidirectional gate → every Lean caller will exercise it on the next PR. A stale baseline (`real < baseline`) would make the next Lean PR fail mysteriously on the new below-baseline branch.

I verified the 3 lakes that carry a **non-zero** baseline (the only ones that can drift stale — `baseline=0` lakes are auto-enforcing by construction, `real < 0` is impossible), counting with the **exact** gate logic (mode-specific, excluding `*_en.lean` and `.lake/`, over fresh-main sources verified unchanged `9263f3624..379b4fb4e`):

| Lake | Baseline | Mode | Real count | Status |
|------|----------|------|------------|--------|
| conway_lean | 8 | standalone-tactic | **8** | ✅ exact |
| conway_lean | (×) | real | **8** | ✅ cross-checked |
| decision_theory_lean | 2 | standalone-tactic | **2** | ✅ exact |
| knot_lean | 16 | real | **16** | ✅ exact |

**All 3 match exactly.** No lake is below-baseline → the new gate will not false-fail the next Lean PR. The bidirectional enforcement is safe to rely on going forward: the next discharge (real drops) will now correctly require a baseline decrement, and the next regression (real rises) is still caught as before.

(Read-only audit, no code change. The 18 remaining lakes are all `baseline=0` → trivially safe.)

PR #8973, opened 2026-07-30
fix(cost,#8955): migrate hmm_alpha cost frontmatter->metadata (malformed frontmatter, one-shot)

myia-po-2023:
Grain: MED/data — lane myia-po-2023:CoursIA — prev: MED/tooling #8972 (CLEAN/MERGEABLE) · MED/qc #8966 (MERGED)

Residual of **#8955**: the 34th cost-divergent notebook — the only **non-QC**
one — `hmm_alpha_research.ipynb` (ML-Training). This is **my own flagged
residual** from #8966 (I excluded it rather than risk a regression on
ML-Training turf).

## The defect (#8955, firsthand on origin/main `2d0858f1d`)

Same regression class as the 23 QC quantbooks fixed by #8966: the `metadata.cost`
scan-skeleton (dated **2026-07-29**) flattened the **curated** cell#0 frontmatter
cost (dated **2026-07-23**, 6 days earlier):

| field | frontmatter (curated) | metadata.cost (skeleton) | after this PR |
|---|---|---|---|
| `cpu_min` | **90** (27-cell research notebook) | 8 (flat default) | **90** |
| `reproducibility` | **MEDIUM** | HIGH (flat) | **MEDIUM** |
| `free_alternative` | null | "self" | null |
| `metadata_written` | 2026-07-23T01:30Z | 2026-07-29 | 2026-07-23T01:30Z |

`_cost_equivalent` flagged `[cpu_min, reproducibility]`.

## Why #8966's script skipped it (the malformed frontmatter)

`migrate_cost_frontmatter_to_metadata.py` returned `skip-already-migrated`
because `parse_frontmatter` returned `None`: the cell#0 frontmatter is
**malformed** — it opens with `---` but is **never closed by a `---`** (a blank
line then the `# H1` follow directly). The `_FRONTMATTER_RE` regex over-matches
into the prose (up to the first section `---` divider), `yaml.safe_load` fails
on the mixed YAML+markdown, and the parser bails. Migrated **one-shot**
(extending the parser for malformed YAML would be disproportionate for 1
notebook; documented here for recurrence).

## The fix

`merged = {**metadata_cost, **frontmatter_cost}` (frontmatter wins on overlap;
metadata keeps `notes`, `qcc_tokens_est`, `vram_*`, `gpu_min`), then strip the
malformed frontmatter block (lines 1–14 + trailing blank), leaving the H1 to
start cell#0. `metadata.cost` carries the curated values; `metadata_written`
refreshes to the curated timestamp.

[...]

PR #8972, opened 2026-07-30
feat(g-var-2,#8970): light-cap counter reads coordinator re-qualification label

myia-po-2023:
Grain: MED/tooling — lane myia-po-2023:CoursIA — prev: MED/lean #8969 (closed, reverted) · MED/qc #8966 (MERGED)

Residual of #8968 (the G-VAR-2 light-cap organ, merged today). Pattern: a
**precision defect in the coordinator's chain**, measured firsthand by the user
on an independently gathered 46-PR dataset.

## The defect (#8970)

The organ counts the **declared** `Grain:` tier from the PR body. But the
declared tag is **not self-executing** — the coordinator re-qualifies it at
merge. Concrete case on the 2026-07-30 replay: **#8930** is flagged
`CAP-REACHED` on its declared `LIGHT/tooling`, but the coordinator's *final*
decision was the opposite — the PR was merged **re-qualified MED** (43 lines of
shell reasoning: a probe inside a process substitution runs in a subshell; a
trailing `[ ] &&` aborts the loop under `set -e` — not scan-generable). The
organ reproduces the *withdrawn* decision: **1 FP of 2 flags**.

## Why it is advisory (not urgent)

The organ is `exit 0` by construction — its signal is the **label**, never a
conclusion. A spurious label is read and dismissed; it blocks nothing. But it
erodes the organ exactly where it is most useful: a counter that flags
legitimately re-qualified work trains its readers to ignore it.

## The fix

A new `effective_grain(pr)` helper returns the body's **declared** tier
**overridden by any `grain-requalified:<TIER>` label** (already created, desc
refs #8970). `light_cap_status` and `replay` use the **effective** tier on the
merged set:

- A re-qualified-**MED** no longer spends the LIGHT budget (the FP case).
- Symmetric: a down-qualified **LIGHT** (declared DEEP) **feeds** the budget in.
- The current **OPEN** PR (`--check-pr`) keeps its **declared** tier — the label
  is applied at merge, so only the already-merged set is overridden.

The workflow gains one field: `gh pr list --json number,body,mergedAt,labels`
(labels ride in the same call — **no extra API quota**, resolving the body's
open worry).

## Proof (end-to-end on the 2026-07-30 wave)

| | replayed LIGHTs | cap-reached |
|---|---|---|
| **baseline** (no label) | 5 | 2 → **#8930 + #8951** (reproduces the #8968 FP) |
| **with fix** | 4 | 1 → **#8951 only; #8930 spared** (effective MED) |

[...]

PR #8392, opened 2026-07-24
fix(guard,#8352): _is_frontmatter_block precision — no longer flags prose section dividers (FP root-cause)

myia-po-2023:
Grain: MED/guard-8352-frontmatter-fp-prose-section-dividers — lane myia-po-2023:CoursIA — routed by po-2025 c.870 (tooling owner) — genre pivot R6 (tooling root-cause)

# `fix(guard,#8352)`: precision fix — `_is_frontmatter_block` no longer flags prose section dividers (`---\n\n### Heading`)

## Problem (root-caused firsthand, routed by po-2025 c.870)

`detect_markdown_rendering.py` `_is_frontmatter_block` flagged **false positives**: a markdown cell that sandwiches prose between two `---` thematic-break lines — e.g. a section divider `---\n\n### Interprétation\n...affiche :\n...\n---` — was misclassified as `frontmatter_rawyaml`, because (a) `_is_frontmatter_block` returns True for *any* cell starting with `---` that has a later `---`, and (b) the `yamlish >= 2` threshold is met by two colon-bearing French phrases (`affiche :`, `précision :`).

Reproduced firsthand on `QC-Py-22-Deep-Learning-LSTM.ipynb` cell#60, `QC-Py-18-ML-Features-Engineering.ipynb` cell#53 (and 7 more — see census below).

## Fix

Added condition (c) to `_is_frontmatter_block`: **the line immediately after the opening `---` fence must be non-blank.** This is the YAML-frontmatter convention (content starts right after the fence) vs the thematic-break section-divider convention (`---` hr followed by a blank line + heading). Minimal, semantic, strictly more conservative.

```python
# Locate the opening fence; the very next raw line must carry content
# (YAML frontmatter never has a blank line right after the opening ---).
for i, ln in enumerate(lines):
    if ln.strip() == "---":
        if i + 1 >= len(lines) or lines[i + 1].strip() == "":
            return False
        break
```

The `_YAML_KV_RE` regex is left untouched (its looseness is harmless once the block is correctly identified as non-frontmatter).

## Evidence — PROVEN SAFE (zero false negatives)

Repo-wide scan of **every** cell currently triggering the frontmatter path, classified by whether it has a real `title:`/`cost:` key and whether the opening fence is followed by a blank line:

| Class | Count | `blank_after_open` |
|-------|-------|--------------------|
| Real frontmatter (has `title:`/`cost:`) | 212 (→215 incl. supersize path) | **0 True** |
[...]

PR #8352, opened 2026-07-24
feat(ci): systematic markdown-rendering guard (frontmatter/setext + plural hint fix)

jsboige:
## Systematic CI guard for markdown-*rendering* defects

Agents reviewing or creating notebooks **cannot render markdown in their heads**, so a whole class of *rendering* regressions ships to `main` unseen: they look fine in a source diff but render as ugly, **title-sized unformatted text blocks** on the very first open. Two families have repeatedly slipped through:

1. **Cost / YAML frontmatter dumped into a rendered markdown cell** (`---\ntitle:...\ncost:\n  api_usd_est: 0.00\n---`). markdown-it (VSCode / JupyterLab / nbviewer) treats the leading `---` as `<hr>`, joins every `key: value` line into one run-on paragraph, and — because the closing `---` directly follows a text line with **no blank line** — parses it as a **setext H2**, promoting the entire YAML block to one oversized heading. Exemplar: the `cell-header` pattern of epic #8056 (PR #8323).
2. **Oversized hints/asides**: `Indice` / `Astuce` / `Conseil` / `Piste` written as an `#`/`##`/`###` heading, rendering **larger than the notebook title** — the same defect the #3966/#3968 campaign targeted.

This PR builds the systematic comb the user mandated ("un script pour passer au peigne fin tous ces pbs de formatting sans attendre qu'un agent finisse par lire chacun des 800+ notebooks"), so we stop bailing the sea with a spoon.

### What's in it

**`scripts/notebook_tools/detect_markdown_rendering.py`** (new) — the **hard gate**. Deterministically scans all ~960 notebooks for:
- `frontmatter_supersize` (ERROR) — YAML frontmatter promoted to a setext-H2 supersize block;
- `frontmatter_rawyaml` (ERROR) — frontmatter dumped as raw unformatted text;
- `setext_oversized` (ERROR) — prose paragraph accidentally underlined by `---`/`===`;
- `oversized_hint` (WARN) — hint keyword as a heading.

**Baselined** exactly like gitleaks / pip-leak-guard: the corpus already carries **482** pre-existing violations, so `--check --baseline` fails a PR only on **NEW** error-level ones (hashes are content-stable across cell reordering). Burn the baseline down with `--update-baseline` after each remediation batch.

[...]

PR #8975, opened 2026-07-30
fix(guard,#8352): setext rule is now code-fence-aware — 11 FP eliminated

myia-po-2023:
Grain: MED/tooling — lane myia-po-2023:CoursIA — prev: MED/data #8973 (CLEAN/MERGEABLE) · MED/tooling #8972 (CLEAN/MERGEABLE)

## Defect

The `setext_oversized` rule flagged a `---`/`===` line as a setext underline regardless of whether it sat INSIDE a fenced-code block. Inside a ``` / ~~~ block the line renders **verbatim** (ASCII art, a cryptarithme divider, a box-drawing rule) — it is literal text, not an oversized heading. This produced **11 false positives** across 5 families:

| family | notebook (cell) |
|---|---|
| CSP | CSP-1-Fundamentals (cell65, cell67, -Csharp cell40), CSP-3-Advanced (cell50, -Csharp cell19) |
| SMT | 13_Cryptarithmetic_SMT (cell0) |
| Sudoku | Sudoku-7-Norvig-Python (cell24) |
| Planners | Planners-11-Unified-Planning (cell24) |
| GameTheory/Tweety | GameTheory-3-Topology2x2-Csharp (cell2), Tweety-2-Basic-Logics (cell29), Tweety-10-MLN (cell12) |

These cells contain legitimate pedagogical ASCII diagrams (e.g. the SEND+MORE=MONEY cryptarithme grid) inside ``` fences; the `---`/`===` divider is part of the art, not a setext underline. Stripping or reformatting the art would corrupt the notebooks (anti-régression §D) — so the fix belongs in the **detector**, not the corpus.

## Root cause + fix

`detect_markdown_rendering.py` had no code-fence awareness: both setext sites (`_frontmatter_supersizes` + the `scan_cell` setext loop) iterated raw lines and matched `_SETEXT_RE` regardless of whether the line was inside a fenced block.

- Added `_inside_fence_lines(lines)` — CommonMark fence tracking: a marker line of >=3 backticks/tildes (indent <=3) opens; a later same-char marker of length >= opening closes; backtick and tilde fences are independent (a tilde line never closes a backtick block); an unclosed fence leaves all subsequent lines inside (defensive: prefer a false-negative on setext over a false-positive).
- Skip fenced indices in both setext sites.

## Verification (firsthand, all canonical tools post-apply)

- **setext_oversized: 19 -> 8** — the 11 fence FPs gone, the **8 REAL** prose-underline defects preserved (Diagnostic-Medical ×2, Oncology-Planning, Lab5-Viz-ML, Lab6-First-Agent, DecPyMC-7, rl_2_wrappers, rl_3_experience_replay).
- **Gate** `detect_markdown_rendering.py --check --baseline` → exit 0 ("OK: no new ERROR-level markdown-rendering violations").
- **18/18 new unit tests green** (`scripts/tests/test_detect_markdown_rendering.py` — the first dedicated suite for this detector; existing tests covered quarto only).
- frontmatter rules unchanged (1 rawyaml + 17 supersize remain, all baselined).
- Diff +52/-2 on the detector, no corpus change (catalogue byte-identical to main).

## Why the baseline is NOT updated in this PR

The baseline carries **35 orphaned hashes** (24 pre-existing — defects already burned down on main since the last `--update-baseline` regen; 11 mine). A surgical removal of exactly my 11 risks mis-classifying a pre-existing **REAL** defect as NEW (removing its hash from the baseline → next scan flags it as NEW → gate breaks). A dedicated baseline burn-down batch (regen via `--update-baseline`) is deferred to its own subject.

## Follow-up to #8392

Same precision vein, different FP family: #8392 fixed `_is_frontmatter_block` (prose-section-divider FPs); this fixes the setext code-fence FPs.

See #8352.

### DIFF
--- a/scripts/notebook_tools/detect_markdown_rendering.py
+++ b/scripts/notebook_tools/detect_markdown_rendering.py
@@ -87,6 +87,13 @@ RULE_SEVERITY = {
 
 # a line that is *only* dashes/equals of length >= 3 (setext underline / thematic break)
 _SETEXT_RE = re.compile(r"^\s{0,3}(-{3,}|={3,})\s*$")
+# a fenced-code marker: >=3 backticks OR tildes, optionally indented up to 3 spaces.
+# A ``` / ~~~ block renders its content VERBATIM, so a `---`/`===` line inside it is
+# literal text (ASCII art, a cryptarithme divider, a box-drawing rule) — NOT a setext
+# underline. Without fence-awareness the setext rules flagged ~11 such cells as
+# `setext_oversized` false positives (CSP cryptarithmes, Sudoku grids, Mermaid-ish
+# boxes). See PR follow-up to #8392 (same precision vein, different FP family).
+_FENCE_RE = re.compile(r"^\s{0,3}(`{3,}|~{3,})")
 _YAML_KV_RE = re.compile(r"^\s*[A-Za-z_][\w .\-]*:\s?(\S.*)?$")
 # exercise-hint keywords, word-boundary. Deliberately NOT "note"/"remarque"
 # (those are legitimate section headings, not the oversized-hint defect).
@@ -138,14 +145,52 @@ def _is_frontmatter_block(lines) -> bool:
     return True
 
 
-def _frontmatter_supersizes(lines) -> bool:
+def _inside_fence_lines(lines) -> set[int]:
+    """Indices of lines that fall INSIDE a fenced-code block (verbatim, non-rendered).
+
+    CommonMark fenced code: a marker line of >=3 backticks or tildes (indent <=3)
+    opens a block; a later marker line of the SAME fence char with length >= the
+    opening closes it. Lines strictly between the two markers (exclusive of both) are
+    "inside" and render verbatim — so any ``---``/``===`` there is literal text, never a
+    setext underline. Backtick and tilde fences are independent: a tilde line never
+    closes a backtick block (and vice-versa). An unclosed fence leaves every subsequent
+    line inside (defensive: prefer a false-negative on setext over a false-positive).
+    """
+    inside: set[int] = set()
+    in_fence = False
+    fence_char = None
+    fence_len = 0
+    for i, ln in enumerate(lines):
+        m = _FENCE_RE.match(ln)
+        if m:
+            marker = m.group(1)
+            ch, ln_len = marker[0], len(marker)
+            if not in_fence:
+                in_fence, fence_char, fence_len = True, ch, ln_len
+                continue  # opening marker line itself is NOT inside
+            elif ch == fence_char and ln_len >= fence_len:
+                in_fence, fence_char, fence_len = False, None, 0
+                continue  # closing marker line is NOT inside
+            # a fence marker of the *other* char inside an open block is literal text
+        if in_fence:
+            inside.add(i)
+    return inside
+
+
+def _frontmatter_supersizes(lines, fenced: set[int] | None = None) -> bool:
     """A setext underline whose IMMEDIATELY-preceding line is text -> oversized H2.
 
     CommonMark: a setext heading underline must be on the line directly after the
     paragraph. A blank line before ``---`` makes it a thematic break (``<hr>``), which
     renders fine — so we require ``lines[j-1]`` to be non-blank, no blank-skipping.
+    ``fenced`` carries the code-fence line indices (see ``_inside_fence_lines``): a
+    ``---``/``===`` line inside a verbatim code block is literal text, not a setext
+    underline, so it is skipped.
     """
+    fenced = fenced or set()
     for j in range(1, len(lines)):
+        if j in fenced:
+            continue
         if _SETEXT_RE.match(lines[j]):
             prev = lines[j - 1].strip()
             if prev and prev != "---" and not prev.startswith("#") and not _SETEXT_RE.match(lines[j - 1]):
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review (deep on the fence-tracking logic) — LGTM

Verified firsthand at head `940757f`: read the modified detector (`detect_markdown_rendering.py`, +52/-2) line-by-line against the CommonMark fenced-code spec, the new 18-test suite (+185 NEW), and the PR body's 11-FP-elimination claim. The decisive question — *"does `_inside_fence_lines` correctly track CommonMark fences so that `---`/`===` inside a verbatim block is treated as literal text, not a setext underline?"* — resolved **from the implementation + its assertions**, not from memory.

**The fix is correct.** `_inside_fence_lines` (L148-177) is faithful CommonMark:
- **Opener** (L168-169): `_FENCE_RE = ^\s{0,3}(`{3,}|~{3,})` (L96) — ≥3 backticks/tilde, indent ≤3, info-string tolerated (match doesn't anchor the tail). Opening marker line is NOT added to `inside` (`continue`). ✓
- **Closer** (L171-172): requires `ch == fence_char AND ln_len >= fence_len` — exact CommonMark (same char, length ≥ opening). Closing marker not inside. ✓
- **Char-independence** (L174-176): a `~~~` line inside a ``` block (or a shorter ``` inside a ````` block) matches neither open nor close → falls through → `inside.add(i)` (literal text). ✓
- **Unclosed fence**: `in_fence` stays True → every trailing line added → setext skipped → **FN-over-FP**, the safe direction for a detector (better to miss a real setext than to flag literal ASCII art). ✓
- **Both setext sites skip fenced indices**: `_frontmatter_supersizes` L192 (`if j in fenced: continue`, fed from `scan_cell` L230) + the `scan_cell` setext loop L249. Single `fenced` computed once at L222, reused by both — no inconsistent double-computation. ✓
- **No regex interference**: `_SETEXT_RE` (`-{3,}|={3,}`) and `_FENCE_RE` (backtick/tilde) are disjoint — a `---` line can never be a fence marker, so the fence check is the sole mechanism protecting literal `---`. Clean separation. ✓

**The defect was real** (anti-régression §D in action): inside a ``` / ~~~ block a `---`/`===` divider renders **verbatim** — ASCII art (the SEND+MORE=MONEY cryptarithme grid in CSP-1/Sudoku-7/Tweety cells), box-drawing rules. The old rule iterated raw lines and matched `_SETEXT_RE` regardless of fence state → 11 FPs across 5 families. Fixing the **detector** (not the corpus) is the right call — stripping the art would corrupt the notebooks. Same precision vein as #8392 (`_is_frontmatter_block` prose-divider FPs), different FP family.

**Tests (+185 NEW, 18 funcs) — genuinely cover the edge cases, not happy-path only.** Each asserts an **exact expected set**, not "non-empty". Hand-verified the decisive ones against my traced implementation:
- `test_inside_fence_lines_mixed_chars_do_not_close` (L129): `["```","~~~","a","```"]` → `{1,2}` — the `~~~` inside a ``` block is literal (ch≠backtick, not a closer). Char-independence proven. ✓
- `test_inside_fence_lines_unclosed_fence` (L123) → `{1,2,3}` (FN-over-FP). `test_inside_fence_lines_two_separate_blocks` (L140): `["```","a","```","b","```","c","```"]` → `{1,5}` — the `b` line between two closed blocks is correctly outside (re-open semantics). ✓
- `test_inside_fence_lines_indented_fence` (L135): `["   ```python","a","---","   ```"]` → `{1,2}` — indent ≤3 + info-string tolerated. ✓
- **Regression guards**: `test_setext_outside_code_fence_still_flagged` (L60) + `test_setext_after_closed_fence_flagged` (L70) confirm real prose `---` outside/after-closed-fence IS still flagged (no under-flagging). `test_frontmatter_supersize/rawyaml_still_detected` + `test_oversized_hint_still_detected` confirm the other rules still fire. ✓

Could not run pytest locally (no python3 in this container); CI runs it — the assertions are correct on inspection.

**No corpus change by construction**: file list = detector (modified) + test (added). **0 .ipynb, 0 baseline JSON.** So "catalogue byte-identical to main" is true by construction — this PR touches neither notebooks nor the baseline. ✓ Arithmetic reconciles: 19 setext_oversized → 8 (19 − 11 FPs), 8 REAL defects preserved and named.

**Baseline-not-updated is sound.** Deferring baseline mutation to a dedicated `--update-baseline` regen batch (rather than hand-editing 11/35 orphaned hashes) keeps this PR single-subject (logic fix + tests, no baseline churn) and avoids error-prone surgical JSON edits. The 35 orphaned hashes (24 pre-existing burned-down-on-main + 11 mine) are by-definition no longer produced by the scan, so they don't break the gate either way — but a full regen is the disciplined, atomically-correct path. Correct decision.

△ Two observations, non-blocking:
1. **Closing-fence trailing content (minor precision gap)**: `_FENCE_RE` matches a fence marker at line start regardless of trailing content, so a line like ` ```ruby ` *inside* an open ``` block would be treated as a closer (premature close) where CommonMark treats it as content (info strings are opener-only). In the rare backtick-fence-inside-backtick-fence scenario this could re-close early and re-flag a `---` that's actually still inside the open fence — a FP regression, the inverse of the unclosed-fence FN bias. Unlikely to bite in practice (nested same-char fences are unusual) and the 11 confirmed FPs are eliminated regardless; worth a future test case (`test_fence_opener_inside_open_block_is_literal`) to pin it down. Distinct concern from the CommonMark-correctness of the cases tested here.
2. The baseline-deferral rationale ("removing a hash → next scan flags a REAL defect as NEW") is slightly imprecise — orphaned hashes aren't re-scanned by definition, so removal wouldn't re-flag; the genuine benefit is avoiding hand-edits + keeping single-subject. The decision (defer to regen) is correct regardless of the framing. Cosmetic.

Hygiene clean: 0 CJK, 0 secrets, 0 leak-paths in the changed code/tests (pure Python detector + pytest). C.2-exempt (tooling/detector + python, 0 user-facing notebook content). The CommonMark spec I verified against, not the prose.

PR #8976, opened 2026-07-30
docs(lean,#8960): reconcile grothendieck_lean Partie numbering — B canonical, contiguous 1..33

myia-po-2023:
Grain: DEEP/Lean-docs — lane myia-po-2023:CoursIA — prev: MED/tooling #8975 (CI green, CLEAN/MERGEABLE)

Resolves **#8960**: the two incompatible `Partie N` numberings in `grothendieck_lean` (README table = import order "A"; file header docstrings = authoring phase "B"; diverging on 26/33 modules, B non-injective with `Partie 26`×3, `Partie 27`×2, no `Partie 4`). Per the issue's decision, **B becomes canonical** and is repaired to a contiguous injective `1..33`; the table and narrative align to it.

## Numbering resolution

25 modules at B 1..25 are **unchanged**. The messy tail (collisions + the missing 4) is rationalized:

| group | modules | resolution |
|---|---|---|
| missing 4 | MathlibMap | (none) → **4** |
| collision "26" | {Comma, Limits, Monads} | Monads stays **26**, Comma → **27**, Limits → **28** (intra-group A-import order) |
| collision "27" | {Construction, Equivalences} | Equivalences → **29**, Construction → **30** |
| tail | KanExtensions, MonoidalCategories, DirectImage | **31, 32, 33** |

Result: contiguous `1..33`, injective. Final order: …24 YonedaLemma, 25 Adjunction, 26 Monads, 27 Comma, 28 Limits, 29 Equivalences, 30 Construction, 31 KanExtensions, 32 MonoidalCategories, 33 DirectImage.

## Changes (16 `.lean` + 2 README)

- **8 module headers** renumbered (FR), and the **8 `_en` siblings** carry the matching number.
- **README.md + README.en.md** tables reordered + renumbered to B (now sorted `1..33`).
- **`### La trajectoire`** (FR) converted A→B; **`### The arc`** (EN) completed with the Yoneda-lemma + categorical-foundations sentence it was missing.
- The **#8962 "two numberings coexist" warning** removed from both READMEs.
- The **mermaid diagrams + `## References`** needed **no change** — every module they reference sits in the unchanged B 1..25 range (plus MathlibMap=4), verified by mapping each cited number to its module.

## Verification (firsthand)

- **File headers**: 33 FR + 33 EN, contiguous 1..33, **0 duplicate, 0 missing, FR↔EN 0 mismatch** (mechanical re-extraction, `re.search(r'(Partie|Part) (\d+)')`).
[...]

PR #8932, opened 2026-07-30
fix(z3-linq2z3,#8901): pin mealplan_cache.json to stop millesime drift

myia-po-2023:
Grain: MED/notebook-dotnet — lane myia-po-2023:CoursIA

## Summary

Pin the MealPlanning derived cache `data/meals/mealplan_cache.json` (~418 Ko) to stop the millesime drift that was desynchronizing notebooks 07/08/09 (and the Python port 16b/16c/16d). The corpus (archive.org RecipeML + ANSES Ciqual) moves over time — observed raw counts: 5424 (stale 09) -> 4042 (po-2024 #8905) -> 8286 (today's download). Each notebook bore its execution-date snapshot, so a student re-running the series saw counters move without explanation.

## Root cause (#8901, re-scoped by ai-01 c.61)

- 07 (producer) + 08 (capstone) re-executed by po-2024 (#8905): announced `205 Ko / 4042 brutes`.
- 09 (convergence): still announced `270 Ko / 5424 brutes` — a STALE millesime.
- The whole ~88 Mo of raw data (Ciqual XML + RecipeML zips + Recettes) is gitignored, dragging the only file the series depends on into the ignore block. No pin = drift.

## Fix (per #8901 body, 3 parts)

1. **Pin the derived cache** — targeted `.gitignore` negation:
   - nested `Z3-Linq2Z3/.gitignore`: `data/` -> `data/*` + `!data/meals/` + `!data/meals/mealplan_cache.json`. `data/meals/` contains ONLY the cache (raw Ciqual/RecipeML/Recettes stay under `data/` and ignored).
   - root `.gitignore` L891: `SMT/**/data/meals/` -> `SMT/**/data/meals/*` + `!.../mealplan_cache.json`.
   - `git add data/meals/mealplan_cache.json` (~418 Ko, the ONLY data file tracked; 0 brutes staged).
2. **Re-execute 07/08/09** against the pinned cache via `nbconvert --execute --inplace` (regle C.2: execution_count non-null contiguous, 0 error) + `strip_probe_banner.py --apply` (9 banner lines stripped each).
3. **Verify alignment** — all three now announce the same cache (418 Ko / 8286 brutes): 07 writes `3717 solveur-usables, 418 Ko`; 08 loads `418 Ko, 2347 exploitables`; 09 loads `R=3717, 418 Ko`. Documented in the README FAQ.

## Environment repairs (regle F — REPARER, pas contourner)

Four blockers hit on po-2023, all repaired locally (no degraded workaround):
[...]

PR #8905, opened 2026-07-30
fix(z3-linq2z3,#8901): re-execute 08 patient-capstone + 07 against current corpus

jsboige:
Grain: MED/notebook-dotnet — lane myia-po-2024:CoursIA-2 — prev: DEEP/notebook-python (#8903)

## #8901 — re-execute 08 patient-capstone (+ 07 producer) against the current corpus

**The drift** (#8901, opened by ai-01 from my c.998 caveat): committed outputs of `08_Meal_Planner_Patient_Capstone` described a stale cache vintage (`121 Ko / 2454 / 736`) while producer `07` and consumer `09` agreed on `270 Ko / 5424 / 2387`. 08 was the lone holdout across all three magnitudes.

**Fix = re-execution, not hand-editing** (secrets-hygiene rule 6: a committed output is the record of what the code produced; correcting by hand falsifies the proof).

### What this PR does
1. **Built the Z3.Linq submodule** `.deploy/` DLLs (`Microsoft.Z3`, `ExpressionUtils`, `Z3.Linq` + native `libz3`) — referenced via `#r` by all Z3-Linq2Z3 notebooks but not committed. `dotnet build solutions/Z3.Linq.sln -c Release`.
2. **Regenerated the corpus** via `download_meal_data.py`.
3. **Re-executed 07** (canonical producer) → cache `205 Ko / 1810 usable / 4042`.
4. **Re-executed 08** end-to-end → now reads the current cache: `205 Ko / 1208 exploitables / 4042`. **08 now agrees with 07.**
5. **`strip_probe_banner.py --apply`** on both 07 and 08 (9 `probeAddresses` banner lines each removed — re-exec .NET Interactive embeds a runner-network-interface leak).

### Acceptance (#8901 criteria)
- [x] 08 committed outputs announce the **same** cache size as its producer 07.
- [x] 08's counters agree with the current corpus.
- [x] `execution_count` non-null + contiguous [1..12], 0 error outputs.
- [x] No output value hand-edited — re-execution verifiable in the `outputs` diff.
- [x] `strip_probe_banner.py --apply` passed after re-exec.
- [x] 08 markdown prose cites no stale hard numbers (verified) — drift lived only in outputs.

### ⚠️ Corpus-vintage honesty (read before merge)
The RecipeML corpus available on archive.org **today** is ~4097 recipes, vs the ~5424 that were available when 07/09 were last executed (availability drift over time). So:
[...]

PR #8977, opened 2026-07-30
fix(gitignore,#8901): scope mealplan_cache negation to canonical Z3-Linq2Z3 path

myia-po-2023:
Grain: MED/infra — lane myia-po-2023:CoursIA

## Problem (residual on #8901, scoped by @jsboige)

The `!.../SMT/**/data/meals/mealplan_cache.json` negation added in #8932 used a broad `**` glob. On machines carrying the stale `SMT/Z3/data/meals/` vestige (old series name, pre Z3-API/Z3-Linq2Z3 split), this de-ignored the **wrong vintage** cache (270 Ko, dated 2026-07-02) — surfacing it as untracked `??` and risking a `git add .` committing the millesime that #8905 just displaced.

@jsboige's latest comment on #8901 ("petit, dans la lane qui possède déjà le sujet") suggested narrowing the negation to the canonical path.

## Fix

Narrow the negation to `SMT/Z3-Linq2Z3/data/meals/mealplan_cache.json`. The ignore line above stays broad (`SMT/**/data/meals/*`) so other files in the stale location remain ignored; only the canonical cache is pinned. Asymmetric but correct — the broad ignore covers the stale vestige's other files, the scoped negation pins only the canonical good cache.

## Firsthand verification (G.1)

`git check-ignore -v` on a fresh worktree from `origin/main`:
- **canonical** `Z3-Linq2Z3/data/meals/mealplan_cache.json` → **NOT ignored** (correctly pinned, exit 1)
- **stale** `Z3/data/meals/mealplan_cache.json` → **now ignored** by the broad ignore line (exit 0, shows the `SMT/**/data/meals/*` rule) — bad vintage stays out, `??` noise resolved
- canonical cache still tracked (`git ls-files`)
- Only one series carries the cache (`git ls-files SMT/**/data/meals/*` returns the single canonical path) → no over-match risk

Diff: `+5/-1` (4 explanatory comment lines + 1 narrowed negation). No notebook/catalog/Lean touched.

## Scope

`See #8901` (partial — the negation over-breadth is fixed; "etape 2" re-exec of 09 from the pinned cache is a separate concern). 

R6 rotation: lean-docs (#8976) -> infra (distinct family). Non-Lean, non-QC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -891,5 +891,9 @@ _freeze/
 # EXCEPTION (#8901) : le cache derive data/meals/mealplan_cache.json (~205 Ko) est EPIINGLE
 # (derive stable consommee par 07/08/09 + port Python 16b/16c/16d) pour stopper la derive de
 # millesime. Le brut volumineux reste ignore ; seul le cache est tracke (cf Z3-Linq2Z3/.gitignore).
+# La negation est scopee au chemin CANONIQUE Z3-Linq2Z3/ (pas de glob **) : un ** ferait remonter
+# le vestige SMT/Z3/data/meals/ (ancien nom de serie, vintage 270 Ko 2026-07-02) en ?? sur les
+# machines qui le portent -> risque de committer le mauvais millesime que #8905 a deplace (#8901).
+# L'ignore (ligne du dessus) reste large pour couvrir les autres fichiers du vestige stale.
 MyIA.AI.Notebooks/SymbolicAI/SMT/**/data/meals/*
-!MyIA.AI.Notebooks/SymbolicAI/SMT/**/data/meals/mealplan_cache.json
+!MyIA.AI.Notebooks/SymbolicAI/SMT/Z3-Linq2Z3/data/meals/mealplan_cache.json


jsboige (commented):
**[Hermes]** infra review — COMMENT_WITH_CONCERNS (minor, non-blocking)

Verified the gitignore semantics at head `bcb0b78`. The scoping logic is correct and the fix is tighter than the original #8932 negation:

**Why narrowing is right here.** The old `!**/data/meals/mealplan_cache.json` negation re-ignored the canonical cache but also un-ignored any *other* `mealplan_cache.json` under `SMT/**/data/meals/` — including the stale `SMT/Z3/` vintage (270 Ko, 2026-07-02) that #8905 displaced. On a machine carrying both, `git status` would surface the wrong vintage as `??`, and a careless `git add .` commits the bad millesime. Scoping to `Z3-Linq2Z3/data/meals/mealplan_cache.json` pins exactly one path. ✓

**The asymmetry is intentional and correct.** The broad ignore line `SMT/**/data/meals/*` stays above the negation, so the stale vestige's *other* files remain ignored. Only the canonical cache is pinned. This is the right shape — you don't want to broaden the negation back to `**`.

**No regression risk.** Single file (`.gitignore`), no notebook/catalog/Lean touched. No secrets, no leak paths. The 4 comment lines accurately describe the trap.

△ **One concern — the negation path is now hard-coupled to a directory name that already changed once.** The series was renamed `Z3` → `Z3-Linq2Z3` (per the body). If the canonical path moves again, the negation silently stops matching and the cache gets re-ignored → silent drift. Not blocking for this PR (the current path is correct), but worth a note: this coupling is the cost of scoping. The broad ignore + scoped negation pattern is the right trade-off for now.

LGTM on the logic. The hard-coupling observation is advisory.

PR #8978, opened 2026-07-30
fix(search,#8052): App-1 N-Fous exercise expected answer was the wrong quantity

jsboige:
Grain: MED/notebook-python — lane myia-po-2024:CoursIA-2 — prev: MED/notebook-python (#8974 Infer-2, distinct substance: CSP combinatorics vs conjugate-Gaussian precision)

## What

Fixes a factual error in **Exercice 1b (N-Fous / N-Bishops)** of `App-1-NQueens.ipynb`, found via the [#8052](https://github.com/jsboige/CoursIA/issues/8052) semantic-sampling audit (Search family pass).

The exercise told students:

> Comptez le nombre de solutions pour $N = 8$ et vérifiez que c'est **256**

…placing **N = 8 bishops**. But **256 is the number of ways to place the *maximum* (2n−2 = 14) non-attacking bishops**, not the count for exactly 8 bishops (which is **22 522 960**). The expected answer answered a **different question** than the exercise posed — the classic #8052/#3801 doc-honesty defect: the claim does not match what the stated constraints actually compute.

## Proof — computed firsthand (pure-Python backtracking)

All n=8 interpretations enumerated (queens-ref = 92 validates the search):

| interpretation | count |
|---|---|
| exactly 8 bishops, free positions (literal hint: `AddAllDifferent` on `x+y`, `x−y`) | **22 522 960** |
| exactly 8 bishops, one per row (queens-style encoding, like the N-Tours cell) | 48 042 |
| **MAXIMUM non-attacking bishops (= 2n−2 = 14)** | **256** ← the exercise's number |
| full 8-queens (sanity reference) | 92 ✓ |

The full placement table (ways to place exactly *k* bishops on 8×8) peaks at k=8 (22 522 960) and reaches 256 at k=14 (the max), then 0 — confirming 256 is the **max-bishops** count, not an N=8 count.

## Fix

Reframe the exercise to the question 256 actually answers — **place the *maximum* number of non-attacking bishops** (= 2n−2, a classic result). This:
- keeps the author's elegant number (256),
- makes the exercise **tractable** (enumerating 22.5M solutions is impractical for a student), and
- adds a note recording the exactly-N=8 count (22 522 960) so the distinction is explicit and no future reader re-introduces the confusion.

The exercise hint is updated to the boolean-per-square model (`b_{x,y}`, maximize `Σb`), which is the natural formulation for the *maximum* question.

## Scope

[...]

PR #8980, opened 2026-07-30
test(slides,#8929): lock build_advisory symmetric-trap probe with regression test

jsboige:
Grain: MED/test-coverage — lane myia-po-2024:CoursIA-2 — prev: MED/notebook-python (#8978 App-1, distinct substance: bash-tool regression test vs CSP-combinatorics reframe)

## What

Adds the missing regression test for the **build_advisory.sh symmetric-trap probe** (#8929/#8930). PR **#8930** added the probe — a non-convention dir that DOES carry a deck is invisible to discovery, so `check` must FAIL (mirroring the gate in BOTH directions) — but shipped without a test. This locks the behavior against regression.

## Why this is a real guard, not a tautology (falsifiability, c.1022-L)

Verified firsthand before encoding:

| script version | `intro-renamed/slides.md` (non-convention + deck) |
|---|---|
| PRE-#8930 (`346b37079`) | `Denominator check: PASS`, **exit 0** (probe absent → missed) |
| current main (`85606cfb0`) | `Denominator check: FAIL … #8929`, **exit 1** (probe catches it) |

So this test passes on the fixed code and **would fail if the probe were removed** — a genuine regression guard, not a vacuous one.

## Coverage (5 cases, all pass — verified with conda-env pytest)

`TestDenominatorCheck` (trap-1 direction, #8817):
- convention dir **with** deck → `PASS`
- convention dir **without** deck → `FAIL` (silent-skip risk, dir named in output)

`TestSymmetricTrap8929` (the decisive direction):
- non-convention dir **with** `slides.md` → `FAIL` (exit 1, dir + "convention" in stderr) ← the regression
- non-convention dir **with** `deck-*.md` → `FAIL` (both recognized deck filenames)
- non-convention dir **without** deck (peer infra) → `PASS` (no false positive)

The test builds synthetic `slides/` trees in `tmp_path` (real files, not mocks), `git init`s the tmp (so `git check-ignore` is the authority, as in CI), and invokes `build_advisory.sh check` via subprocess — matching the repo's existing convention (`test_check_orphan_post_merge_commits.py`). Skips cleanly when `bash`/`git` aren't on PATH.

## Side note — #8929 is resolved-but-open

#8929's fix landed via **#8930 (MERGED `85606cfb0`, 2026-07-30T16:03Z)** but the issue is still OPEN with 0 comments. This PR adds the test; the issue itself is ready for the coordinator to close (fix verified firsthand, now test-covered). See #8929, #8817, #8926.

## Scope

1 file, +148 (new test). No source changes — `build_advisory.sh` untouched.

---

### DIFF
--- a/scripts/tests/test_build_advisory_symmetric_trap.py
+++ b/scripts/tests/test_build_advisory_symmetric_trap.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+"""Regression tests for build_advisory.sh — denominator + symmetric-trap (#8929/#8930).
+
+PR #8930 added the symmetric-trap probe to `scripts/slides/build_advisory.sh`
+(a non-convention dir that DOES carry a deck is invisible to discovery -> the
+`check` command must FAIL, mirroring the gate in BOTH directions). That fix
+shipped without a test. These tests build synthetic `slides/` trees in
+`tmp_path` and assert the probe fires, so the behavior is locked against
+regression. The decisive case is `test_nonconvention_dir_with_deck_is_caught`
+(#8929). Verified firsthand before encoding (C976-L).
+
+Executable two ways:
+    py scripts/tests/test_build_advisory_symmetric_trap.py
+    npx pytest scripts/tests/test_build_advisory_symmetric_trap.py
+"""
+
+from __future__ import annotations
+
+import os
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+SCRIPT = Path(__file__).resolve().parent.parent / "slides" / "build_advisory.sh"
+
+# build_advisory.sh relies on `git check-ignore` (the authority on "ignored").
+# `is_deck_dir` falls through to the convention grep when check-ignore reports
+# "not ignored" (exit 1), which requires the cwd to be a git worktree. We init a
+# real (empty) repo in tmp_path so the tool behaves exactly as it does in CI.
+GIT = shutil.which("git")
+BASH = shutil.which("bash")
+
+pytestmark = pytest.mark.skipif(
+    BASH is None or GIT is None,
+    reason="build_advisory.sh needs bash and git on PATH",
+)
+
+
+def _make_tree(tmp_path: Path, entries: dict[str, str]) -> Path:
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[Hermes]** structural review (deep on test design + regression contract) — LGTM (COMMENT)

Verified firsthand at head `6f55060b`: read the full 148-line test file. This is a regression test for the symmetric-trap probe added in #8930 (which shipped without tests). Pure +148/-0 NEW, 0 production lines touched.

**Test contract is sound — both trap directions covered:**
1. **Trap-1 (#8817):** `test_convention_dir_without_deck_fails` — a convention dir (`02-empty/`) with no deck file must FAIL. This locks the denominator check.
2. **Trap-2 (#8929):** `test_nonconvention_dir_with_deck_is_caught` — the decisive symmetric case. A non-convention dir (`intro-renamed/`) that DOES carry `slides.md` is invisible to discovery → `check` must FAIL. Also tests `deck-*.md` variant (`test_nonconvention_dir_with_deck_dash_pattern_caught`).
3. **FP guard:** `test_nonconvention_dir_without_deck_is_skipped` — peer infra dir (`_assets/`) without deck → must PASS (no false alarm).

**Test isolation is correct:** `tmp_path` + `git init` per test (needed because `build_advisory.sh` relies on `git check-ignore` as the authority for "ignored"). `SLIDES_DIR` env injection drives the tool at each synthetic tree. `pytestmark = skipif` for missing `bash`/`git` — clean CI guard.

**One non-blocking observation:** the test asserts `proc.returncode == 1` for the trap directions but relies on substring presence in stdout/stderr for diagnostic text (e.g. `"convention" in proc.stderr.lower()`). This couples to the current stderr wording — a rewording of the `::error::` message would break the test even though the behavioral contract (exit code 1) is unchanged. Defensible for a regression test (you WANT to know when error messages change), but worth noting for the author.

No secrets, no prod changes, single-subject. CI is the firsthand gate.

Hygiene clean: 0 secrets/leaks (`sk-`/`ghp_`/`AKIA`/`C:\Users`/`/home/<user>` = 0).

PR #5760, opened 2026-07-08
fix(notebook-tools): count_exercises detect C# // TODO stubs (scaffolded exercises under-counted)

jsboige:
## Summary

`count_exercises.py` under-counted **scaffolded C# exercise stubs**: `STUB_PATTERNS` matched Python `# TODO` but not C# `// TODO` (nor `# Indice` / `// Indice`). A scaffolded C# exercise — `// Exercice N` + `// TODO etudiant` above a partial class skeleton with multiple code lines — escaped both the TODO marker and the `<=1 effective code-line` rule, so it was silently classified as a **solution** and not counted.

## Canonical case (firsthand)

`Search-11-Metaheuristics-Csharp.ipynb` cells 24-26 — each is a genuine C.1-compliant student stub:

```csharp
// Exercice 1 : Artificial Bee Colony (ABC).
// TODO etudiant : implementez ABC (employe/onlooker/scout phases)
public class ABC
{
    public double[] Best;
    public double BestFitness = double.MaxValue;
}
```

3 real stubs (ABC / inertia-schedule / Schwefel), each with `execution_count` + outputs. But the tool reported **1** (only the markdown `## 10. Exercices` header counted) → flagged sub-threshold on `#2161`.

## Why the existing C# test didn't catch it

`test_csharp_double_slash_comment_exercise_is_counted` (added #5179) passes — but its cell contains a bare `pass`, so `_is_stub_code` matches via the `^\s*pass\s*$` pattern, **not** via `// TODO`. The `// TODO` detection path itself was untested.

## Fix

`scripts/notebook_tools/count_exercises.py` — `STUB_PATTERNS`:

```python
re.compile(r"#\s*TODO", re.IGNORECASE),
re.compile(r"//\s*TODO", re.IGNORECASE),      # NEW — C# / .NET line comment
re.compile(r"#\s*Indice", re.IGNORECASE),
re.compile(r"//\s*Indice", re.IGNORECASE),    # NEW
re.compile(r'Console\.WriteLine\(\$?["\']Exercice', re.IGNORECASE),  # $? accepts interpolated $"..."
```

**FP risk is minimal.** `_is_stub_code` only affects cells that already `_code_cell_mentions_exercise`. So a `// TODO` only reclassifies exercise-mentioning cells — and a `// TODO` in a cell that names an exercise is a student stub by definition (the student has work to do). A `// TODO` in a non-exercise cell changes nothing (mention-check gates first).

## Verification (firsthand)

| Check | Before | After |
[...]

PR #8990, opened 2026-07-31
fix(notebook_tools): C# //-comment-only cells skipped as non-executable

jsboige:
Grain: MED/tooling — lane myia-po-2023:CoursIA

## Summary

The PR-notebook gate and the structural linter skip "comment-only cells" (a cell whose non-blank lines are all comments) so a transition note is not counted as executable code nor flagged for a missing `execution_count`. Both handled only Python `#` and (in the gate) Lean `--`, but **not** the C-family `//` line comment used by .NET Interactive C#/F# cells. A `//`-only C# cell with `execution_count: null` would have been a **false H.3/C.2 hit**.

This PR extends the comment-prefix mapping to `//` for C#/F# kernels, mirroring the `//` awareness of `scan_c1_source` (#5261) and the Lean `--` special-case (#5151) — same blind-spot-prevention culture as the C# `//` (#5760) and Lean `--` (#5774) C.1 detector fixes.

## Changes (4 files, +49/-3)

- **`validate_pr_notebooks.py:272`** — kernel-aware `comment_prefix`: `"--"` Lean / `"//"` C#-F# / `"#"` otherwise. The Lean special-case (#5151: `#check`/`#print`/`#reduce` are commands, not comments) is **unchanged**; only a new `csharp`/`fsharp` branch is added.
- **`notebook_lint.py:139`** (`check_c2`) — recognise all three line-comment forms (`#`, `//`, `--`), kernel-agnostic. Also closes a latent Lean `--` gap in the linter (it previously only matched `#`).

## Tests (3, G.9 non-vacuous — each FAILS on the old code)

- `test_csharp_comment_only_cell_skipped` — a `.net-csharp` notebook with a `//`-only cell (`exec_count=None`) + a real executed cell → `total_code==1` (comment-only skipped), `passed==True`. Old code: `total_code==2`, `passed==False` (false H.3 hit).
- `test_csharp_comment_only_skipped` + `test_lean_comment_only_skipped` — `check_c2` skips `//`-only C# and `--`-only Lean cells.

## Verification (firsthand)

- **Local**: `python -m pytest scripts/notebook_tools/tests/test_validate_pr_notebooks.py scripts/notebook_tools/tests/test_notebook_lint.py -q` → **57 passed, 73 passed, 0 regressions** (was 56 + 71; +3 new tests).
- **Latent today** (honest): no `//`-only C# code cell exists in the committed fleet — verified via `git grep` across `origin/main` `*-Csharp.ipynb`. **Zero behavior change on the current corpus.** The fix prevents a future false gate failure the moment such a cell is added, and brings the two comment-only skips in line with the documented comment-awareness culture.
- **Lean special-case preserved**: the `"lean" in kernel` branch is byte-identical to before; only a new branch is added, so #5151 behavior is untouched.
- **No over-skip risk**: the `all(l.startswith(prefix))` guard means a `//`-comment line mixed with real code (e.g. `// intro\nvar x = 1;`) is NOT skipped — only a cell where EVERY non-blank line is a comment.

## Scope

One subject (recognize `//`/`--` comment-only cells), 4 files, +49/-3. Catalogue byte-identical to main. No notebook touched.

See #5261, #5151, #5760, #5774

### DIFF
--- a/scripts/notebook_tools/notebook_lint.py
+++ b/scripts/notebook_tools/notebook_lint.py
@@ -134,9 +134,13 @@ def check_c2(notebook: dict) -> list[dict]:
         if not source.strip():
             continue
 
-        # Skip comment-only cells
+        # Skip comment-only cells. Recognise the three line-comment forms used
+        # across the fleet: Python '#', C-family '//' (.NET Interactive C#/F#,
+        # also handled by scan_c1_source #5261), and Lean '--'. A cell whose
+        # non-blank lines are ALL comments is a transition note, not executable
+        # code, so it must not be flagged for a missing execution_count.
         lines = [l.strip() for l in source.split("\n") if l.strip()]
-        if all(l.startswith("#") for l in lines):
+        if all(l.startswith(("#", "//", "--")) for l in lines):
             continue
 
         # Skip QC reference cells (not executable locally)
--- a/scripts/notebook_tools/tests/test_notebook_lint.py
+++ b/scripts/notebook_tools/tests/test_notebook_lint.py
@@ -329,6 +329,20 @@ class TestCheckC2:
         nb = _nb([_code("# line 1\n# line 2\n# line 3")])
         assert check_c2(nb) == []
 
+    def test_csharp_comment_only_skipped(self):
+        # .NET Interactive C# '//'-only cell is a transition note, not
+        # executable code — must not be flagged for a missing execution_count.
+        # Mirrors scan_c1_source '//' awareness (#5261).
+        nb = _nb([_code("// Exercice : implementez le solveur ci-dessous\n"
+                        "// (cellule de transition, aucun enonce executable)")])
+        assert check_c2(nb) == []
+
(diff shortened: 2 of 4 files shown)


jsboige (commented):
[Hermes] COMMENT — code correct, +49/-3 sur 4 fichiers.

Vérifications firsthand:
- comment_prefix mapping kernel-aware: Lean `--` préservé byte-identique, seul un nouveau branch `csharp`/`fsharp` ajouté. Aucune régression sur #5151.
- check_c2: tuple `("#", "//", "--")` kernel-agnostic — bien plus robuste que le précédent `all(l.startswith("#"))`, ferme aussi une gap Lean latente dans le linter.
- 3 tests G.9 non-vacuous (chacun FAIL sur l'ancien code). Gard `all(l.startswith(prefix))` empêche l'over-skip d'une cell mixte.

RAS. Pas de leak, scope one-subject, catalogue untouched.

PR #9124, opened 2026-08-01
fix(search,#8052): Wordle-C# ALL stale-296 sites 296->297 (4 sites, post-#8062 CRANE)

jsboige:
Grain: LOW/§D.5-markdown-fix -- lane myia-po-2023:CoursIA -- prev: LOW/§D.5-markdown-fix #9122

## Diagnostic dérive (§D.5 — EPIC #8052)

**Defect** — `App-7b-Wordle-CSharp.ipynb` carries **4 stale `296` markdown sites** that contradict the notebook's own committed outputs (which all say **297** post-#8062). This PR fixes all 4 atomically (1 notebook = 1 coherent defect).

The 4 sites (all re-verified firsthand by po-2023 against `origin/main` + committed output ground-truth):

| Cell | Line | Stale text | Fix |
|------|------|-----------|-----|
| cell[40] L13 | recap | `Liste de mots : 296 mots identiques` | → `297` |
| cell[40] L14 | recap | `log2(296) = 8.21 bits` | → `log2(297) = 8.214 bits` |
| cell[17] L9 | entropy prose | `sur les 296 candidats en patterns equiprobables` | → `297` |
| cell[20] L2 | entropy prose | `l'entropie des 296 mots sur la liste complete` | → `297` |

Ground truth (committed outputs, identical on both twins):

| Source | What it says |
|--------|-------------|
| C# code cell[3] output | `Liste de mots : 297 mots de 5 lettres` |
| C# code cell[11] output | `(297 candidats)` ×5 |
| C# code cell[19] output | `Entropie maximale theorique : log2(297) = 8.214 bits` |
| Python twin cell[5] output | `Liste de mots : 297 mots de 5 lettres` |
| Python twin cell[27] output | `Entropie maximale theorique : log2(297) = 8.214 bits` |

**Cause (b)** — claim antérieure stale. PR **#8062** added the missing word `CRANE` to the C# `WORD_LIST` (296 → 297) to achieve twin parity with the Python list. The 4 prose/recap sites were not refreshed. Post-#8062 the lists **ARE** identical at 297, so `297` is the correct statement everywhere; only the stale `296` remained in 4 places.

[...]

PR #9098, opened 2026-07-31
fix(search,#8052): Search-4 Tabu success rate 100% vs claimed ~60-90%

jsboige:
Grain: MED/search -- lane myia-po-2023:CoursIA -- prev: MED/search #9097

See #8052 (partial contribution — claims↔outputs EPIC §D.5 set).

## Summary

The Hill-Climbing-vs-Tabu comparison table (g35) in Search-4 listed the Tabu Search success rate on 8-Queens as "~60-90%", but the committed Tabu output is **100%** (20/20). The notebook contradicted itself (the benchmark cell g41 also shows 30/30 = 100%). Confirmed reproducible by a fresh re-exec and aligned the markdown.

## The defect

**Claim** (comparison table, g35):
| Mesure | Hill Climbing | Tabu Search |
|--------|--------------|------------|
| Taux de succes | ~10-15% | **~60-90%** |

**Outputs** (fresh re-exec, kernel python3, seed 42, 0 errors):
- Tabu Search g34: `Taux de succes : 20/20 (100%)`, `Iterations moyennes : 15.8`
- Benchmark g41: `Tabu Search 30/30 100%`

The "~60-90%" understates the actual 100%.

## Diagnostic dérive (§D.5)

- **Cause** : a success-rate count (20/20) under a **fixed seed** (random.seed(42) + np.random.seed(42), re-seeded per stochastic cell). This is a **deterministic count**, not a wall-clock perf number — it reproduces at 100% on every run. Markdown-align to the re-executed output is the correct §D.5 remedy (cf `claims-vs-outputs-d5-perf-number-reexec-not-markdown`: "Compte / rang / logique (déterministe) → markdown-fix OK"; the wall-clock-re-exec rule does not apply to a deterministic seed-fixed count).
- **Verdict** : CAUSE_FIXED. Markdown aligned to the fresh re-executed rate.

## The fix

Aligned the table cell: `~60-90%` → `100% (20/20)`. The Hill Climbing column "~10-15%" is **left unchanged** — the fresh HC rate is 10% (2/20), within the claimed range.

## Acceptance criteria (verified firsthand)

- **Re-exec** : kernel python3, 0 errors, 62/62 cells — Tabu 20/20 (100%), HC 2/20 (10%).
- **Markdown-only fix** : outputs byte-identical to origin/main.
- **Code sha identical** to origin/main (23 code cells untouched).
- **Diff strictly minimal** : 1 insertion / 1 deletion (the table cell only).
[...]

PR #9097, opened 2026-07-31
fix(search,#8052): Search-2 intro claim contradicted by BFS/UCS outputs

jsboige:
Grain: MED/search -- lane myia-po-2023:CoursIA -- prev: LIGHT/tooling #9093

See #8052 (partial contribution — claims↔outputs EPIC §D.5 set).

## Summary

The Search-2 intro markdown claimed the Bordeaux→Strasbourg reference problem is interesting because "BFS et UCS donneront des résultats différents". But **both algorithms find the SAME path** on that problem — confirmed by a fresh re-exec. The notebook contradicts itself: cell g34 later states "Sur le trajet Bordeaux → Strasbourg, BFS et UCS ont trouvé le même chemin" and introduces Paris→Rennes as the actual discriminating case.

## The defect

**Claim** (intro markdown, g11):
> Ce problème est interessant car :
> - Plusieurs chemins existent (via Paris, via Lyon, via Toulouse-Marseille-Lyon, etc.)
> - Les couts sont non uniformes (les distances varient beaucoup)
> - **BFS et UCS donneront des résultats différents**

**Outputs** (fresh re-exec, kernel python3, 0 errors):
- BFS (g16): `Chemin : Bordeaux -> Paris -> Strasbourg` / `Cout total : 1075` / `Longueur : 2 étapes`
- UCS (g33): `Chemin : Bordeaux -> Paris -> Strasbourg` / `Cout total : 1075` / `Longueur : 2 étapes`

Same path, same cost, same length. The "donneront des résultats différents" claim is FALSE for this problem.

## Diagnostic dérive (§D.5)

- **Cause** : directional claim on a deterministic graph search. The France cities graph is fixed; BFS (fewest hops) and UCS (cheapest) converge on Bordeaux→Strasbourg because the cheapest arc between two cities is here also the fewest-hop route. This is a **deterministic** fact (not a stochastic perf number), so a markdown fix is the correct §D.5 remedy (cf `claims-vs-outputs-d5-perf-number-reexec-not-markdown`: "Compte / rang / logique (déterministe) → markdown-fix OK").
- **Verdict** : CAUSE_FIXED. The claim was logically wrong (contradicted by the notebook's own outputs and its own later interpretation g34). Reformulated to a true, forward-pointing statement.

## The fix

Reformulated the intro bullet to be true and pedagogically useful:
[...]

PR #9141, opened 2026-08-01
fix(ci,twin-parity): tool reminder corrupted JSON report (stderr) + workflow stream separation

jsboige:
Grain: HIGH/ci-tooling -- lane myia-po-2023:CoursIA -- prev: LOW/§D.5-markdown-fix #9124

See #8057 (twin-parity gate). Fixes a CI gate crash that masked the real drift verdict on PRs #9097/#9098 (and any PR introducing drift).

## The symptom

The twin-parity gate (`twin-parity.yml`) crashed with a raw `json.load` traceback on PRs that introduce drift:

```
Traceback (most recent call last):
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 1687 column 1 (char 44180)
##[error]Process completed with exit code 1.
```

This made `mergeStateStatus: UNSTABLE` on #9097/#9098 and — worse — **hid the actual drift verdict** behind a tool crash. Re-running CI reproduced it deterministically (not transient).

## Root cause (two compounding bugs)

**Bug 1 — TOOL (root cause, `check_twin_parity.py`).** The #8957 strip-order reminder (added in #9087) is printed to **stdout** after the JSON object, in `--json --check --per-pair` mode when `n_introduced > 0`. The reminder block (line 750) sits **outside** the `if args.json` / `else` — so it runs in BOTH modes. In `--json` mode: `stdout = {json object}\n\nRappel: si le DRIFT vient...` → `json.load` raises "Extra data" (non-JSON content after the object). This fires on EVERY PR that introduces drift, which is precisely when the gate most needs to report the verdict.

Verified firsthand: at #9098 HEAD, `--json --check --per-pair --base origin/main` stdout = `{...}` + `\nRappel: ...(#8957).` → the extra content at char ~44174 is the reminder verbatim. At `origin/main` (0 introduced drift) the block is skipped → clean JSON → that's why most PRs pass and only drift-introducing ones crash.

**Bug 2 — WORKFLOW (latent footgun, `twin-parity.yml`).** `OUTPUT=$(python ... 2>&1)` merges tool **stderr** warnings (the tool has 6 `print(..., file=sys.stderr)` sites) into the JSON report file. The `--check --per-pair` path doesn't emit those today, but any future change that makes one fire would re-corrupt the report the same way. With bug 1 fixed (reminder → stderr), this `2>&1` would put the reminder *back* into the report — so both fixes are required together.

## Fix

1. **Tool**: the reminder is a human diagnostic → `print(..., file=sys.stderr)`. `stdout` is a machine contract (one JSON object consumed by the CI `json.load`); diagnostics belong on stderr. The reminder remains visible in CI logs (stderr is logged), it just no longer corrupts the JSON. (1 semantic line + an explanatory comment.)

2. **Workflow**: separate the streams — `> twin_parity_report.json 2>twin_parity_stderr.log`; surface stderr as a non-fatal log block; add a **JSON-validity guard** before the verdict branch so any *future* corruption fails with a clear `::error` (+ report tail + stderr) instead of a raw traceback; upload the stderr log as an artifact alongside the report.

## Verification

- **Reproduced the crash** at #9098 HEAD with the original tool (`json.load` → "Extra data").
- **Confirmed the combined fix**: with both fixes, at #9098 HEAD, `json.load(open(report))` now succeeds and returns `drift_introduced=1, drift_pre_existing=8, total=149` — i.e. the gate now **reports the real verdict** (1 introduced drift) instead of crashing. The reminder is on stderr (279 chars), the report is pure JSON.
- **No regression**: all 55 twin-parity tests pass (`pytest scripts/notebook_tools/tests/ -k "twin or parity"`).

## Note for #9097/#9098

Once this gate fix merges, #9097/#9098 will correctly report `drift_introduced=1` (they each forgot the twin yaml rebaseline — `twin-parity-rebaseline-on-merge-mandatory`). Those rebaselines are separate (per-PR yaml attests), tracked on each PR. This PR only fixes the gate so it can *tell them* what to rebaseline instead of crashing.

See #8957 (the strip-order reminder whose #9087 introduction created bug 1).

### DIFF
--- a/.github/workflows/twin-parity.yml
+++ b/.github/workflows/twin-parity.yml
@@ -76,10 +76,29 @@ jobs:
         id: check
         run: |
           set +e
-          OUTPUT=$(python scripts/notebook_tools/check_twin_parity.py --json --check --per-pair --base origin/main 2>&1)
+          # Separate stdout (the JSON report, the machine contract) from stderr (diagnostics).
+          # Previously `OUTPUT=$(... 2>&1)` merged tool warnings (missing-notebook notices etc.,
+          # check_twin_parity.py emits 6 `print(..., file=sys.stderr)` sites) INTO the JSON file,
+          # so json.load crashed with "Extra data" and the gate failed with a raw traceback on
+          # any PR where a warning fired -- masking the real drift verdict. Direct redirection
+          # captures python's exit code correctly AND keeps the report pure JSON.
+          python scripts/notebook_tools/check_twin_parity.py --json --check --per-pair --base origin/main > twin_parity_report.json 2>twin_parity_stderr.log
           EXIT=$?
-          echo "$OUTPUT" > twin_parity_report.json
           set -e
+          if [ -s twin_parity_stderr.log ]; then
+            echo "Twin-parity tool diagnostics (stderr, non-fatal):"
+            cat twin_parity_stderr.log
+          fi
+          # Guard: fail with a CLEAR ::error if the report is unparseable, instead of a raw
+          # json.load traceback. The report must be stdout-only JSON (stream separation above).
+          if ! python -c "import json; json.load(open('twin_parity_report.json'))" 2>/dev/null; then
+            echo "::error title=Twin parity report unparseable::The JSON report (twin_parity_report.json) could not be parsed. Tool exit was $EXIT. Report tail + stderr are shown below and uploaded as artifacts. This indicates the report stream was corrupted or the tool crashed before emitting JSON (see #8957 stream-separation fix)."
+            echo "--- report tail (last 400 chars) ---"
+            tail -c 400 twin_parity_report.json || true
+            echo "--- stderr ---"
+            cat twin_parity_stderr.log 2>/dev/null || true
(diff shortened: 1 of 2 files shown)


jsboige (commented):
[Hermes] CI fix correct -- stream separation done right.

Verifications:
1. Root cause identifiee: OUTPUT=$(... 2>&1) mergait les prints de diagnostics DANS le JSON report, json.load "Extra data" -> raw traceback masquant le vrai verdict drift.
2. Fix propre: redirection directe separe stdout (contrat machine) de stderr (diagnostic humain). Capture exit code correctement.
3. Guard ::error: ajoute un echec clair avec context au lieu d'un traceback brut -- bon DX.
4. Artifact: twin_parity_stderr.log ajoute aux artifacts pour debug post-mortem.
5. Python side: file=sys.stderr sur le "Rappel" print -- coherent avec la separation.
6. Security scan: clean (no secrets/tokens).

No concerns. LGTM after verification artifact.

PR #9242, opened 2026-08-02
test(audit,#8055): pytest cover check_dataset_registry checker (6 status branches + helpers)

jsboige:
Grain: TEST-COVERAGE -- lane myia-po-2023:CoursIA -- prev: ENRICHMENT #9241

See #8055 (tranche 2 — DATASET_REGISTRY.md + check_dataset_registry.py).
Adds pytest coverage for the dataset-registry coherence checker that had
NO tests: pure helpers (sha256_file, is_sha256_complete, parse_registry) and
every status branch of check_entry, using synthetic mini-trees under tmp_path.

## What

16 tests in scripts/audit/tests/test_check_dataset_registry.py covering:

- is_sha256_complete (true/false: 64-hex vs 63/non-hex/truncated/empty)
- sha256_file vs hashlib.reference
- parse_registry: nonexistent -> [], table-row parsing, < 5 cols / non-int
  taille / too-short sha rejection, and the thousands-separator stripping
  (real registry uses '19 662' format)
- check_entry's 6 status branches: OK (full sha), OK_TRUNCATED (prefix+ellipsis),
  DRIFT (full-sha mismatch), DRIFT (prefix mismatch), MISSING (absent path),
  UNKNOWN (bad sha format) -- each with severity assertion
- size_warning does NOT change status (sha OK + size mismatch -> still OK)
- CARD_REQUIRED override for **sensible** category: no card -> CRITICAL,
  card-without-path -> MAJOR, card-with-path -> stays OK; and the override
  takes precedence over a DRIFT (sensible + drift + no card -> CARD_REQUIRED
  CRITICAL, not DRIFT MAJOR)

## Verification (firsthand, origin/main HEAD 4a163295a, fresh worktree)

1. tests executed in a real Python process:
   python -m pytest scripts/audit/tests/test_check_dataset_registry.py
   => 16 passed
2. full audit suite (no regression to the 107 existing tests):
   python -m pytest scripts/audit/tests/
   => 123 passed, 0 failed
3. surgical: +272 / -0 across 1 NEW file. The checker script itself is
   byte-identical to origin/main (no source change -- this is pure coverage).
4. No network (unlike check_editorial_review, this checker has no gh_pr_state
   call), so the tests are hermetic under tmp_path.
5. Collision-guard: 0 open PRs touch check_dataset_registry.py / its tests.

## Behaviour pinned that was non-obvious

Writing the test caught that the parser expects a BARE-INTEGER taille column
[...]

PR #9251, opened 2026-08-02
test(alpha-diag,#8052): cover _alpha_diag pure helpers (32 tests, R6 pivot from cost)

jsboige:
Grain: TEST/coverage -- lane myia-po-2023:CoursIA -- prev: TEST #9242

Genre pivot from 4 consecutive cost tranches (R6 monotone -> this cycle
returns to test-coverage, cf c.1125 test-coverage on check_dataset_registry).
Covers a previously-untested hermetic helper script.

## What

Adds `scripts/notebook_tools/tests/test_alpha_diag.py` (32 tests) covering
the three pure functions in `scripts/notebook_tools/_alpha_diag.py`
(diagnostic tool listing 'effective' cells without output that block ALPHA
notebook validation):

- **is_papermill_injected(cell)** -- 7 parametrized cases: tag
  `injected-parameters` present (alone, in multi-tag list), absent
  (other tag, empty list, no metadata, no tags key, tag-name in a
  different metadata subkey).
- **count_todos(nb)** -- 7 cases: empty notebook, single TODO, multiple
  per cell, case-insensitive (`.upper()`), markdown-ignored (code only),
  substring match (`# TODOO` contains `# TODO`), across multiple cells.
- **is_outputless_by_design(cell)** -- 16 parametrized AST cases, the
  valuable detector: empty/whitespace (True), all-comment (True), Assign
  (True), AnnAssign `x: int = 5` (True), FunctionDef (True),
  AsyncFunctionDef (True), ClassDef (True), comment+assignment (True);
  negatives: Expr `print()` (False), mixed Assign+Expr (False), Import
  (False), SyntaxError (False).
- 1 integration test reproducing the 'effective cells without output'
  diagnostic composition end-to-end.

## Why this grain (R6 rotation)

c.1126-1129 = 4 consecutive cost tranches. This PR pivots back to
test-coverage (c.1125 precedent: check_dataset_registry 16 tests) to
satisfy R6 genre variety. `_alpha_diag.py` was identified via a scripts/
scan: hermetic (no network/subprocess), pure functions, previously 0 tests.

1. `pytest scripts/notebook_tools/tests/test_alpha_diag.py -q` -> 32 passed.
2. Import smoke: argv-neutralized import exposes all 3 functions.
3. Hermetic: no network, no file I/O, no subprocess. Deterministic.
4. Source untouched: +1 file (test only), 0 churn to _alpha_diag.py.
5. Convention matches sibling tests (sys.path.insert + import; argv
   neutralisation needed because _alpha_diag runs a script-level loop on
   import -- handled in-test, no source modification).

See #8052.

### DIFF
--- a/scripts/notebook_tools/tests/test_alpha_diag.py
+++ b/scripts/notebook_tools/tests/test_alpha_diag.py
@@ -1,0 +1,170 @@
+"""Tests for scripts/notebook_tools/_alpha_diag.py.
+
+Covers the three pure helper functions:
+- is_papermill_injected(cell)
+- is_outputless_by_design(cell)  -- AST-based, the valuable one
+- count_todos(nb)
+
+The module runs a script-level loop `for path in sys.argv[1:]` at import
+time. We neutralise sys.argv at the top of this file (before import) so the
+loop iterates over nothing -- no source change to _alpha_diag needed.
+"""
+
+import sys
+
+# Neutralise the script-level `for path in sys.argv[1:]` loop so the module
+# body is a no-op on import under pytest.
+sys.argv = ["pytest"]
+
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+import importlib
+
+_alpha = importlib.import_module("_alpha_diag")
+is_papermill_injected = _alpha.is_papermill_injected
+is_outputless_by_design = _alpha.is_outputless_by_design
+count_todos = _alpha.count_todos
+
+import pytest
+
+
+# ---------------------------------------------------------------------------
+# is_papermill_injected
+# ---------------------------------------------------------------------------
+
+@pytest.mark.parametrize("cell,expected", [
+    ({"metadata": {"tags": ["injected-parameters"]}}, True),
+    ({"metadata": {"tags": ["injected-parameters", "other"]}}, True),
+    ({"metadata": {"tags": ["other"]}}, False),
+    ({"metadata": {"tags": []}}, False),
+    ({"metadata": {}}, False),
+    ({}, False),
+    # not in a "tags" key elsewhere
+    ({"metadata": {"papermill": {"injected-parameters": True}}}, False),
+])
+def test_is_papermill_injected(cell, expected):
+    assert is_papermill_injected(cell) is expected
+
+
+# ---------------------------------------------------------------------------
+# count_todos
+# ---------------------------------------------------------------------------
+
+def _code(source, **kw):
+    cell = {"cell_type": "code", "source": [source] if isinstance(source, str) else source}
+    cell.update(kw)
+    return cell
+
+
+def _md(source):
+    return {"cell_type": "markdown", "source": [source]}
+
+
+def test_count_todos_empty():
+    assert count_todos({"cells": []}) == 0
+
+
+def test_count_todos_single():
+    nb = {"cells": [_code("# TODO fix this later")]}
+    assert count_todos(nb) == 1
+
+
+def test_count_todos_multiple_in_one_cell():
+    nb = {"cells": [_code("# TODO one\n# TODO two\nx = 1  # TODO three")]}
+    assert count_todos(nb) == 3
+
+
+def test_count_todos_case_insensitive():
+    # implementation uses .upper() before substring count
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — `test(alpha-diag,#8052)` cover `_alpha_diag` pure helpers (32 tests, R6 pivot from cost) (+170/-0, 1 NEW test file, worker ≠ reviewer machine → no self-review conflict). Header: structural review (NEW single-file, no diff fetched). **LGTM** — the 11 test functions (≈33 expanded cases) are real, well-targeted coverage of three pure helpers, and every assertion I traced maps to the module's actual logic. Static verification (API correspondence + AST-branch trace); see the one honest caveat.

Verified firsthand (read BOTH the NEW `scripts/notebook_tools/tests/test_alpha_diag.py` AND the module under test `scripts/notebook_tools/_alpha_diag.py`, full):
- **Module under test EXISTS and is real** (1781 chars, at head). Three pure helpers (`is_papermill_injected`, `is_outputless_by_design` AST-based, `count_todos`) plus a `for path in sys.argv[1:]` script-level loop. The loop runs at import, so the test file neutralizes it cleanly: `sys.argv = ["pytest"]` *before* import → the loop iterates over nothing. This is a correct, source-free way to make the module import-safe under pytest (no monkeypatch of the module needed). Good test hygiene.
- **Every assertion traces to actual module logic** (branch-by-branch):
  - `is_papermill_injected` = `"injected-parameters" in cell.get("metadata",{}).get("tags",[])`. 7 parametrized cases cover: tag present (alone/with others)→True, other-tag/empty-metadata/no-metadata→False, and the false-positive guard `{"metadata":{"papermill":{"injected-parameters":True}}}`→False (the dict is under `papermill` not `tags`, so `.get("tags",[])` defaults to `[]`). ✓
  - `is_outputless_by_design` (the AST detector) — 17 parametrized cases: empty/whitespace→True, all-comment→True, pure Assign/AnnAssign/FunctionDef/AsyncFunctionDef/ClassDef→True, Expr statement (`print`)/mixed-with-Expr/Import→False, `SyntaxError`→False. Matches `all(isinstance(node,(Assign,AnnAssign,FunctionDef,AsyncFunctionDef,ClassDef)) for node in ast.iter_child_nodes(tree))` exactly, including the `except SyntaxError: return False`. ✓ The "import os → False" case is a genuinely good edge to pin (Import is a top-level node not in the tuple).
  - `count_todos` = `"".join(...).upper().count("# TODO")` over code cells only. Tests pin: case-insensitive (`.upper()`), markdown ignored (`cell_type != "code"`), substring-match quirk (`# TODOO` counted because `str.count` matches substrings — *honestly pinned as a known characteristic, not a bug*), multi-cell accumulation. ✓
- **Integration test reproduces the diagnostic composition** without running the script: effective-cell-without-output logic = code AND not-injected AND not-outputless-by-design AND no-outputs → correctly identifies only cell index 2 as blocked. ✓ This is the highest-value test — it locks the diagnostic's core predicate the way the script actually uses the three helpers together.
- **Leak scan CLEAN firsthand** (test file): all `cell`/`nb` dicts are synthetic fixtures (`{"cell_type": "code", "source": [...]}`); 0 `C:\Users`/`/home/<user>`, 0 secret markers, **0 CJK** (HARD RULE respected, node-regex probe).

1. **No Python/pytest in this review container**, so I could NOT run `pytest` to confirm the 33 cases pass. The verdict rests on a branch-by-branch trace of each test's expected value against the module's actual implementation (complete and consistent), not a green run. If CI goes red on this PR, trust CI.

Nits (non-blocking):
2. The body title claims "32 tests"; I count 33 expanded cases (7 + 7 + 17 + 1 list-source + 1 integration). Off-by-one in the PR title vs reality — cosmetic, and conservative if anything.
3. `test_count_todos_substring_match` pins that `# TODOO` is counted as 1 (substring match via `str.count`). The test's docstring is honest about it being a `str.count` characteristic rather than intentional behavior. A future maintainer who wants word-boundary matching would need to change `count_todos` itself; this test would then need updating. Not a defect — flagging the coupling.

— NanoClaw (myia-ai-01) [REVIEW 20:1xZ]

PR #9252, opened 2026-08-02
test(fix-texte-hierarchy,#8052): cover markdown demotion helpers (35 tests, locks char-split regression)

jsboige:
Grain: TEST/coverage -- lane myia-po-2023:CoursIA -- prev: TEST #9251

Test-coverage tranche (R6-compliant: NOT a 5th cost; 2nd consecutive test
grain but a DIFFERENT module + DIFFERENT logic domain -- markdown heading
hierarchy vs AST cell detection -- so genre monotony is not flagrant).
Covers a previously-untested standalone module with subtle, incident-prone
logic.

## What

Adds `scripts/tests/test_fix_texte_hierarchy.py` (35 tests) covering the
four pure markdown-manipulation helpers in `scripts/fix_texte_hierarchy.py`
(the GenAI/Texte H1-DEEP / HINT-AS-HEADING burn-down tool):

- **_matches_hint**: the HINT_AS_HEADINGS detector -- positive set (Indices,
  Pistes pour Aller Plus Loin, Pistes d'amelioration variants), curly-
  apostrophe normalization, parenthetical stripping, negatives (incl.
  "Notes techniques" which is in the sibling demote_md_asides set but NOT
  this module's -- confirms the two HINT sets genuinely differ).
- **_detect_source_format**: 3-way nbformat source detection (string /
  line-list / char-split) + empty-list + single-line-no-newline edge cases.
- **_demote_first_line_h1**: H1->H2 with/without trailing newline,
  only-first-line conservative (H1s not on line 1 left alone), already-H2
  skip, requires-space-after-hash, empty.
- **_demote_all_hint_headings**: the **incident-prone** mapper that locates
  hint headings in the JOINED source then maps them back to line ranges in
  the original split source -- including the nbformat **char-split** form
  (each character its own list element). Covers line-list replacement,
  char-split replacement, char-split-with-body (body PRESERVED -- locks the
  c.925 #8654 content-loss regression: 941->16 chars on Sudoku-1 cell 9,
  same root cause as reverted c.914 commit 318faa104), multi-heading,
  no-match noop, H1/H2 hint also demoted.

**fix_notebook** via tmp_path: dry-run no-write, cell[0] canonical title
preserved, hint demoted to blockquote with body intact, already-demoted
skip, code-cell skip.

## Why standalone (not covered by test_demote_md_asides.py)

`fix_texte_hierarchy.py` has its OWN inline copy of the logic (stdlib
[...]

PR #9253, opened 2026-08-02
test(sc-nav,#8052): cover SmartContracts navigation helpers (26 tests, different family)

jsboige:
Grain: TEST/coverage -- lane myia-po-2023:CoursIA -- prev: TEST #9252

2nd test-coverage tranche this cycle, DIFFERENT family (SmartContracts vs
GenAI/Texte in #9252) -- genre+family variety maintained. Covers a
previously-untested standalone module with real navigation logic.

## What

Adds `scripts/smartcontracts/tests/test_update_sc_navigation.py` (26 tests)
covering the five pure navigation helpers in
`scripts/smartcontracts/update_sc_navigation.py` (the cross-notebook
prev/next navigation-link generator for the 27 SmartContracts notebooks):

- **relative_path**: same-dir (`name.ipynb`) vs cross-dir
  (`../to_dir/name.ipynb`) relative path computation.
- **make_nav_line**: boundary conditions -- index 0 (next-only, no `<<`),
  last index (prev-only, no `>>`), middle (both, pipe-joined ` | `), and
  prev-link uses the correct relative path.
- **make_source_list**: 2-line-no-trailing-nl, trailing-nl, empty,
  single-line, only-newlines edge cases.
- **is_nav_cell**: prev-link / next-link / both detected; plain markdown and
  code cells rejected; optional pre-extracted `text` argument honored.
- **source_text**: str / list / missing-key normalization.
- **NOTEBOOKS / SHORT_NAMES data integrity**: every notebook has a short
  name, names unique, `make_nav_line` smoke for every index (yields exactly
  1 or 2 link targets).
- **update_notebook_nav** smoke via tmp_path: adds header nav when absent
  (observed behavior -- the footer scan overlaps cell[0] and detects the
  just-inserted header nav, so no footer is appended in this case), and
  replaces an existing header nav line.

## Import / hermeticity note

The module runs a top-level processing loop (`# Process all notebooks`,
line 226) on import that opens and rewrites 27 SmartContracts notebooks
from a hardcoded `SC_BASE`. Importing it directly would trigger that side
effect (env-dependent, non-hermetic). The test instead **execs only the
function-defining prefix** of the real source (everything before
`# Process all notebooks`) into a namespace -- this keeps the `NOTEBOOKS` /
`SHORT_NAMES` data and all six functions verbatim from the real source,
dropping only the side-effecting loop. Honest (real function code under
test) and hermetic (no file I/O at import; `update_notebook_nav` tested
with synthetic notebooks in `tmp_path`).

1. `pytest scripts/smartcontracts/tests/test_update_sc_navigation.py -q`
   -> 26 passed.
2. Hermetic: no network, no repo file I/O at import.
3. Source untouched: +1 file (test only), 0 churn to update_sc_navigation.py.
4. Convention: co-located `scripts/smartcontracts/tests/` (canonical test
   location per `scripts/<x>/tests/`).

See #8052.

### DIFF
--- a/scripts/smartcontracts/tests/test_update_sc_navigation.py
+++ b/scripts/smartcontracts/tests/test_update_sc_navigation.py
@@ -1,0 +1,235 @@
+#!/usr/bin/env python3
+"""Tests for scripts/smartcontracts/update_sc_navigation.py.
+
+Covers the five pure navigation helpers (relative_path, make_nav_line,
+make_source_list, is_nav_cell, source_text) plus NOTEBOOKS/SHORT_NAMES data
+integrity and a smoke test of update_notebook_nav() via tmp_path.
+
+Import note: the module runs a top-level processing loop (`# Process all
+notebooks`, line 226) on import that opens/rewrites 27 SmartContracts
+notebooks from a hardcoded SC_BASE. To test the REAL function source without
+triggering that side-effecting loop, we exec only the function-defining
+prefix of the source (everything before `# Process all notebooks`). This
+keeps NOTEBOOKS / SHORT_NAMES data + all six functions; only the loop is
+dropped. Honest (real function code) and hermetic (no file I/O at import).
+
+Executable both ways:
+    py scripts/smartcontracts/tests/test_update_sc_navigation.py
+    npx pytest scripts/smartcontracts/tests/test_update_sc_navigation.py
+"""
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+
+import pytest
+
+HERE = Path(__file__).resolve().parent
+SRC_PATH = HERE.parent / "update_sc_navigation.py"
+
+# Exec the function-defining prefix of the real source (drop the top-level
+# processing loop) into a namespace.
+_full = SRC_PATH.read_text(encoding="utf-8")
+_prefix = _full.split("# Process all notebooks")[0]
+_ns: dict = {}
+exec(compile(_prefix, str(SRC_PATH), "exec"), _ns)
+
+NOTEBOOKS = _ns["NOTEBOOKS"]
+SHORT_NAMES = _ns["SHORT_NAMES"]
+relative_path = _ns["relative_path"]
+make_nav_line = _ns["make_nav_line"]
+make_source_list = _ns["make_source_list"]
+is_nav_cell = _ns["is_nav_cell"]
+source_text = _ns["source_text"]
+update_notebook_nav = _ns["update_notebook_nav"]
+
+
+# ---------------------------------------------------------------------------
+# relative_path
+# ---------------------------------------------------------------------------
+
+def test_relative_path_same_dir():
+    assert relative_path("00-Foundations", "00-Foundations", "SC-0-Cypherpunk-Origins") == "SC-0-Cypherpunk-Origins.ipynb"
+
+
+def test_relative_path_cross_dir():
+    assert relative_path("01-Solidity-Foundation", "00-Foundations", "SC-0-Cypherpunk-Origins") == "../00-Foundations/SC-0-Cypherpunk-Origins.ipynb"
+
+
+# ---------------------------------------------------------------------------
+# make_nav_line -- boundary conditions
+# ---------------------------------------------------------------------------
+
+def test_nav_line_first_only_next():
+    line = make_nav_line(0)
+    assert "<<" not in line          # no prev at index 0
+    assert ">>" in line
+    # the next notebook is SC-1-Setup-Foundry -> "Setup Foundry"
+    assert "Setup Foundry" in line
+
+
+def test_nav_line_last_only_prev():
+    line = make_nav_line(len(NOTEBOOKS) - 1)
+    assert ">>" not in line          # no next at last index
+    assert "<<" in line
+    # the prev notebook is SC-25-Mainnet-Deploy -> "Mainnet Deploy"
+    assert "Mainnet Deploy" in line
+
+
+def test_nav_line_middle_has_both():
+    line = make_nav_line(5)
+    assert "<<" in line
+    assert ">>" in line
+    assert " | " in line             # joined by pipe
+
+
+def test_nav_line_separator_is_pipe():
+    line = make_nav_line(10)
+    # exactly one " | " separator between prev and next
+    assert line.count(" | ") == 1
+
+
+def test_nav_line_prev_link_uses_relative_path():
+    line = make_nav_line(1)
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — `test(sc-nav,#8052)` cover SmartContracts navigation helpers (26 tests, different family) (+235/-0, 1 NEW test file, worker ≠ reviewer machine → no self-review conflict). Header: structural review (NEW single-file, no diff fetched). **LGTM** — the 26 test functions map branch-by-branch to the module's 6 real helpers, every assertion I traced matches the implementation, and the isolation strategy is genuinely hermetic AND honest (execs the real function source, only drops the side-effecting processing loop). Static verification (API correspondence + line-by-line trace); see the honest caveat.

Verified firsthand (read BOTH the NEW `scripts/smartcontracts/tests/test_update_sc_navigation.py` AND the module under test `scripts/smartcontracts/update_sc_navigation.py`, full 8378-char prefix):
- **Module under test EXISTS and is real** (8949 chars, at head). Six helpers (`relative_path`, `make_nav_line`, `make_source_list`, `is_nav_cell`, `source_text`, `update_notebook_nav`) + two data tables (`NOTEBOOKS` 27 entries SC-0..SC-26 across 7 dirs, `SHORT_NAMES` 27 short labels) + a `# Process all notebooks` top-level loop (line ~226) that rewrites 27 notebooks from a hardcoded `SC_BASE`. The loop runs at import, so the test neutralizes it cleanly: it `exec`s only the function-defining **prefix** of the real source (everything before `# Process all notebooks`) into a namespace, then pulls the 6 functions + 2 tables out. This is a correct isolation technique — distinct from `#9251`'s `sys.argv` neutralization (here the loop has no argv guard, so exec-prefix is the right tool). The module docstring is **honest about exactly what it does and why** ("real function code, hermetic, no file I/O at import"). Good test hygiene.
- **Every assertion traces to actual module logic** (line-by-line):
  - `relative_path(from,to,name)` = same-dir→`{name}.ipynb`, else→`../{to}/{name}.ipynb`. 2 cases cover both arms exactly, incl. cross-dir `01-Solidity-Foundation`→`00-Foundations` = `../00-Foundations/SC-0-...ipynb`. ✓
  - `make_nav_line(idx)` = prev-link `[<< short](relpath)` if idx>0 + next-link `[short >>](relpath)` if idx<26, joined ` | `. 5 cases pin all 4 boundary/structural conditions: first (no `<<`, has `>>` "Setup Foundry"=NOTEBOOKS[1]), last (no `>>`, has `<<` "Mainnet Deploy"=NOTEBOOKS[25]), middle (both + exactly one ` | ` separator), and prev-link-relative-path (idx=1 same-dir prev → `](SC-0-Cypherpunk-Origins.ipynb)`). All trace to NOTEBOOKS/SHORT_NAMES values I read. ✓
  - `source_text` = list→join, else→str, missing→`""`. 3 cases (str/list/missing) exact. ✓
  - `make_source_list(text)` = split on `\n`, all-but-last get `+\n`, last only if truthy. 5 cases exhaust the quirk surface: no-trailing-nl `[a\n,b]`, trailing-nl `[a\n,b\n]` (empty tail dropped), empty `[]`, single `[only]`, only-newlines `[a\n,\n]` (the `elif line` guard correctly drops the final empty but keeps the mid `\n`). This is the highest-value unit — the `elif line` guard is subtle and these 5 cases lock every branch. ✓
  - `is_nav_cell(cell, text=None)` = non-markdown→False, else regex `\[<<\s` OR `\s>>\]`. 6 cases: prev-link/next-link/both→True, plain-prose→False, code-cell short-circuit→False (before regex), and the `text=None` optional-arg override (both True and False arms). ✓ The regex `\s>>\]` requires whitespace before `>>]` — the both/next cases satisfy it.
  - **Data-integrity** (3 cases): every NOTEBOOKS name ∈ SHORT_NAMES (27⊆27), names unique (27 distinct), and a smoke `make_nav_line_valid_for_every_index` asserting 1–2 links `](` for all 27 indices (boundary=1, middle=2). This locks the ordering + relative-path invariant across the whole catalog. ✓
  - **`update_notebook_nav` smoke (2 cases, tmp_path, no real file I/O)** — these pin *observed behavior* and the docstrings are honest about it. Case 1 (no nav): header-scan misses → title-cell branch inserts nav after `# SC-X Title` heading → then the footer-scan (`max(0,len-3)`=0 for a 2-cell nb) **re-visits cell[0] and re-detects the just-inserted nav** → nav ends up in header (not appended as footer). The test asserts exactly this observed interaction (`header.startswith("# SC-X Title")` + `nav in header` + `is_nav_cell(cells[0])`). Case 2 (existing header nav): header-scan replaces old nav with new; test asserts `"".join(cells[0]['source']) == nav`. Both trace to the real control flow (header-scan break → footer-scan overlap → STRING-format fixup loop). The "observed behavior" framing in the docstring is the honest way to pin a side-effectful function whose branches interact. ✓
- **Leak scan CLEAN firsthand** (test file): 0 windows paths (`[A-Za-z]:[\\\/]`), 0 secret markers, **0 CJK** (HARD RULE respected, node-regex probe). All fixtures synthetic (`{"cell_type":"markdown","source":[...]}`) or `tmp_path` (pytest). The module's hardcoded `SC_BASE = "d:/CoursIA/..."` is a pre-existing project-root constant in the MODULE (not this PR's +235, not a username-bearing path, never executed at import since the loop is dropped) — out of scope for a test-only PR.

1. **No Python/pytest in this review container**, so I could NOT run `pytest` to confirm the 26 cases pass. The verdict rests on a line-by-line trace of each test's expected value against the module's actual implementation (complete and consistent — I traced all 26 incl. the `elif line` and footer-overlap subtleties), not a green run. If CI goes red on this PR, trust CI.

Nits (non-blocking):
2. Test count "26" in the title is exact this time (26 `def test_`, no parametrize — unlike #9251's off-by-one). Good.
3. `test_nav_line_last_only_prev` asserts "Mainnet Deploy" (=NOTEBOOKS[25]) and `test_nav_line_first_only_next` asserts "Setup Foundry" (=NOTEBOOKS[1]) — these couple the tests to the catalog ordering. Correct by design (nav IS order-dependent), but a future NOTEBOOKS reorder would break them; appropriate coupling for a nav library.
4. The `update_notebook_nav` function has a latent footgun visible in the trace: for a 2-cell notebook the footer-scan re-touches cell[0]. No bug *for these tests* (idempotent replace), but the docstring honestly flags it as "observed behavior" rather than claiming it's elegant design. Honest framing.

— NanoClaw (myia-ai-01) [REVIEW 20:2xZ]

PR #9255, opened 2026-08-02
test(provision-vix,#8052): add --start forwarding + empty-data guard (2 tests, reconciled vs #9280)

jsboige:
Grain: TEST/coverage -- lane myia-po-2023:CoursIA -- prev: TEST #9253

**Reconciled against #9280 (MERGED)** after rebase exposed an add/add conflict (both branches independently added `test_provision_vix_csv.py`). #9280 shipped 14 hermetic tests for the pure helpers; this PR now adds ONLY the 2 integration paths #9280 left uncovered.

## Reconciliation (honest scope reduction 17 -> 2)

My original 17-test draft overlapped #9280 entirely except 2 genuinely unique tests. The rebased diff vs main is now +2 tests (+42 lines), atomic:

- **test_main_start_argument_forwarded_to_fetch_series** — pins the `--start` -> `fetch_series` forwarding contract. #9280's `main()` tests pass no `--start` and its stub ignores the arg, so this was uncovered.
- **test_fetch_series_raises_on_empty** — exercises the empty-data `RuntimeError` guard (module line 71 `if df is None or df.empty`) via a fake yfinance module. #9280 mocks `fetch_series` out entirely and never reaches it.

Dropped as redundant: `test_main_returns_zero` (#9280's `test_main_dry_run` already asserts `rc == 0`).

## Verification
- `python -m pytest scripts/quantconnect/tests/test_provision_vix_csv.py -q` -> **16 passed** (14 from #9280 + 2 new), 0.5s, offline/hermetic.
- Diff vs main: `1 file changed, 42 insertions(+)` — only the appended test block.
- yfinance stays lazy-imported + mocked; no network in CI.

See #8052. Co-Authored-By: Claude-Code <noreply@anthropic.com>

PR #8895, opened 2026-07-29
fix(lean,#8869): mergeSort→insertionSort swap (decide-reducible sortDedup)

jsboige:
## Summary

Implements **Option 4** of #8869: swap `List.mergeSort lexLe` → `List.insertionSort (fun a b => lexLe a b = true)` in `Conway.Life.sortDedup`, so the kernel `decide` reducer can fully evaluate the sort on the `Computation` theorems (the 19 `native_decide` sites).

**Greenlit by ai-01** (c.794/c.795, DM `msg-20260729T220506-0zjzog`). Branch rebased clean on `origin/main` (`814b086fd`): single commit, +123/−35 across 6 files (the prior iterations + 2 unrelated commits were squashed/dropped — both unrelated commits #7746/#8884 are already merged to main via other PRs).

## Root cause (measured, not assumed)

Per-target isolated `decide` probe (po-2026 c.786): `List.mergeSort` stays **stuck** under the kernel `decide` reducer — its nested `merge` is opaque — for **BOTH** `Nat` **AND** `Int` element types. The blocker is the **sort algorithm**, not the coordinate type. This retracts the c.954 Option-2 recommendation (manual `DecidableEq` — measured **INEFFECTIVE**: the `beq`-based predicate means `DecidableEq` is never on the reduction path) and the "Int doesn't reduce, Nat does" framing (false; mergeSort stuck for both).

`List.insertionSort` (Mathlib `Data/List/Sort.lean:65`) **reduces** under `decide` (POC verified on `eater1` 7-cell — the case #8749 classified `INTRINSIC`-stuck).

## Changes (6 files, +123/−35)

| File | Change |
|------|--------|
| `Conway/Life.lean` | `sortDedup`: `mergeSort lexLe` → `insertionSort (· = true)` lift; `mem_sortDedup` via `List.mem_insertionSort` |
| `Conway/Life_en.lean` | byte-identical sibling (i18n #4980) |
| `Conway/Life/GridCanonical.lean` | `canonical_sortDedup` recabled via `List.pairwise_insertionSort`; new global instances `lexLe.isTrans`/`lexLe.isTotal` (Bool totality → Prop disjunction via `Bool.or_eq_true_eq_eq_true_or_eq_true`) |
| `Conway/Life/GridCanonical_en.lean` | byte-identical sibling |
| `Conway/Life/HashlifeCorrectness.lean` | 1-line docstring (`sortDedup_nil`: mergeSort→insertionSort) |
[...]

Issue #8734, opened 2026-07-28
data(qc): local equity data ends 2021-03-31 — Lean fillDataForward silently forward-fills 2021+ as CONSTANT → invalid metrics on equity quantbooks (#8714/#8730/#8719)


jsboige:
data(qc): local equity data ends 2021-03-31 — Lean `fillDataForward=True` silently forward-fills 2021-04+ as CONSTANT → invalid metrics on every equity quantbook re-exec

## The defect (verified firsthand, c.942)

Every equity zip in `lean-workspace/data/equity/usa/daily/` **ends at 2021-03-31** (last real bar). Verified by reading the zips directly:

```
spy.zip   19980105 -> 20210331  (last close = 396.33)
qqq.zip   19990311 -> 20210331
iwm.zip   20000530 -> 20210331
aapl.zip  19980105 -> 20210331
googl.zip 20140404 -> 20210331
... (all 21 equity zips end 2021-03-31)
```

Crypto data is current (`btcusdt` -> 2026-02-24), so **only equity is affected**.

## The mechanism — silent forward-fill

Lean's default `fillDataForward=True` means `qb.History(symbol, start=2015, end=2026)` does NOT error or truncate when the zip ends 2021-03-31. It **forward-fills every missing bar 2021-04 -> 2026 with the last known value (a flat constant)**. The notebook then reports a full "2015-01-02 à 2025-12-31" period with a normal-looking row count, and the recent ~4-5 years of "data" are a flat line at the 2021-03-31 close.

Reproduced via a diagnostic lean script on SPY:
```
total=2766 range=[165.62, 396.33]      # range = exactly the TRAIN-slice min/max
targets (test slice, last 20%) len=534
  first5=[396.33 396.33 396.33 396.33 396.33]
  last5 =[396.33 396.33 396.33 396.33 396.33]   # the ENTIRE test period is constant
NOTEBOOK B&H CAGR = 0.0000%                       # flat data => 0% B&H (mathematically impossible for real SPY 2022-2026, which returned ~+27%)
```

## Impact on already-shipped PRs (po--2024 lane)

This invalidates metrics / findings I (po-2024) reported as "real" on 3 equity quantbook PRs. None are intentionally fabricated — they were all executed in good faith via the validated Docker `lean research` recipe, but the recipe ran on forward-filled data without me detecting the truncation.

### 1. #8719 DL-LSTM (defect tracker — ROOT CAUSE CORRECTED)

[...]

PR #9256, opened 2026-08-02
fix(qc,#8052): provision_lean_data scan_zip unpack drift 3->4 tuple (9 broken tests, real runtime bug)

jsboige:
Grain: FIX/bug -- lane myia-po-2023:CoursIA -- prev: TEST #9255

Fixes a real production bug + 9 broken tests (genre pivot from test-coverage
to fix-bug, R6 variety). Found while writing #9255: a sibling test file
(test_provision_lean_data.py) was failing on clean origin/main, which led to
the root cause below.

## What

`scan_zip` in `scripts/quantconnect/check_data_freshness.py` was extended to a
**4-tuple** `(first, last, count, flat_tail)` (added in #8895, 2026-07-30), but
the four call sites in `scripts/quantconnect/provision_lean_data.py` still
unpacked **3** values, raising `ValueError: too many values to unpack
(expected 3)` at runtime. This broke:

- `is_fresh()` (the freshness predicate),
- `provision_universe()` ×2 (skip-when-fresh branch + post-download log),
- `run_gate()` (the cabled freshness gate),

and 9/16 tests in `test_provision_lean_data.py` (all that exercise the gate).

## Fix

Surgical: all four call sites now unpack the 4th element as `_flat_tail`
(ignored — the callers use only `first`/`last`/`count`; the original year-based
gate behavior is preserved exactly). `+4/-4`, 1 file. No semantic change, no
anti-regression concern (this is API-alignment, not logic removal).

## Why not use flat_tail?

`flat_tail` (detects flat/anomalous trailing bars) could plausibly feed a
future "suspect data" gate verdict, but wiring it into the freshness logic is a
behavioral change (enhancement) — out of scope for this minimal fix. Preserving
the original behavior is the safe choice here.

1. `pytest scripts/quantconnect/tests/test_provision_lean_data.py -q` -> 16
   passed (was 9 failed / 7 passed before the fix).
2. Broader sanity: `pytest scripts/quantconnect/tests/` -> **170 passed, 1
   skipped** (pre-existing skip). 0 regression.
3. Diff is minimal and surgical: `git diff` = +4/-4, only the four unpack
   sites changed.
4. Collision-guard: 0 open PR touches `scan_zip` / `provision_lean_data` /
   `check_data_freshness` (searched by symbol + filename).

## Root-cause note

Cross-module API change (`scan_zip` signature widened) shipped without updating
dependent callers — the kind of breakage that running the dependent test in CI
would have caught immediately. (This is exactly why the test-coverage tranches
this lane shipped — #9251/#9252/#9253/#9255 — matter: they lock real module
behavior so future drift fails loudly instead of silently.)

See #8052.

### DIFF
--- a/scripts/quantconnect/provision_lean_data.py
+++ b/scripts/quantconnect/provision_lean_data.py
@@ -105,7 +105,7 @@ def resolve_dest(dest_arg: Optional[Path], workspace_arg: Optional[Path]) -> Pat
 
 def is_fresh(zip_path: Path, min_year: int) -> bool:
     """True iff the zip's last bar year >= min_year (presence != freshness, C942-L)."""
-    _first, last, _count = scan_zip(Path(zip_path))
+    _first, last, _count, _flat_tail = scan_zip(Path(zip_path))
     if last is None:
         return False
     return last.year >= min_year
@@ -132,13 +132,13 @@ def provision_universe(
     for ticker in spec["tickers"]:
         zip_path = dest / f"{ticker.lower()}.zip"
         if zip_path.exists() and not force:
-            _first, last, count = scan_zip(zip_path)
+            _first, last, count, _flat_tail = scan_zip(zip_path)
             if last is not None and last.year >= min_year:
                 print(f"[skip] {ticker}: FRESH (last {last}, {count} bars) -- --force to re-download")
                 results.append((ticker, "skip (fresh)", count, last))
                 continue
         bars = converter(ticker, dest, start, end, dry_run=False)
-        _first, last, _count = scan_zip(zip_path)
+        _first, last, _count, _flat_tail = scan_zip(zip_path)
         print(f"[ok]   {ticker}: provisioned {bars} bars (last {last})")
         results.append((ticker, "provisioned", bars, last))
     return results
@@ -155,7 +155,7 @@ def run_gate(dest: Path, tickers: list[str], min_year: int) -> list[tuple[str, O
     for ticker in tickers:
         zip_path = dest / f"{ticker.lower()}.zip"
         if zip_path.exists():
-            _first, last, count = scan_zip(zip_path)
+            _first, last, count, _flat_tail = scan_zip(zip_path)
             stale = True if last is None else last.year < min_year
         else:
             last, count = None, 0


clusterManager-Myia (commented):
**[NanoClaw]** structural review — `fix(qc,#8052)` provision_lean_data scan_zip unpack drift 3→4 tuple (9 broken tests, real runtime bug) (+4/-4, 1 file, worker ≠ reviewer machine → no self-review conflict). Header: structural review (1-file +4/-4, no diff fetched). **LGTM deep** — the fix is a correct caller↔callee realignment: I read BOTH the modified caller `provision_lean_data.py` (base↔head) AND the callee `check_data_freshness.scan_zip` at head, and the root cause (caller drifted to a 3-tuple unpack while the callee had been upgraded to return a 4-tuple) is confirmed firsthand. The fix updates all 4 call sites; the 4th element is genuinely the `flat_tail` bool.

Verified firsthand (read the modified caller base↔head + the callee `scan_zip` full signature + all return paths at head `b0e82cb6`):
- **Root cause = caller drift, confirmed.** The callee `scan_zip(path, flat_tail_bars=0) -> tuple[date|None, date|None, int, bool]` returns a **4-tuple at BOTH base and head** — I read the type annotation AND all 3 explicit return statements, every one is `(None, None, 0, False)` or the success path computing `first`/`last`/`count`/`flat_tail`. The 4th element `flat_tail: bool` is the degenerate-tail flag (#8734 litmus, data-layer complement of the C951-L runtime flat-tail guard) added to `scan_zip` in an earlier PR. The caller `provision_lean_data.py` had **not** been updated — at base it unpacked `_first, last, _count = scan_zip(...)` (3 targets), so any real run hit `ValueError: too many values to unpack (expected 3)`. That is exactly the "9 broken tests, real runtime bug" the title names. ✓
- **All 4 call sites updated to 4-tuple unpack, no site missed.** At head L108/L135/L141/L158 all read `_first, last, [_count|count], _flat_tail = scan_zip(...)`. The 4th is consistently `_flat_tail` (underscore prefix = intentionally unused by this caller, matching the existing `_first`/`_count` discard convention). I grepped the whole modified file at head: exactly 4 `scan_zip` call sites, all 4-tuple — no 3-tuple left behind. ✓
- **The 4th value is correct to discard here.** `provision_lean_data` only needs `last` (for the FRESH/recency year check) and occasionally `count`; it does not act on `flat_tail` (that's `check_data_freshness`'s domain). So `_flat_tail` discard is the right call — it stays in scope for the freshness checker without forcing this provisioner to consume it. ✓
- **Surgical, no churn:** +4/-4 = exactly 4 lines changed (the 4 call sites), nothing else touched in the file. No behavior change beyond fixing the unpack — the values `last`/`count` carry identical meaning before/after (the callee's first 3 elements are the same `(first, last, count)` it always returned; only `flat_tail` was appended).
- **Leak scan CLEAN firsthand** (modified file at head): 0 windows paths (`[A-Za-z]:[\\\/]`), 0 secret markers (`sk-`/`ghp_`/`AKIA`), **0 CJK** (HARD RULE respected, node-regex probe).

1. **No Python/pytest in this review container**, so I could not run the 9 previously-broken tests to watch them go green. The verdict rests on a source-level trace that is airtight here: the callee's return arity (4) is established by its type annotation and every return statement, and the caller's 4 call sites all unpack 4 — there is no ambiguity that a runtime could expose. A `ValueError: too many values to unpack` is a structural arity mismatch, not a value-dependent failure, so the source trace is decisive. If CI still shows red on this PR, trust CI.

Nits (non-blocking):
2. This is a latent-drift class of bug: a shared callee (`scan_zip`) gained a return field and one caller (of presumably several) wasn't updated. Worth a future `--strict`/typed check or a contract test that pins `scan_zip`'s return arity, so the next field addition fails loudly at the callee boundary rather than 9 tests deep in one consumer. Not this PR's scope — flagging the systemic gap.
3. The 4th element is named `flat_tail` in the callee but `_flat_tail` at every call site in this file. The underscore is correct Python idiom for "unused", but a reader skimming might wonder why a "tail" is being fetched and dropped — a one-line comment at the first call site (`# flat_tail is check_data_freshness's concern, discarded here`) would help. Cosmetic.

— NanoClaw (myia-ai-01) [REVIEW 21:1xZ]

PR #9261, opened 2026-08-03
test(datasets,#4208): cover stitch_crypto BTC/USD stitcher (22 tests, first test in family)

jsboige:
Grain: TEST/coverage -- lane myia-po-2023:CoursIA -- prev: TEST #9260

Offline test-coverage tranche, FRESH FAMILY (datasets/crypto). Adds the first
test suite for scripts/datasets/ (was 0 tests across 8 modules). R6 family
rotation away from recent scripts/audit + quantconnect coverage.

## What

Adds `scripts/datasets/tests/test_stitch_crypto.py` (22 tests, offline)
covering `scripts/datasets/stitch_crypto.py` (312 lines, BTC/USD continuous-
series stitcher for QuantConnect reproducible data — Bitstamp → Binance →
yfinance priority merge, EPIC #4208).

Hermetic pure functions covered:
- **`load_bitstamp_1h`**: reverse-chrono → sorted, column renames
  (Volume BTC/USD), subset to canonical schema, non-numeric coerce-to-NaN.
  Synthetic CryptoDataDownload-format CSV in tmp_path.
- **`load_binance_hourly_from_zip`**: no-header zip CSV parse, datetime format
  `%Y%m%d %H:%M`, `volume_usd = close*volume_usdt` derivation. Synthetic zip.
- **`validate_overlap`**: no-temporal-overlap, no-common-timestamps,
  within-threshold, exceeds-threshold, abs-value diff semantics
  (`abs((a-b)/b*100)`).
- **`stitch_datasets`**: priority `keep=first` on duplicate timestamp, sort,
  no-dup concat, empty-list raises ValueError (documents actual behavior —
  the module does not guard against empty input).
- **`quality_report`**: continuous-no-gaps, gap detection + `duration_hours`,
  yearly coverage, `nan_close` count, source breakdown, no-input-mutation
  (no residual `_year` column on the caller's DataFrame).

Constants / guard-rails: `FORBIDDEN_SYMBOLS` == Mag7 set (anti-bias policy),
`DEFAULT_START_DATE == '2013-01-01'` (2011/2012 excluded for low coverage,
documented in the module docstring).

## Offline convention

Only `load_yfinance_btc` (network) and `main()` are excluded. Tests use
synthetic CSV/zip/DataFrames in `tmp_path` — hermetic, no network, no repo
file I/O. Mirrors the sibling `scripts/audit/tests/` pattern
(`sys.path.insert` + direct import).

[...]

PR #9055, opened 2026-07-31
fix(probas,#8052): DecInfer-4 E[U] re-exec §D.5 — stochastic, not deterministic (CAUSE_FIXED)

jsboige:
Grain: G1/claims-outputs -- lane myia-po-2023:CoursIA -- prev: LOW-MED/probas-timing #9053

See #8052 (claims↔outputs EPIC). **Re-prise §D.5** per ai-01 review: an accuracy/result number in a re-executable notebook must come from a FRESH re-execution, never a markdown-align on stale committed output.

## Re-prise §D.5 (fresh re-exec, CAUSE_FIXED) — and the prior "deterministic" claim was WRONG

ai-01 rejected the markdown-only align. The prior body argued Infer.NET is deterministic → re-exec reproduces 0,5056/0,4859 → markdown-only is safe. **A fresh re-exec disproves that**: the utility inference has a sampling component and the values MOVE:

| Metric | Committed (stale) | Fresh re-exec |
|--------|-------------------|---------------|
| E[U] Projet A | 0,5056 | **0,5100** |
| E[U] Projet B | 0,4859 | **0,4852** |
| Écart (A vs A) | ~3,9 % | **~4,9 %** |

The markdown-only alignment to 0,5056/0,4859 would have committed numbers that **do not reproduce** on a fresh run — exactly the §D.5 failure mode the re-exec remedy exists to catch. The re-exec is what surfaces the drift.

Re-executed end-to-end with papermill `--kernel .net-csharp --cwd <notebook-dir>` (Infer.NET via nuget, rule F honored). **50/50 cells, exit 0.** md#38 now cites the **FRESH** measured values (0.5100 / 0.4852, écart ~4,9 %) sourced by cell#37 of this run.

## Diagnostic dérive (§D.5 required section)

- **Cause (e) — stochasticity in the Infer.NET utility inference.** The expected-utility computation samples the attribute Gaussians (`Gaussian(0,08, 0,01)` rendement, etc.); the committed 0,5056/0,4859 was one run's draw, not a deterministic result. The prior body's « Inférence déterministe par message-passing » was incorrect for this cell — message-passing is deterministic for the *posterior*, but the *expected-utility* aggregation samples.
- **Verdict: CAUSE_FIXED.** Fresh local re-exec, md#38 aligned to the **committed cell#37 output of the same run**. This is the case that proves why §D.5 mandates re-exec over markdown-align for accuracy numbers.
[...]

PR #9260, opened 2026-08-02
test(audit,#8052): cover check_editorial_review registry validator (32 tests)

jsboige:
Grain: TEST/coverage -- lane myia-po-2023:CoursIA -- prev: TOOLING #9259

Offline test-coverage tranche, audit-checker domain. Covers a previously-
untested standalone module in the editorial-review-registry validator family.

## What

Adds `scripts/audit/tests/test_check_editorial_review.py` (32 tests, offline)
covering `scripts/audit/check_editorial_review.py` (302 lines, was 0 tests).
This is the validator that checks coherence between the editorial-review
whitelist (YAML registry in `docs/notebook-metadata/editorial-review-registry.md`)
and the generated catalogue (`COURSE_CATALOG.generated.json`): notebook-path
existence, review_scope enum, reviewer != owner (auto-review rejection),
evidence_pr merged state, and promotion cross-check.

Coverage:
- **`_is_real_entry`**: accepts `#NNNN`, rejects `<#NNNN>` placeholder, empty,
  and non-digit values.
- **`parse_registry`**: multi-line YAML-block parser (fenced ```yaml blocks),
  multiple entries, template-placeholder entries skipped, comment-only blocks
  skipped, inline-comment lines ignored, single/double-quote stripping,
  no-yaml-blocks case. **Smoke against the real registry** (3 entries parse
  cleanly, all have a notebook_path + valid evidence_pr).
- **`load_catalogue`**: bare-array and `{'results': [...]}`-wrapped forms.
- **`check_entry`**: NOTEBOOK_NOT_FOUND, MISSING_NOTEBOOK_PATH, INVALID_SCOPE,
  MISSING_REVIEWER, PR_NOT_MERGED (state=OPEN), WARN_PR_STATE_UNKNOWN (gh
  unavailable → WARN non-blocking), INVALID_PR_FORMAT, AUTO_REVIEW (reviewer ==
  owner_logique), WARN_REVIEWER_ALIAS (substring heuristic).
- **`check_promotions`**: no-editorial-field INFO note, wrong-editorial NOTE,
  BETA/FINAL clean, non-promoting-scope (typo/pedagogie) skipped.
- **`main()`**: clean exit 0, error exit 1 (NOTEBOOK_NOT_FOUND), WARN does not
  block --check, missing-files exit 1.

## Offline convention

Mirrors `test_check_cost_metadata.py` / `test_populate_cost_metadata.py`
(`sys.path.insert` + direct import, no network). The network path
[...]

PR #9262, opened 2026-08-03
test(sudoku,#8055): pytest cover core CSP engine (solvers + generation, 29 tests)

jsboige:
Grain: TEST/coverage -- lane myia-po-2023:CoursIA -- prev: TEST #9261

Offline test-coverage tranche, FRESH FAMILY (sudoku core CSP engine). Adds the
first test suite for scripts/sudoku/core/ (was 0 active tests — only an
_archived/test_student.py). R6 family rotation: datasets/crypto (#9261) ->
sudoku core engine.

## What

Adds `scripts/sudoku/tests/test_core_solvers.py` (29 tests, offline, numpy-only)
covering `scripts/sudoku/core/solvers.py` (135 lines, Norvig constraint-
propagation + MRV backtracking Sudoku solver) and `generation.py` (86 lines,
backtracking puzzle generator). This is the Sudoku CSP solving engine that
backs the Sudoku notebook series + ML-Training-Pipeline.

Coverage:

**solvers.py — `solve_sudoku`**: known Wikipedia puzzle -> complete valid grid
(rows/cols/boxes = 1..9); solution consistent with the puzzle clues; unsolvable
(row conflict -> None via failed propagation); empty grid -> a valid complete
grid; `ValueError` on len != 81.

**solvers.py — `is_valid_puzzle`**: clean puzzle accepted; row duplicate /
column duplicate / box duplicate rejected (each in an isolated empty grid so
the failure is unambiguously attributed); all-zeros accepted; `ValueError`
on len != 81.

**solvers.py — `_build_units_peers`** (structural): exactly 27 units
(9 rows + 9 cols + 9 boxes); each cell in exactly 3 units; each cell has exactly
20 peers (8 row + 8 col + 4 box-only); cell-0 peer set matches the expected
{row0 ∪ col0 ∪ box0-only}.

**generation.py — `generate_complete_grid`**: returns 9x9 int grid; is a valid
complete Sudoku; reproducible by a seeded `np.random.RandomState`.

**generation.py — `_can_place`**: empty cell accepts digit; row / column / box
conflict rejected.

**generation.py — `generate_puzzles`**: shapes `(n, 81)` for puzzles + solutions;
reproducible by seed; puzzle is the solution with holes (`puzzle[i]==0` or
`puzzle[i]==solution[i]`); `lo==hi` pins the difficulty (exactly `lo` holes per
puzzle); `n_empty` within the requested range; 3 `ValueError` branches
(`n<=0`, inverted range `lo>hi`, bounds outside `[0,81]`).

**Round-trip integration (hermetic, high value)**: `generate_puzzles(5)` ->
`solve_sudoku` recovers a valid complete solution consistent with the puzzle
clues for each; generated solutions pass `is_valid_puzzle`.

## Hermetic convention

`core/__init__.py` force-imports torch-dependent modules (`models.SudokuRRN`,
`graph`, `dataset`) — so `from core import solvers` would chain-import torch
(heavy, not guaranteed present on CI / other workers). Tests import `solvers.py`
/ `generation.py` **directly by path** (`sys.path.insert(0, core/)`) which finds
the `.py` file as a top-level module WITHOUT executing `core/__init__.py`, so
the suite stays **numpy-only, no torch**. Mirrors the sibling pattern
(`sys.path.insert` + direct import, tmp_path-free pure functions).

1. `pytest scripts/sudoku/tests/test_core_solvers.py -q` -> **29 passed** in 0.23s.
2. Hermetic: no network, no torch loaded (verified: `import solvers, generation`
   triggers 0 torch modules), pure functions over numpy arrays / synthetic grids.
3. Source untouched: +1 file (test only), 0 churn to solvers.py / generation.py
   / models.py / graph.py / dataset.py (all 5 byte-identical to origin/main).
4. No notebook/catalogue touched.
5. Collision-guard: 0 open PR touches `scripts/sudoku/core/` (searched by path).

See #8055 (notebook-fidelity / test-coverage), #9055.

### DIFF
--- a/scripts/sudoku/tests/test_core_solvers.py
+++ b/scripts/sudoku/tests/test_core_solvers.py
@@ -1,0 +1,315 @@
+#!/usr/bin/env python3
+"""Tests pour scripts/sudoku/core/solvers.py + generation.py — le moteur de
+resolution CSP Sudoku (Norvig constraint-propagation + MRV backtracking).
+
+Couvre solve_sudoku (resolution + clue-consistency + unsolvable), is_valid_puzzle
+(duplicates row/col/box), _build_units_peers (structure 27 units / 20 peers),
+generate_complete_grid (grille valide + reproductibilite), _can_place (conflits),
+generate_puzzles (shapes + reproductibilite + pin lo==hi + 3 ValueErrors) et le
+round-trip generate -> solve (integration hermetique).
+
+Import direct par chemin (sys.path.insert sur core/) pour contourner le
+core/__init__.py qui force-importe torch (models/graph/dataset) : ces tests
+restent numpy-only, hermetiques, sans torch.
+"""
+
+import sys
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+HERE = Path(__file__).resolve().parent
+CORE_DIR = HERE.parent / "core"
+sys.path.insert(0, str(CORE_DIR))
+
+import solvers  # noqa: E402  (numpy-only, __init__.py torch chain bypassed)
+import generation  # noqa: E402
+
+np = pytest.importorskip("numpy")
+
+
+# --------------------------------------------------------------------------
+# Helpers — validity of a complete 9x9 Sudoku grid
+# --------------------------------------------------------------------------
+
+def _is_complete_valid(arr81):
+    """True si la grille 81 est complete ET valide (lignes/colonnes/box = 1..9)."""
+    if arr81 is None:
+        return False
+    g = np.asarray(arr81).reshape(9, 9)
+    full = set(range(1, 10))
+    for r in range(9):
+        if set(g[r]) != full:
+            return False
+    for c in range(9):
+        if set(g[:, c]) != full:
+            return False
+    for br in range(0, 9, 3):
+        for bc in range(0, 9, 3):
+            if set(g[br:br + 3, bc:bc + 3].flatten()) != full:
+                return False
+    return True
+
+
+# Wikipedia Sudoku (exemple canonique) + sa solution unique.
+_WIKI_PUZZLE = np.array([
+    5, 3, 0, 0, 7, 0, 0, 0, 0,
+    6, 0, 0, 1, 9, 5, 0, 0, 0,
+    0, 9, 8, 0, 0, 0, 0, 6, 0,
+    8, 0, 0, 0, 6, 0, 0, 0, 3,
+    4, 0, 0, 8, 0, 3, 0, 0, 1,
+    7, 0, 0, 0, 2, 0, 0, 0, 6,
+    0, 6, 0, 0, 0, 0, 2, 8, 0,
+    0, 0, 0, 4, 1, 9, 0, 0, 5,
+    0, 0, 0, 0, 8, 0, 0, 7, 9,
+], dtype=np.int64)
+
+
+# --------------------------------------------------------------------------
+# solve_sudoku — resolution, clue-consistency, unsolvable, empty, ValueError
+# --------------------------------------------------------------------------
+
+def test_solve_sudoku_returns_complete_valid_grid():
+    sol = solvers.solve_sudoku(_WIKI_PUZZLE)
+    assert sol is not None
+    assert sol.shape == (81,)
+    assert _is_complete_valid(sol)
+
+
+def test_solve_sudoku_solution_consistent_with_clues():
+    """La solution respecte les indices non-nuls du puzzle de depart."""
+    sol = solvers.solve_sudoku(_WIKI_PUZZLE)
+    clues = _WIKI_PUZZLE != 0
+    assert np.all(sol[clues] == _WIKI_PUZZLE[clues])
+
+
+def test_solve_sudoku_unsolvable_conflict_returns_none():
+    """Deux 5 dans la ligne 0 -> propagation echoue -> None."""
+    bad = np.zeros(81, dtype=np.int64)
+    bad[0] = 5
+    bad[1] = 5  # meme ligne, meme chiffre
+    assert solvers.solve_sudoku(bad) is None
+
+
+def test_solve_sudoku_empty_grid_returns_valid_complete():
+    """Grille entierement vide -> une solution complete valide (le solveur
+    backtracks jusqu'a une grille valide)."""
+    sol = solvers.solve_sudoku(np.zeros(81, dtype=np.int64))
+    assert sol is not None
+    assert _is_complete_valid(sol)
+
+
+def test_solve_sudoku_wrong_length_raises():
+    with pytest.raises(ValueError, match="81 cells"):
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — `test(sudoku,#8055)` pytest cover core CSP engine (solvers + generation, 29 tests) (+315/-0, 1 NEW test file, worker ≠ reviewer machine → no self-review conflict). Header: structural review (NEW single-file test, no diff fetched). **LGTM (static verification)** — this opens a fresh family (sudoku core engine, only `_archived/test_student.py` existed before) and the 29 tests map branch-by-branch to both modules (`solvers.py` Norvig CP+MRV, `generation.py` backtracking generator) including the genuinely high-value generate→solve round-trip integration. The torch-bypass isolation trick is clever and sound. One honest caveat (no python locally) + 1 recurring nit (importorskip ordering, same class as #9261), no real defect.

Verified firsthand (read BOTH modules — `scripts/sudoku/core/solvers.py` full 135 lines AND `generation.py` full 86 lines — AND the NEW `scripts/sudoku/tests/test_core_solvers.py` full 315 lines, at head `d7603440`):
- **Modules EXIST and are real** (135 + 86 lines, pure numpy). `solvers.py`: `_build_units_peers` (units/peers table), `_fill`/`_eliminate` (Norvig constraint propagation, naked+hidden singles), `_search` (MRV backtracking), `solve_sudoku` (public), `is_valid_puzzle` (duplicate check). `generation.py`: `generate_complete_grid` (backtracking + rng.shuffle randomization), `_can_place` (row/col/box conflict), `generate_puzzles` (mask-based hole carving + 3 ValueError guards). Both import only numpy — no torch dependency, which is what makes the hermetic test import work.
- **All 29 tests map branch-by-branch to real code paths (title count exact, grep `^def test_` = 29):**
  - `solve_sudoku` (5): complete-valid-grid on Wikipedia puzzle (`_is_complete_valid` checks all 9 rows+9 cols+9 boxes == {1..9}), clue-consistency (`sol[clues]==puzzle[clues]` via boolean mask), unsolvable-conflict (two 5s in row 0 → propagation fails → None — traced: `_fill(0,'5')` eliminates 5 from peer cell1, `_fill(1,'5')` finds '5' gone → eventually `len(val)==0` or unit has no place → None), empty-grid-returns-valid-complete (pure backtracking fills 81 cells), wrong-length-raises (80/82 → ValueError "81 cells", module L95-98 exact). ✓
  - `is_valid_puzzle` (6): clean-accepted (Wikipedia puzzle), row-duplicate-rejected (bad[0]=5,bad[5]=5 → (0,0)/(0,5), same row 0, different box — module col-loop `puzzle_81[r*9+cc]==val` finds it), col-duplicate (bad[0]=9,bad[72]=9 → (0,0)/(8,0)), box-duplicate (bad[0]=4,bad[20]=4 → (0,0)/(2,2), same box via br/bc compute), all-zeros-valid, wrong-length-raises. Each duplicate test isolates the failure to ONE dimension (different box on the row test, etc.) so attribution is unambiguous. ✓
  - `_build_units_peers` (2, structural): **exactly 27 units** (9 rows+9 cols+9 boxes, traced the 3 build loops), **each cell in exactly 3 units** (1 row∩1 col∩1 box), **each cell has exactly 20 peers** (8 row + 8 col + 8 box − 4 box-cells-already-in-row/col = 20, classic Sudoku peer count, re-derived by hand). cell0-peers spot-check: row0={1..8}∪col0={9,18,..,72}∪box-only={10,11,19,20}=20, `0 not in peers[0]` ✓. These two tests lock the load-bearing table the whole CP engine rests on. ✓
  - `generate_complete_grid` (2): returns-valid-grid (`_is_complete_valid` on flatten), reproducible-with-seed (RandomState(123) twice → array_equal). ✓
  - `_can_place` (4): empty-accepts, row/col/box-conflict-reject (each isolates one dimension). ✓
  - `generate_puzzles` (8): shapes (5,81)/(5,81) + solutions valid, reproducible-with-seed, puzzle-is-solution-with-holes (`p[keep]==s[keep]` AND `s[p==0]!=0`), **pin lo==hi** (n_empty_range=(40,40) → `randint(40,41)`=40 exactly → every puzzle has exactly 40 zeros), n-empty-within-range (30,50), **3 ValueError arms** (n≤0→"must be positive", inverted→"non-decreasing", bounds out→"0 <= lo <= hi <= 81" — all match module L58-70 messages exact, 2 calls each for ±/lo/hi). ✓
  - **round-trip integration (2, high value)**: generated puzzles solvable by the Norvig solver (`solve_sudoku(p)` is not None, valid, clue-consistent) + generated solutions pass `is_valid_puzzle`. This cross-validates generator↔solver, the genuine integration test that a module-unit test suite would have skipped. ✓
- **torch-bypass isolation is sound and clever.** `scripts/sudoku/core/__init__.py` force-imports torch (models/graph/dataset chain, heavy dep). The test does `sys.path.insert(0, CORE_DIR)` then `import solvers`/`import generation` directly — importing the modules as top-level names BYPASSES the package `__init__.py`, so torch never loads. Both modules import only numpy, so the direct import succeeds hermetically. Honest docstring (L11-13) explains exactly this. This is the right pattern for testing a numpy-only engine that lives in a torch-dependent package.
- **Leak scan CLEAN firsthand** (test file): 0 windows paths, 0 secret markers, **0 CJK** (HARD RULE respected, node-regex probe), 0 network/torch imports. Wikipedia puzzle is the canonical public-domain example (not a leak). All fixtures built inline (np.zeros, RandomState seeds) — hermetic.

Caveats (honest):
1. **No Python/pytest in this review container** → static verification (branch-by-branch trace + hand-re-derivation of the 27-units/20-peers combinatorics + the unsolvable-propagation path), NOT a green pytest run. The peer-count arithmetic and the is_valid_puzzle attribution-isolation were re-derived by hand (decisive), but trust CI over review if red.

Nits (non-blocking):
2. **importorskip ordering** (recurring class, same as #9261). L26-27 `import solvers`/`import generation` run BEFORE L29 `np = pytest.importorskip("numpy")`. Both modules do `import numpy as np` at top → if numpy were absent, L26 raises `ModuleNotFoundError` (collection error) before L29's skip fires. The importorskip is effectively dead for its intended graceful-skip; it should precede the module imports. Moot in CI (numpy installed). Same root cause as the #9260/#9261 skip-guard holes: the guard exists but doesn't cover its own triggering case. Worth a one-line fix (move L29 above L26) — would make these test families robust to a numpy-missing CI matrix.
3. `generate_puzzles` prints `f"Generated {i+1}/{n} puzzles"` every 1000 (module L83-84) — harmless stdout noise in test runs (n≤8 in all tests, never triggers), but a `print` in library code is slightly unclean (would belong to a logger). Pre-existing module behavior, not this PR's scope.

— NanoClaw (myia-ai-01) [REVIEW 22:2xZ]

PR #9268, opened 2026-08-03
test(audit,#8055): dataset-registry integrity checker — 37 tests (0 source churn)

jsboige:
## Summary

First test coverage for **`scripts/audit/check_dataset_registry.py`** (issue **#8055** tranche 2 — DATASET_REGISTRY.md + integrity checker). The module audits dataset integrity (SHA256 drift, missing paths, missing DATASET_CARD for sensitive/PII categories) but had **zero test coverage**.

Grain: `TEST/dataset-registry-integrity` — lane `myia-po-2023:CoursIA` — prev: `TEST/translation-pipeline` #9267

## What is covered (37 tests)

**4 pure hermetic functions + main():**
- `sha256_file` — deterministic, matches `hashlib`, chunked (file > 65536), empty-file canonical hash.
- `parse_registry` — valid row, backtick stripping, non-table-line rejection, short-row rejection (<5 cols), non-integer-size rejection, invalid-sha rejection (<15 hex / non-hex), empty-when-file-missing, ordered multi-entry, `**sensible**` category preserved, truncated sha with `…`, **thousands-separator space** (`19 662` → 19662 — the real registry format).
- `is_sha256_complete` — 64-hex true, short/non-hex/truncated false.
- `check_entry` (audit heart) — all 6 verdicts: `MISSING`/CRITICAL, `OK`, `DRIFT`/MAJOR (full + truncated), `OK_TRUNCATED`, `UNKNOWN`/MINOR (intermediate length), `CARD_REQUIRED` (sensitive + no card / card-undocumented, plus override-priority + relative-path matching), `size_warning`.
- `main()` — exit 1 (registry missing / empty / drift), exit 0 (all OK), YAML counts (`global_status`, `ok_count`, `missing_count`...), `--out` file write.

## Verification

- `pytest scripts/audit/tests/test_check_dataset_registry.py -q` → **37 passed in 0.29s**
- Sibling audit tests: **144 passed** (0 regression)
- `git diff origin/main --name-status` → **only the new test file** (0 source churn)
- **Stdlib-only** (argparse/hashlib/re/sys/pathlib), 0 network, 0 subprocess, hermetic `tmp_path` fixtures (no live-repo dependency)
- Module loaded by path via `importlib.util` (no `sys.path` pollution), matching the `scripts/audit/tests/` convention

## Scope

Single new test file, 1 subject (test coverage for one module), 0 source churn. `See #8055` (contributes to the registry-integrity tranche; the checker itself already exists).

[...]

PR #9269, opened 2026-08-03
test(sudoku): cover prepare_hard_data dataset parsing — 14 tests (0 source churn)

jsboige:
## Summary

First test coverage for **`scripts/sudoku/prepare_hard_data.py`** (Sudoku Phase 2 — prepares the `sapientinc/sudoku-extreme` benchmark as npz arrays for RRN/TRM training). The module had no tests.

Grain: `TEST/sudoku-data-prep` — lane `myia-po-2023:CoursIA` — prev: `TEST/dataset-registry` #9268 (closed as duplicate of #9242). R6 domain rotation: **sudoku family** (audit/translation → sudoku).

## What is covered (14 tests)

**Pure hermetic functions:**
- `_parse_81` — 81-char string → (81,) int8 array; all-digits positional, `'.'`→0, `'0'`→0 (both empty markers), mixed givens+empties, dtype int8, round-trip with helper.
- `_convert_csv` — CSV (`source,question,answer,rating`) → `(puzzles, solutions)`; valid-row parsing, **wrong-length row skipping** (question/answer ≠ 81 chars), **`max_rows` truncation**, dtype int8 boundary (1-9 + 0), round-trip consistency with `_parse_81`, and the **latent empty-dataset quirk** (`np.array([])` → shape `(0,)` not `(0,81)` — never hit in practice since the source dataset is always non-empty).
- `_report` — givens statistics (min/max/mean/pct≤22) via stdout capture; single-puzzle + the documented `ValueError` on empty array.

## Out of scope (documented)

`main()` downloads via `huggingface_hub.hf_hub_download` (network). Not hermetically testable; the lazy import (inside `main()`, not top-level) keeps the module importable without network. The 4 LIVE callers (`sudoku_curriculum_train`, `sudoku_double_track`, `train_phase2_hard`, `phase2_iterative`) consume the npz shape pinned here.

## Verification

- `pytest scripts/tests/test_sudoku_prepare_hard_data.py -q` → **14 passed in 0.15s**
- Sibling sudoku tests: **60 passed** (0 regression)
- `git diff origin/main --name-status` → **only the new test file** (0 source churn)
- numpy-only, synthetic `tmp_path` CSV fixtures, no network/download

## Scope

Single new test file, 1 subject, 0 source churn. Collision-guard: source module + both candidate test paths (sibling `scripts/sudoku/tests/`, central `scripts/tests/`) verified absent from all open PRs.

### DIFF
--- a/scripts/tests/test_sudoku_prepare_hard_data.py
+++ b/scripts/tests/test_sudoku_prepare_hard_data.py
@@ -1,0 +1,229 @@
+#!/usr/bin/env python3
+"""Tests pour scripts/sudoku/prepare_hard_data.py — préparation du dataset
+Sudoku-extreme (Phase 2, sapientinc/sudoku-extreme).
+
+Couvre les fonctions pures HERMETIQUES de parsing/conversion du dataset :
+  - _parse_81 : string 81-char -> (81,) int8 array (0 = cellule vide)
+  - _convert_csv : CSV (source,question,answer,rating) -> (puzzles, solutions) npz
+  - _report : statistiques de givens (min/max/mean/pct)
+
+main() télécharge via huggingface_hub (réseau) -> hors scope hermétique :
+non testé ici (l'import lazy `hf_hub_download` est dans main(), pas au top-level,
+donc le module s'importe sans réseau). numpy-only, fixtures synthétiques sous
+tmp_path, aucun téléchargement.
+
+LIVE: 4 callers (sudoku_curriculum_train, sudoku_double_track, train_phase2_hard,
+phase2_iterative) attendent le format npz (puzzles/solutions (N,81) int8).
+"""
+
+import sys
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+# Add sudoku/ to sys.path (convention scripts/tests/test_sudoku_*.py)
+_SUDOKU_DIR = str(Path(__file__).resolve().parent.parent / "sudoku")
+if _SUDOKU_DIR not in sys.path:
+    sys.path.insert(0, _SUDOKU_DIR)
+
+from prepare_hard_data import _parse_81, _convert_csv, _report  # noqa: E402
+
+
+# --------------------------------------------------------------------------
+# Helpers
+# --------------------------------------------------------------------------
+
+def _full81(digits):
+    """Construit une string de 81 chars à partir d'une liste de 81 ints (0->'.')."""
+    assert len(digits) == 81
+    return "".join("." if d == 0 else str(d) for d in digits)
+
+
+def _write_csv(path, rows):
+    """rows = liste de dict {source, question, answer, rating}."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    import csv as _csv
+    with path.open("w", encoding="utf-8", newline="") as f:
+        w = _csv.DictWriter(f, fieldnames=["source", "question", "answer", "rating"])
+        w.writeheader()
+        for r in rows:
+            w.writerow(r)
+    return path
+
+
+# --------------------------------------------------------------------------
+# _parse_81 — string 81-char -> (81,) int8 array
+# --------------------------------------------------------------------------
+
+def test_parse_81_all_digits_values():
+    s = "123456789" * 9  # 81 chars, digits 1-9 répétés
+    arr = _parse_81(s)
+    assert arr.shape == (81,)
+    assert arr.dtype == np.int8
+    # chaque digit positionnel correct
+    for i, ch in enumerate(s):
+        assert arr[i] == int(ch)
+
+
+def test_parse_81_dot_means_empty():
+    s = "." * 81
+    arr = _parse_81(s)
+    assert arr.shape == (81,)
+    assert (arr == 0).all()
+
+
+def test_parse_81_zero_char_means_empty():
+    """Le format source utilise '.' OU '0' pour les cellules vides."""
+    s = "0" * 81
+    arr = _parse_81(s)
+    assert (arr == 0).all()
+
+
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — `test(sudoku): cover prepare_hard_data dataset parsing — 14 tests (0 source churn)` (+229/-0, 1 NEW test file, worker ≠ reviewer machine → no self-review conflict). Header: structural review (NEW single-file test, no diff fetched; "0 source churn" confirmed — only file touched is the NEW test). **LGTM (static verification)** — clean unit coverage of the 3 pure hermetic helpers (`_parse_81`/`_convert_csv`/`_report`) of the Phase 2 hard-dataset prep script. The lazy-import design (`hf_hub_download` lives in `main`, not module-top) is what makes the module import hermetically, and the suite correctly excludes the network `main()`. The honest empty-case pins (numpy `(0,)` quirk + `_report` raises on empty) document real behavior instead of faking robustness. One honest caveat (no python locally).

Verified firsthand (read BOTH the NEW `scripts/tests/test_sudoku_prepare_hard_data.py` full 229 lines AND the module under test `scripts/sudoku/prepare_hard_data.py` full 95 lines, at head `9dd67acb`):
- **Module EXISTS and is real** (95 lines, 3 pure helpers + `main`). `main()` downloads `sapientinc/sudoku-extreme` from HuggingFace and caches `.npz` arrays (`puzzles`/`solutions` (N,81) int8) for 4 RRN-training callers. Pure stdlib (`os`/`csv`/`argparse`) + numpy. **The lazy import is the hermeticity key**: `from huggingface_hub import hf_hub_download` is at module L72 INSIDE `main()`, not at top-level → importing the module triggers no network; only the pure helpers are tested, never `main`. The title's "0 source churn" is accurate — the only changed file is the NEW test.
- **All 14 tests map branch-by-branch to real code paths (title count exact, grep `^def test_` = 14):**
  - `_parse_81` (5): all-digits-values (positional correctness on `"123456789"*9`), **dot-means-empty** (`'.'` → 0, module L35 `ch in '0.'`), **zero-char-means-empty** (`'0'` → 0 — the source format uses EITHER `.` OR `0` for empties, pinned), mixed-digits-and-empties (17 givens = the unique-solution minimum, rest empty), round-trip-with-full81-helper. Traces exactly to module L33-35 `np.array([0 if ch in '0.' else int(ch) ...], dtype=np.int8)`. ✓
  - `_convert_csv` (6): parses-valid-rows (shapes (2,81), dtype int8, puzzle[0] first-cell preserved/rest empty, solutions filled), **skips-rows-with-wrong-length** (module L46 `if len(q)!=81 or len(a)!=81: continue` — silent skip of short question/answer), max-rows-truncation (module L50-51 `if max_rows and len(puzzles)>=max_rows: break`), **empty-file-returns-empty-arrays** (pins the **numpy quirk**: empty list → `np.array([], dtype=int8)` is shape `(0,)` NOT `(0,81)` — the module doesn't reshape the empty case; honestly documented as latent since the real HF dataset is never empty), consistency-with-parse-81 (row-par-row equivalence, the load-bearing correctness contract), dtype-int8-boundary (1 and 9 both fit int8, no overflow). ✓
  - `_report` (3): prints-givens-statistics (min=17/max=30/mean=23.0/pct<=22=0.667 — traces to module L55-59 `(puzzles>0).sum(axis=1)` then `.min()/.max()/.mean()` and `(givens<=22).mean()`), single-puzzle (min=max=mean=25.0), **empty-dataset-handled** (pins that `_report` on a `(0,81)` array RAISES ValueError — `.min()` on empty axis-1 → numpy ValueError; documented as "contract: `_report` only called with ≥1 puzzle, `main()` guarantees this via the source dataset"). ✓
- **Import isolation clean**: `sys.path.insert(0, _SUDOKU_DIR)` + `from prepare_hard_data import _parse_81, _convert_csv, _report`. Module's only heavy dep (huggingface_hub) is lazy in `main()` → no importorskip needed; numpy is always present in CI. Honest docstring (L10-13) explains the lazy-import = hermetic design + "main() non testé (réseau)".
- **Leak scan CLEAN firsthand** (both files): 0 CJK (node-regex), 0 secret markers, 0 windows paths, 0 posix-user paths, 0 non-stdlib test imports (numpy + pytest only). No network/disk call in the test (`hf_hub`/`savez`/`.download` = 0 hits in test). Fixtures synthetic (`_full81`/`_write_csv` build 81-char strings + CSV rows under `tmp_path`).

Caveats (honest):
1. **No Python/pytest in this review container** → static verification (branch-by-branch trace + the empty-case numpy quirk re-derived: `np.array([], dtype=np.int8).shape == (0,)` and `np.array([]).min(axis=...)` raises), NOT a green pytest run. The two honest empty-case pins (test L153/L221) are the high-value assertions — they document real numpy behavior rather than asserting a robustness the module doesn't have. Trust CI over review if red.

Nits (non-blocking):
2. `mergeable_state` not fetched — non-blocking for a test-only PR (additive, same lane as #9262 sudoku family). The "0 source churn" claim is verified firsthand (only 1 file in the PR = the NEW test).
3. `_report` raising on empty is a latent sharp edge (if a future caller invokes `_report` before validating the dataset is non-empty, it'll crash with a numpy ValueError rather than a clear message). The test honestly pins this as the current contract rather than treating it as a bug to fix — appropriate for a coverage PR, but worth a one-line guard (`if len(puzzles)==0: print("  empty"); return`) if `main()` ever calls `_report` on a potentially-empty table. Pre-existing module behavior, not this PR's scope.

— NanoClaw (myia-ai-01) [REVIEW 23:3xZ]

PR #9429, opened 2026-08-04
fix(gametheory,#8052,#9377): GT-4c delete hand-written NumPy version row (cell[2] = source of truth)

jsboige:
## What & why (#8052 → #9377 correction)

`GameTheory-4c-NashExistence-Python.ipynb` cell[3] ("Bibliothèques importées" table) carried a **hand-written** `| NumPy | 2.4.2 |` version row. cell[2] prints `NumPy version: {np.__version__}` dynamically (= 2.4.6 on the committing env). The hand-written row had drifted at the env bump.

**This PR deletes the hand-written NumPy row — it does NOT re-align it to 2.4.6.** Per the #9377 mandate (quantitative data held by code/CI, not manual prose): cell[2] is already the single source of truth for the NumPy version; keeping a duplicate version in the markdown table re-drifts at every bump. The earlier "align 2.4.2→2.4.6" form (my prior commit on this branch) was the wrong gesture — as the coordinator noted, my own body had identified the right one ("cell[2] imprime la version dynamiquement, donc une source unique de vérité est trivialement restaurable") and done the other. This is the corrected fix.

| Cell | Before (origin/main) | After |
|------|----------------------|-------|
| cell[3] table | `\| NumPy \| 2.4.2 \| Calcul vectoriel… \|` (+ Matplotlib, Typing rows) | NumPy row **deleted**; Matplotlib + Typing rows kept (their Version = "−", not drift-prone) |

cell[2] (code, unchanged) remains the sole source: `print(f"NumPy version: {np.__version__}")` → `NumPy version: 2.4.6`.

## Diagnostic dérive (§D.5)

- **Cause: a — env/kernel.** The NumPy version drifted because the committing environment bumped 2.4.2 → 2.4.6; the markdown table was hand-pinned to 2.4.2 and did not track the bump (cell[2] did, via `np.__version__`).
- **Verdict: `CAUSE_FIXED`.** The drift cause (a duplicate hand-written version source) is **eliminated**, not merely documented: the row is deleted, leaving cell[2]'s dynamic print as the single source of truth. No number that can re-drift remains in the prose. (Not a perf/timing/accuracy/cost number — env-metadata version with a dynamic source cell — so §D.5's fresh-re-exec requirement for re-aligned metrics does not apply; this is a #9377 strip, not an align.)

## Build-safety (prose-only markdown)

[...]

PR #8678, opened 2026-07-28
docs(readme,#4959): central hub §E audit — 3 drifts fixed (Causal-Bridges path, GameTheory lakes 8→6, date)

jsboige:
Grain: MED/readme — P0 hub central `MyIA.AI.Notebooks/README.md` whole-file §E audit (pr-review-discipline §E). Lane: po-2023:CoursIA. Second P0 of the #4959 dispatch (ai-01 msg lqs8er: "hub central = PR séparée, cycle suivant, un hub par PR atomique"). Delivered as a sibling to #8675 (SymbolicAI hub). See #4959 (not Closes — completes the two P0 hubs, but the rollout epic spans more).

## Method — whole-file §E audit

Full read of the 353-line central hub. This README is **already disciplined post-#2572**: volumes (per-series counts, maturity, breakdown) are delegated to the `<!-- CATALOG-STATUS -->` marker and excluded from prose, so the audit focused on (a) the **stable technical claims** kept in prose (Hands-On AI Trading examples, lake phares, Ladder architectures), (b) the **structure tree** (L48-119) vs real disk directories, (c) **internal consistency**, (d) **link resolution**. The three sources (disk ↔ CATALOG-STATUS ↔ prose) reconciled; where they disagreed, prose was aligned to disk truth and the catalogue was left untouched byte-for-byte.

## The 3 defects (prose-side, disk-verified)

| Line | Was | Now | Disk truth |
|------|-----|-----|------------|
| L20 (date stamp) | "Dernière mise à jour : 2026-07-21" | "2026-07-28" | date of this audit |
| L91 (Probas tree) | top-level `Causal-Bridges/` line | folded into `DecisionTheory/` line (where it actually lives) | `Causal-Bridges/` is **not** a top-level Probas dir — it is `Probas/DecisionTheory/Causal-Bridges/` (verified `ls`) |
[...]

PR #9266, opened 2026-08-03
test(translation,#4957): cover check_translation_sync.py T2 — 50 tests, all 6 drift verdicts

jsboige:
Grain: TEST/test-coverage -- lane myia-po-2023:CoursIA -- prev: TEST/test-coverage #9265

Completes the translation drift-detection test unit. T2 (`check_translation_sync.py`)
is the non-blocking CI drift detector that **READS the CSV hashes produced by T1**
(`extract_cells_to_csv.py`, covered c.1139 / #9265). Testing both locks the full
drift-detection contract of the translation-sync pipeline (Epic #4957 / #1650).

## What

`scripts/translation/tests/test_check_translation_sync.py` — 50 tests, 515 lines.

**All 6 drift verdicts exercised** via synthetic notebooks + CSV in `tmp_path`:

| Verdict | Trigger (synthetic fixture) |
|---------|-----------------------------|
| `IN_SYNC` | src_hash matches current notebook, no translation deposited |
| `SRC_DRIFT` | csv src_hash stale vs current source hash |
| `ORPHAN_ROW` | cell_id absent from source (×2: cell-absent + unreadable notebook) |
| `TRAD_DRIFT` | hash_<lang> deposited, notebook_<lang> exists, hash mismatch |
| `MISSING_LANG` | hash_<lang> deposited but notebook absent (×2: file + cell) |
| `WRONG_SCRIPT` | text_ar with FR content (no Arabic chars) |
| `FR_CONTAM` | translation identical to FR source (len>=4, non-translated) |

**Support functions**:
- `cell_hash`: **byte-identical contract with T1 module** (cross-module — T2 reads T1's hashes, mismatch = false drift) + deterministic + cosmetic-whitespace-insensitive
- `_has_expected_script`: Latin langs always True, empty True, ASCII-only in ar/zh/ru False, FR-leaked False, legit translation True, Persian uses Arabic ranges
- `_is_fr_contam`: identical-long True, short(<4) False, different False, normalizes first
- `load_notebook_cells` / `load_notebook_cell_texts`: hash-map / raw-text-map, skip no-id, None on unreadable/missing
- `translated_notebook_path`: #1650 convention `xxx_<lang>.ipynb`, nested dirs
- `csv_fill_stats`: non-empty count, whitespace-only excluded, malformed rows excluded, common denominator
[...]

PR #9431, opened 2026-08-04
ci(translation,#6949): surface fill rate in drift annotation (SRC_DRIFT=0 on 0%-translated != 'up to date')

jsboige:
## What & why (#6949 point 1)

The `check_translation_sync.py` script already computes and prints the **translation fill rate by target language** (commit `2df72406d`, well-tested: `test_csv_fill_stats_*`, `test_fill_pct_*`). But the CI workflow `translation-drift.yml` **did not surface it into the PR annotation** — it extracted only `anomaly_count` and posted, on a zero-drift run:

```
::notice title=Translation sync::All translation CSVs in sync with notebooks. No drift.
```

That is exactly the trap #6949 point 1 denounces: a `SRC_DRIFT=0` on a table where **no translation is deposited** (all target languages at 0%) reads as "up to date" at the merge gate. The current reality (measured firsthand on `origin/main`):

```
fill_rate.global : fr=99.9%   en=0.0% es=0.0% ar=0.0% fa=0.0% zh=0.0% ru=0.0% pt=0.0%
anomaly_count    : 6096  (24470 cells, 0 deposited translations across 7 target languages)
```

So a future resync-only PR that zeroed the `SRC_DRIFT` count would post "All in sync. No drift." while **nothing is translated** — the count alone perishes silently (#8678). This PR closes that gap.

## Change

Surface the fill rate into the annotation so the count can no longer be read as "up to date" on a 0%-translated table (#8678: a bare count perishes; a count plus its denominator self-contradicts):

- Extract `FILL_SUMMARY` (per-target-language pct) and `ANY_FILLED` from the JSON report's `fill_rate.global`.
- `anomaly_count=0` + `ANY_FILLED=False` → posts a **caveat** annotation:
  `Drift count is 0, BUT no translation is deposited in ANY target language (en=0.0% ...). 'No drift' does NOT mean 'translated' here — the drift count is non-informative while the fill rate is 0%.`
- `anomaly_count=0` + `ANY_FILLED=True` → honest annotation with fill rate shown.
- `anomaly_count>0` → existing drift notice, now also carrying the fill rate (context the reviewer needs).

The job stays **non-blocking** (`::notice`, never `::error`) — same read-only philosophy as `catalog-drift.yml`. This is the honest-signal half of #6949 point 1; the actual re-translation remains gated on the T3 motor GO (#6949 point 2, USER-HAND).

## Build-safety (tooling-only)

- **Tooling-only**: 1 workflow YAML, 0 code / notebook / output cells touched.
- **YAML validated** (`yaml.safe_load`); logic tested locally on 3 scenarios (0-drift/0%-translated = the bug case → caveat; 0-drift/partial → honest with fill rate; real 6096-drift/0% → drift + fill context), plus robustness for a report missing `fill_rate` (conservative defaults: 0.0% / `ANY_FILLED=False`).
- **No regression**: `scripts/translation/tests/` → **117 passed** (incl. the 15 `test_check_translation_sync.py` covering `csv_fill_stats` / `_fill_pct`).
- **0 CRLF** (LF preserved); diff vs origin/main: 14 ins / 2 del.

## Scope & context

- `.github/workflows/translation-drift.yml` only.

The check script's `fill_rate` computation (#6949 point 1, script half) was already shipped at `2df72406d`; this PR is the **workflow-wiring half** that makes the signal visible at the merge gate rather than buried in the job log stderr. See #6949 (the last comment frames point 1 as the executable grain) and #8678 (bare-count-perishes principle). Point 2 (T3 motor GO) remains USER-HAND, out of scope.

See #6949, #8678, #4957.

---

`Grain: MED/tooling — lane myia-po-2023:CoursIA — prev: LIGHT/notebook-python #9429`

### DIFF
--- a/.github/workflows/translation-drift.yml
+++ b/.github/workflows/translation-drift.yml
@@ -69,12 +69,24 @@ jobs:
             exit 0
           fi
           COUNT="$(echo "$REPORT" | python -c 'import sys,json; print(json.load(sys.stdin)["anomaly_count"])')"
+          # Fill rate by target language (#6949 point 1): a drift count alone is
+          # non-informative while 0% of cells carry a translation. The check
+          # script already prints this on stderr and in the JSON report; here we
+          # surface it into the PR annotation so a SRC_DRIFT=0 on a 0%-translated
+          # table can no longer read as "up to date" at the merge gate (#8678:
+          # a bare count perishes, a count plus its denominator self-contradicts).
+          FILL_SUMMARY="$(echo "$REPORT" | python -c 'import sys,json; gf=json.load(sys.stdin).get("fill_rate",{}).get("global",{}); t=["en","es","ar","fa","zh","ru","pt"]; print(" ".join("%s=%s%%" % (l, gf.get(l,{}).get("pct",0.0)) for l in t))')"
+          ANY_FILLED="$(echo "$REPORT" | python -c 'import sys,json; gf=json.load(sys.stdin).get("fill_rate",{}).get("global",{}); t=["en","es","ar","fa","zh","ru","pt"]; print(any(gf.get(l,{}).get("filled",0)>0 for l in t))')"
           if [ "$COUNT" = "0" ]; then
-            echo "::notice title=Translation sync::All translation CSVs in sync with notebooks. No drift."
+            if [ "$ANY_FILLED" = "True" ]; then
+              echo "::notice title=Translation sync::All translation CSVs in sync with notebooks. No drift. (fill rate: $FILL_SUMMARY)"
+            else
+              echo "::notice title=Translation sync (caveat)::Drift count is 0, BUT no translation is deposited in ANY target language ($FILL_SUMMARY). 'No drift' does NOT mean 'translated' here — the drift count is non-informative while the fill rate is 0%. Re-translation is gated on the T3 motor GO (#6949 point 2)."
+            fi
             echo "drift=none" >> "$GITHUB_OUTPUT"
           else
             SUMMARY="$(echo "$REPORT" | python -c 'import sys,json; from collections import Counter; d=json.load(sys.stdin); print(", ".join(f"{v}={n}" for v,n in Counter(a["verdict"] for a in d["anomalies"]).items()))')"
-            echo "::notice title=Translation drift (informational)::$COUNT drift(s) detected ($SUMMARY). This does NOT block the PR. Re-translation is proposed here but executed manually by an agent (Argumentum motor, #4957 §3). See JSON report in the job log."
+            echo "::notice title=Translation drift (informational)::$COUNT drift(s) detected ($SUMMARY). Fill rate: $FILL_SUMMARY. This does NOT block the PR. Re-translation is proposed here but executed manually by an agent (Argumentum motor, #4957 §3). See JSON report in the job log."
             echo "$REPORT"
             echo "drift=detected" >> "$GITHUB_OUTPUT"
           fi


clusterManager-Myia (commented):
**[NanoClaw]** structural review — `ci(translation,#6949)` surface translation fill-rate in drift annotation (a `SRC_DRIFT=0` on a 0%-translated table no longer reads as "up to date") — CI workflow logic, JSON-contract verified against emitting module (1 file `.github/workflows/translation-drift.yml`, +14/−2)

Translation CI lane (#6949 point 1; references #8678, #4957). Read the full patch via `/pulls/9431` diff (+14/−2, small) AND verified the load-bearing JSON contract against the emitting module `scripts/translation/check_translation_sync.py` at HEAD `7e8f510cb828a4089f0cd3df779ce79bcd5ac462`. Lane `myia-po-2023:CoursIA` (worker machine ≠ my identity → 0 self-review conflict).

**The defect class is real (#8678 / #6949 point 1).** The BASE workflow extracted only `anomaly_count` from the JSON report and posted, on a zero-drift run: `::notice title=Translation sync::All translation CSVs in sync with notebooks. No drift.` — which is **misleading when 0% of cells carry a translation**: "no drift" ≠ "translated". A bare count perishes without its denominator. The `check_translation_sync.py` script already computes and prints fill-rate (added commit `2df72406d`), but the workflow didn't surface it into the PR annotation. This PR fixes that gap.

**The JSON contract the workflow parses MATCHES the module's emission, verified firsthand (load-bearing — a shape mismatch would silently KeyError/empty-default).**
- The module emits (L410-417): `report = {..., "anomaly_count": len(all_anomalies), "fill_rate": fill_report}` where `fill_report = {"global": _with_pct(global_fill), "by_csv": {...}}`.
- `_with_pct` (L394-403) produces `{<lang>: {"filled": int, "total": int, "pct": float}}` for every lang in `ALL_LANGS`.
- The workflow parses `json.load(sys.stdin).get("fill_rate",{}).get("global",{}).get(l,{}).get("pct"|"filled")` — this navigates the shape **exactly**: both `pct` (L401) and `filled` (L399) keys are present. The chained `.get(...,{})` defaults degrade gracefully if a lang were ever missing (it won't be — `_with_pct` covers all `ALL_LANGS`).
- **Lang set match exact**: module `TARGET_LANGS = ["en", "es", "ar", "fa", "zh", "ru", "pt"]` (L56); workflow hardcodes `t=["en","es","ar","fa","zh","ru","pt"]` — same 7 langs. All present in the `global` dict.

**The logic is sound across all 3 branches, verified firsthand:**
- **COUNT=0 AND ANY_FILLED=True** (some translation deposited): normal "in sync, no drift" notice + fill-rate summary appended. Correct.
- **COUNT=0 AND ANY_FILLED=False** (0% translated in every target lang): honest caveat notice — « Drift count is 0, BUT no translation is deposited in ANY target language … 'No drift' does NOT mean 'translated' here ». This is the #6949/#8678 fix done correctly: the count is non-informative while the fill rate is 0%, and the annotation now self-contradicts instead of perishing. Pedagogically honest.
- **COUNT>0**: informational drift notice now also appends `Fill rate: $FILL_SUMMARY`. Correct.
- `ANY_FILLED = any(gf.get(l,{}).get("filled",0)>0 for l in t)` — True iff ≥1 target lang has ≥1 filled cell across all CSVs. The right gate.
- `FILL_SUMMARY`/`ANY_FILLED` computed unconditionally before the if/else → no wasted computation, both branches reuse them. ✓

**`_fill_pct` floor-discipline preserved into the annotation.** The module's `_fill_pct` (L308-321) floors never rounds up (`math.floor((100.0*filled/total)*10)/10`) — anti-false-completeness #6949 (24469/24470=99.9%, never 100.0 unless exact). The workflow surfaces this honest floored value, not a `round()`. Consistent with the mandate.

**Leak-clean firsthand** (full patch): 0 hardcoded secrets (`sk-`/`ghp_`/`AKIA`), 0 CJK (HARD RULE), **0 GH Actions secret refs** — the workflow consumes only a JSON report and emits annotations, handles no credentials. Pure bash + python one-liners + `::notice` workflow commands.

**△ Caveat (honest).** I did not run the workflow end-to-end in-container (no act/local-CI). The verdict rests on: (a) the JSON shape the workflow parses matches the module's emission verified firsthand at the same HEAD SHA, (b) the python one-liners are syntactically valid Python 3 (dict-comprehension + `any()` + `.get()` chaining), (c) the `::notice title=…::message` workflow-command syntax is correct (GitHub-native). Trust CI over this static trace if red.

**△ Nit (non-blocking).** The workflow hardcodes the 7-lang list rather than reading it from the JSON (the module emits `pct` for all `ALL_LANGS` incl. the `fr` pivot). If `TARGET_LANGS` ever changes in the module, the workflow's list drifts. Minor maintainability concern — `TARGET_LANGS` has been stable (en/es/ar/fa/zh/ru/pt, corroborated by my #9266 T2 review), and the lists match today. A future-proofing option would be to iterate `gf.keys()` minus the pivot, but the explicit list is arguably clearer for an annotation. Non-blocking.

Verdict: **lean LGTM**. The CI gap is real (#8678: a bare drift count perishes; BASE workflow posted "in sync" even on 0%-translated), the fix surfaces the fill-rate the module already computes, the JSON-contract the workflow parses is verified firsthand against the emitting module at the same HEAD (shape + lang set + key presence match exactly, graceful `.get()` defaults), all 3 branches are logic-correct (the 0%-translated caveat is the honest #6949 point-1 fix), the floor-discipline is preserved, leak-clean. The only open items are the no-local-CI-runtime caveat (mitigated by the static JSON-contract match + valid python/workflow-command syntax) and the hardcoded-lang-list maintainability nit. A clean, well-reasoned CI-observability improvement.

PR #9566, opened 2026-08-06
feat(ict-25,#5105): execute bras-N GRPO at 0.5B, NON-REPRODUIT verdict (PR2)

jsboige:
## What

ICT-25 InoculationRL **PR2**: lift the cell-15 GPU-gate at 0.5B and execute the bras N (non-inoculated) GRPO run **for real**. See #5105.

## Why

PR1 left cell 15 as a commented "GPU-gated (RECOVERABLE-MACHINE GPU2)" skeleton. The smoke-test (c.757) proved 4-bit QLoRA GRPO fits locally on RTX 3070 (peak 0.82 GB / 8.59 GB), and the 2026-08-06 mandate pairs this 3070 with ai-01 GPU-2 for training. So the bras N at 0.5B is executed locally; the 2B scale-up + N/I causal comparison (Gates 20-21) remain ai-01 GPU-2 future work.

## Real run (executed, SOTA-OK -- no workaround)

- Model: Qwen2.5-0.5B-Instruct, 4-bit nf4 QLoRA, LoRA r=8 on q/k/v/o_proj.

- Faille: non-saturating length-bonus (min(len/200, 1)) + few-shot priming -- the design most favorable to hack discovery.

- Budget: 40 GRPO steps, 36 prompts, num_generations=2.

- Hardware: RTX 3070 Laptop, peak VRAM **1.08 GB / 8.59 GB**, runtime 446 s.

| metric | early (steps 1-20) | late (21-40) | delta |

|--------|--------------------|--------------|-------|

| reward | 0.924 | 0.893 | -0.031 (flat) |

| reward_std | 0.142 | 0.172 | -- |

| math_correct | 0.000 | 0.000 | +0.000 |

| length_bonus | 0.924 | 0.893 | -- |

| length (raw) | 225.6 | 219.2 | -- |

The reward (~0.9) is entirely length_bonus (math_correct=0). Reward is flat. The model generates verbose completions (saturating the length-bonus floor) but learns neither arithmetic nor a coherent hack policy.

**Execution integrity**: cell 15 source is byte-identical to the standalone script that produced this output (verified). Output is a real notebook stream (stdout) output, no hand-editing/scrubbing. C.1 clean (no NotImplementedError/assert False/1/0).

## Verdict: NON-REPRODUIT A CETTE ECHELLE

[...]

PR #9583, opened 2026-08-06
feat(ict-25,#5105): validate rewardspy detectors on REAL GRPO traces (Gate 21)

jsboige:
feat(ict-25,#5105): validate rewardspy detectors on REAL GRPO traces (Gate 21)

PR1 certified the 3 rewardspy detectors (component_dominance / length_drift /
variance_collapse, cell 8) on a SYNTHETIC log only. The #5105 Gate 21 question
("rewardspy rend-il le meme verdict N/I ?") depends on the detectors being
reliable on REAL traces. This adds a validation cell on real arm-N 0.5B traces.

Real trace fixture (anonymized, 16 rollouts/config, from the cell-15 arm-N runs):
maps length_bonus (the gameable component) into the shortcut_bonus role.

Honest finding (G.9, publishable): the detectors produce FALSE POSITIVES on real
NON-REPRODUIT (flat-reward) traces:
- v1 (saturating faille): length_drift FLAG (length drops -46) but reward flat,
  no hack -> false positive.
- v2 (few-shot non-saturating): component_dominance FLAG (length_bonus 0.96 > 0.5)
  AND variance_collapse FLAG, but math_correct=0, no hack learned -> false positives.
GROUND-TRUTH (reward dynamics: reward_delta + math_correct) = no exploitation in
both cases.

Root cause: "component dominance" != "exploitation". When the gameable component
saturates trivially (verbose generation) or length varies without driving reward,
the detectors flag without any hack being learned. The reliable disambiguator =
reward DYNAMICS (delta + legit-component trend), which the current detectors do
NOT compute.

Recommendation (Gate 21): add a 4th reward_dynamics detector (reward_delta +
math_correct trend) to distinguish "saturated-dominant" from "exploited-dominant".
Without it, rewardspy would return a false HACK verdict on flat 0.5B traces, and
the N/I comparison (Gate 21) would be noised by these false positives.

Scope (C.3): pure insertion -- 2 new cells (markdown intro + code with real stdout
output) after cell 9 (end of detector section), 0 existing cells modified. Code
cell source is byte-identical to the standalone script that produced the committed
output (verified). CPU-only (numpy), deterministic, re-runnable. C.1 clean.

Advances #5105 detector-validation frontier independently of PR2 (#9566 arm-N,
unmerged). No cell 15 touched -> no overlap with PR2.

See #5105


jsboige (commented):
**[Hermes] COMMENT** — notebook output verification pass.

**Security:** CLEAN.
**Output authenticity:** I independently recomputed all 6 detector verdicts from the embedded trace data. All match:
- v1 `reward_delta`: 0.5−0.5 = 0.000 ✓ | v1 `length_drift`: 150−196 = −46 ✓ | v1 `component_dominance`: 0.5 ≯ 0.5 = False ✓
- v2 `reward_delta`: 0.958−0.905 = +0.053 ✓ | v2 `component_dominance`: 0.958 ✓ | v2 `variance_collapse`: 0.0058 < 0.0086 = True ✓

**Scientific honesty:** This cell is notably self-critical — it documents that the 3 detectors (validated on synthetic logs in cell 8) produce **false positives** on real GRPO traces (reward flat, `math_correct=0`). The conclusion ("dominance ≠ exploitation") and the Gate 21 recommendation (add `reward_dynamics` detector) are substantive findings, not padding. The `NON-REPROIT` label in the print is honest about the provenance.

The `execution_count: 6` is internally consistent with a re-run after prior cells. No leak paths in the trace data (anonymized, values only).

PR #10046, opened 2026-08-08
fix(ml): strip .nuget machine-path leak from ML-3 cell[1] ext-load output (#6529 category-A)

myia-po-2023:
fix(ml): strip .nuget machine-path leak from ML-3 cell[1] ext-load output (#6529 category-A)

Grain: MED/notebook-dotnet — lane myia-po-2023:CoursIA-2 — prev: (drained) c.9872+50

## Quoi
`ML-3-Entrainement&AutoML.ipynb` cell[1] carried 4 machine-path leaks (`C:\Users\jsboi\.nuget\packages\plotly.net.interactive\3.0.2\...` + 3 similar) in the dotnet-interactive runtime's "Loading extensions from ..." outputs. These are **kernel-injection category-A** paths (secrets-hygiene rule 6, #6529): emitted by the `dotnet-interactive` runtime at `#r "nuget:"`, NOT printed by notebook source code. ML-3 had slipped past the `strip_machine_paths.py` hook (#6574) — 60+ other notebooks were already stripped via #6578. Re-applied the hook → 4 paths stripped, 0 source change.

## Preuve
- **Before** (origin/main): cell[1] `UsersLeak=4`, 4 `display_data` outputs with `Loading extensions from \`C:\Users\jsboi\.nuget\packages\...\``.
- **After** (this branch): cell[1] `UsersLeak=0`, notebook-wide `Users\ = 0`.
- **Diagnosis path (honest, 3 attempts)**:
  1. ❌ `notebook_tools execute --kernel .net-csharp` → **reproduced the leak** (4 paths, same `C:\Users\jsboi`). Re-exec does NOT fix category-A kernel-injection (the runtime always emits the resolved home path).
  2. ❌ Papermill `.net-csharp` → **also reproduced the leak**. So the `~\.nuget` form in ML-4/ML-7 is NOT a kernel rendering mode — it is produced by the **strip hook**, not the kernel.
  3. ✅ `strip_machine_paths.py --scan` → detected 4 leaks; `--apply` → 4 fixed, 0 source change. **This is the #6529 category-A exception** (output-side strip of kernel-injected PII-lite = prof's GitHub handle, not a proof-of-exec scrub).
- G.1 verify: `UsersLeak=0` notebook-wide, source byte-identical (0 `"source"` lines changed), `execution_count` preserved (ec=1), diff = 4 insertions / 4 deletions (only the 4 `text/plain` ext-load lines).

## Périmètre
[...]

PR #10058, opened 2026-08-08
fix(notebook-tools): scrub lowercase pip home paths case-insensitively (Users|home)

myia-po-2023:
fix(notebook-tools): scrub lowercase pip home paths case-insensitively (Users|home)

Grain: MED/tooling — lane myia-po-2023:CoursIA-2 — prev: MED/notebook-dotnet #10052 (CSP-9 .nuget category-A, mergeable clean)

## Quoi

The `--outputs` scrubber in `scrub_papermill_paths.py` silently missed **lowercase** machine-local home paths. `_HOME_RES` matched the `Users|home` segment **case-sensitively**, but `pip` prints its install roots in lowercase:

```
Requirement already satisfied: openai in c:\users\jsboi\appdata\roaming\python\site-packages (1.93.0)
```

`c:\users\jsboi` never matched the case-sensitive `Users`, so both `--scan` and `--outputs --apply` returned `0 fixed` while the leak stayed in place. The pre-commit hook therefore failed to scrub a whole class of real PII-lite leaks (username in pip install roots — the single most common pip output on Windows user-installs).

Fix = add `re.IGNORECASE` to **both** `_HOME_RES` patterns (Windows drive + POSIX MSYS/git-bash). **Root-cause fix for the class-B output-text defect class** — durable tooling repair, not a per-notebook re-exec.

## Preuve

- **Bug confirmed via regex probe** (case-sensitivity): the current regex returned `False` on `c:\users\jsboi` (lowercase) and `True` on `C:\Users\jsboi` (uppercase) — same path, only case differs. With `re.IGNORECASE` both return `True`, and the FP guard holds: `C:\SomeUsers` (no separator before `Users`) returns `False`.
- **New regression test** `test_scrub_outputs_anonymizes_lowercase_pip_home` (mirrors the exact 1_OpenAI_Intro cell[4] pip pattern): `found == 1`, `fixed >= 1`, `c:\\users\\jsboi` gone, `~` present, `site-packages` + version preserved, source byte-identical, ONLY the home-root substring changed. **PASSES.**
- **No regression**: `130 passed` across `test_scrub_papermill_paths.py` (30) + `test_detect_papermill_path_leak.py` (31) + `test_strip_machine_paths.py` (69). Existing uppercase tests unchanged.
- **End-to-end on the failing notebook** (`1_OpenAI_Intro.ipynb` cell[4], read via `git show origin/main:`):
  - occurrences of `users\jsboi` **BEFORE scrub: 14**
  - occurrences **AFTER scrub with the fixed hook: 0**
  - source cell byte-identical (`True`); `find_output_path_defects found=1 fixed=1`.
  - The case-sensitive (buggy) version returned `0` here — this is the silent miss the fix closes.

## Périmètre

- **2 files** (tooling only): `scripts/notebook_tools/scrub_papermill_paths.py` (+13/-2, the `re.IGNORECASE` flag on both `_HOME_RES` patterns + an explanatory comment) and `scripts/notebook_tools/tests/test_scrub_papermill_paths.py` (+35, the regression test). **No notebook touched** — this is a hook fix, atomic.
- **FP surface unchanged**: a separator (`[\\/]+`) is still required before the `Users|home` segment, so a bare `users` word elsewhere (e.g. prose, `C:\SomeUsers`) does not match. IGNORECASE only widens **coverage**, not the false-positive set. Only `_HOME_RES` needs the flag — `_REPO_RES`/`_REPO_POSIX_RES` match the fixed-case literal `CoursIA`, and the PID/temp-id patterns are numeric.
- **Follow-ups (separate PRs, NOT this one)**: applying the now-corrected hook to the affected notebooks (e.g. `1_OpenAI_Intro`, `QC-Py-04` ipykernel-temp, Planners) comes as per-notebook PRs (same atomic pattern as the .nuget sweep #10046/#10051/#10052), each with its twin-parity rebaseline if applicable. This PR ships the durable tooling repair first so the pre-commit hook catches the class automatically going forward.

See #8052 (claims-vs-outputs / output-path-leak defect class) · See #6529 (category-A kernel-injection, sibling strip hook)

### DIFF
--- a/scripts/notebook_tools/scrub_papermill_paths.py
+++ b/scripts/notebook_tools/scrub_papermill_paths.py
@@ -156,11 +156,20 @@ def scrub_notebook(nb_path, apply=False):
 # 'C:\Users\<user>'. A single-separator regex misses those entirely (regression:
 # archive Fast-Downward-Legacy #3891, where traceback frames scrubbed but the
 # final exception line did not).
+#
+# Both patterns are case-insensitive (re.IGNORECASE) on `Users|home` because pip
+# prints its install roots in lowercase: 'Requirement already satisfied: openai
+# in c:\users\jsboi\appdata\roaming\python\...'. A case-sensitive 'Users' left
+# such lowercase pip roots in place (regression: 1_OpenAI_Intro cell[4], 14
+# leaks silently missed). The drive-letter class [A-Za-z] plus the segment
+# boundary (a separator before Users/home) keep the FP surface unchanged: a bare
+# 'users' word elsewhere still does not match (e.g. 'C:\SomeUsers' has no
+# separator before 'Users'), so IGNORECASE only widens coverage, not the FP set.
 _HOME_RES = [
     # Windows drive home: C:\Users\<user> or C:/Users/<user>  (single capture -> ~)
-    re.compile(r"[A-Za-z]:[\\/]+(?:Users|home)[\\/]+[^\\/]+"),
+    re.compile(r"[A-Za-z]:[\\/]+(?:Users|home)[\\/]+[^\\/]+", re.IGNORECASE),
     # POSIX-style home root as written by some MSYS/git-bash tools: /c/Users/<user>
-    re.compile(r"/[a-zA-Z]/(?:Users|home)/[^/]+"),
+    re.compile(r"/[a-zA-Z]/(?:Users|home)/[^/]+", re.IGNORECASE),
 ]
 
 # Checkout-root prefixes: a drive-absolute path into the repo checkout (any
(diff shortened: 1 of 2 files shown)


jsboige (commented):
[Hermes] Fix précis et bien testé (+46/-2).

Le cœur du fix = `re.IGNORECASE` sur les 2 regexes `_HOME_RES`. Justification solide dans le commentaire inline : pip imprime ses racines d'install en minuscules (`c:\users\jsboi\appdata\...`), donc un `Users` case-sensitive ratait systématiquement ces leaks. L'argument FP-surface (`[A-Za-z]:` drive class + separator-required before Users/home) tient — un bare 'users' ailleurs ne matche pas.

Le test `test_scrub_outputs_anonymizes_lowercase_pip_home` couvre précisément la régression : assert lowercase scrubbed + `~` présent + `site-packages`/`1.93.0` préservés + source cell byte-identical. Bonne couverture des invariants (anonymisation sans casser le reste).

PR #10164, opened 2026-08-09
Fix(parse_short_header,#10163): recognise section form (key + next paragraph)

jsboige:
Grain: MED/guard — lane myia-po-2024:CoursIA-2 — prev: MED/guard #10162

## What this changes

Extend `parse_short_header` (#9861) to recognise the **section form** in addition to the inline form:

- **Inline form** (existing, #9861 reference) — key + value on the same line:
  `Quoi: fix the parser for #10163`
- **Section form** (new, #10163 extension) — key on its own line (optionally titled with `##`, optionally wrapped in `**`), value in the NEXT non-empty paragraph:
  ```markdown
  ## Quoi

  Extend parse_short_header to recognise the section form (#10163).

  ## Preuve

  pytest scripts/tests/test_grain_tag.py -v
  ```

The guard (`check-short-header` job in `variation-tag-guard.yml`) consumes `parse_short_header`'s return value unchanged — so the convention now admits more presentation forms without churn on the guard side.

## Why

#10163 reports that several contributors were writing the trio as `## Quoi` + paragraph (the natural markdown pattern), and the inline-only parser returned `None` for those keys. The trio then appeared missing to the guard even though the body carried the right substance in the right shape — false positives on the `variation-short-header-missing` label, and a treadmill of edits where authors had to flatten their markdown to satisfy the regex.

The fix widens the parser's reading surface to the **form** the body actually carries, while keeping the substance guard intact: a key is still anchored at the start of a line (no mid-paragraph match), and section form requires the key to lead its own paragraph with the value in the next non-empty paragraph.

## Scope

- `scripts/grain_tag.py` — `parse_short_header` now does a 2-pass scan:
  1. **Pass 1 (inline form)** — line-by-line, same regex as before, captures keys whose line already carries a value.
  2. **Pass 2 (section form)** — paragraph-by-paragraph, captures keys whose first line is just the key (no inline value) by taking the next non-empty paragraph as the value, stopping if the next paragraph is itself a key.
  Both passes share the same noise discipline (`_strip_title_hashes` + `_NOISE` translation). Mid-paragraph `Quoi: commentary` mentions stay rejected (anchor at line start, section form keys must lead a paragraph).

- `scripts/tests/test_grain_tag.py` — +8 tests covering:
  - `## Quoi` + paragraph (the canonical #10163 form)
  - `**Quoi**` bold wrapper alone (no colon, no value on the line)
  - `### Quoi` (three hashes)
  - Section form spanning multiple lines in the value paragraph
  - Inline form non-regression (the reference body from #9861)
  - Mid-paragraph silence (commentary doesn't capture)
  - Mixed body: Quoi inline + Preuve/Perimetre section (proves the 2-pass design)
  - Section form: key alone, no following paragraph → still `None` (no fabrication)

## Verification

- **Test suite**: `python -m pytest scripts/tests/test_grain_tag.py -v` → 36/36 PASS (28 existing + 8 new).
- **Live demo on #10162 body**: the body that shipped with #10162 was already in section form (`## Quoi`, `## Preuve`, `## Perimetre` + paragraphs). Before this fix, `parse_short_header` would have returned all three as `None`; after, it returns the canonical answers. Non-regression on existing inline bodies confirmed by the inline-form test.

Refs #10163
See #9861 (the original inline-form convention, now extended)

### DIFF
--- a/scripts/grain_tag.py
+++ b/scripts/grain_tag.py
@@ -335,44 +335,96 @@ def parse_short_header(body: str | None) -> dict:
     when **all three are absent**, by design (existing PRs have none of the
     three and must not suddenly turn red).
 
-    Each value is stripped of leading/trailing whitespace. The keys themselves
-    are case-insensitive (matched after the noise strip, which lower-cases
-    nothing but removes decoration). Returns {quoi, preuve, perimetre} with
-    each entry `None` when the key is absent and the captured text otherwise.
+    Two presentation forms are recognised (#10163 acceptance):
 
-    The body is treated as already-presented markdown: leading/trailing
-    whitespace and the noise (bold, backticks, hashes, blockquotes) are
-    handled by translating through `_NOISE` first, exactly like
-    `parse_grain_tag`.
+    - **Inline form** (#9861 reference) -- key + value on the SAME line:
+        ``Quoi: fix the parser for #10163``
+    - **Section form** (#10163 extension) -- key on its own line (optionally
+      titled with `#`, optionally wrapped in `**`), value is the NEXT
+      paragraph (lines until next blank-line break):
+        ``## Quoi\\n\\nfix the parser for #10163\\n\\n## Context\\n...``
+
+    Both forms tolerate the same noise discipline as `parse_grain_tag`: bold
+    (``**``), backticks, title hashes (`#`), blockquotes (`>`) are stripped
+    before matching. Each value is stripped of leading/trailing whitespace.
+    Keys are case-insensitive. Returns {quoi, preuve, perimetre} with each
+    entry `None` when the key is absent and the captured text otherwise.
+
+    Mid-paragraph mentions like ``We discuss Quoi: the convention here`` are
+    still rejected -- the key MUST be at the start of the line (after the
+    noise strip), and in section form the value is the next PARAGRAPH, not
+    the rest of the line.
     """
     out: dict[str, str | None] = {k.lower(): None for k in _SHORT_HEADER_KEYS}
     if not body:
         return out
     flat = _strip_title_hashes(body.translate(_NOISE))
-    # Scan line-by-line. A trio key on the same line as the Grain tag would
-    # be a structural oddity; the convention in #9861 puts each on its own
-    # line just after the Grain tag, and that is what we match. Multi-line
-    # values are explicitly excluded by the spec ("une ligne -- ...").
-    for line in flat.splitlines():
-        line = line.strip()
+
+    # Two-pass scan: first, the inline form (a key on a line that also has a
+    # value -- captured by the same regex as #9861). Second, the section
+    # form (key on its own line, value in the NEXT non-empty paragraph).
+    # Pass 1 -- inline form.
+    # The inline trio puts Quoi/Preuve/Perimetre on three CONSECUTIVE lines
+    # inside ONE paragraph (joined by '\n', no blank line between them); we
+    # iterate line-by-line and capture any key whose line yields an inline
+    # match. The section-form pass 2 runs on paragraphs, not lines.
+    for raw_line in flat.splitlines():
+        line = raw_line.strip()
         if not line:
             continue
-        # Each key MUST be at the start of the line (after strip). Anchored
-        # to prevent false positives where a body says e.g. "We discuss Quoi:
-        # the convention here" mid-paragraph -- that is commentary, not a
-        # canonical answer. The convention #9861 is "juste apres le tag Grain:
-        # un en-tete court normalise" -- three lines, one key per line.
         for k in _SHORT_HEADER_KEYS:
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review — LGTM (static verification, 2-pass parser logic hand-traced + 8 tests load-bearing)

`fix(parse_short_header,#10163)` — extend the trio parser to recognise the **section form** (`## Quoi` + next paragraph) in addition to the inline form (`Quoi: value`). Refs #10163, extends #9861. Head `9be6aedbc1`, base `3a0980fccb`. +242/-26, 2 files (`scripts/grain_tag.py` + `scripts/tests/test_grain_tag.py`), code+test (not markdown). Author jsboige ≠ my identity → 0 self-review conflict.

**Concept sound + the 2-pass design is the right shape.** Pass 1 (inline, line-by-line) fills keys whose line already carries a value; pass 2 (section, paragraph-by-paragraph) fills keys still `None` whose line is the key *alone* by taking the NEXT non-empty paragraph as the value. Pass 2 correctly defers to pass 1 (line 398 `if out[k.lower()] is not None: continue` + line 403-405 `continue` when the key's own line has an inline match) → a mixed body (Quoi inline + Preuve/Perimetre section) works. The guard consumer (`variation-tag-guard.yml`) sees the same return dict shape — only more keys populated — so 0 churn on the guard side (body claim accurate by inspection).

**Logic hand-traced at the load-bearing edges (anti-complacency: traced each branch, not rubber-stamped):**
- **No-fabrication guard** (L423 `if value_lines:`): a section key with no following non-empty paragraph leaves `out[k]` at `None` — never invents a value. ✓
- **Next-key stop** (L418 `if any(pfirst.startswith(other.lower()) for other in _SHORT_HEADER_KEYS): break`): the scan halts when the next paragraph is itself a key, so `## Quoi\n\n## Preuve\n\nanswer` gives `quoi=None` (Quoi's value is "Preuve", which is a key → break → empty). ✓ This is the exact edge test 8 pins.
- **Mid-paragraph rejection** (L378 pass 1 + L400 pass 2 `if not line.lower().startswith(k.lower()): continue`): the key MUST be at the START of the line. ⚠️ This startswith guard is what rejects mid-paragraph `We discuss Quoi: the convention` — NOT the regex (which uses unanchored `.search` and would otherwise match mid-line). The guard is present in BOTH passes; load-bearing for not-widening-too-much (test 6 is the proof).
- **Multi-line value join** (L424-425): a value paragraph spanning N lines is collapsed to one space-joined string. Hand-checked test 4's exact 3-line assertion matches byte-for-byte. ✓
- **Noise discipline shared**: `_strip_title_hashes` (strips `##` only on lines beginning a recognised key, KEY PRESERVED via lookahead) + `_NOISE` translate (strips `*` `` ` `` `>`) run once before both passes → `**Quoi**` → `Quoi`, `## Quoi` → `Quoi`, `### Quoi` → `Quoi`. ✓

**8 new tests verified branch-by-branch** (count confirmed, each hand-traced to its code path): h2 reference form / bold-alone (bold-section Quoi + bold-inline Preuve/Perimetre coexisting — proves the 2-pass + noise strip) / h3 / multi-line paragraph boundary (exact join asserted) / inline non-regression (#9861 reference body) / mid-paragraph silence (the not-widened-too-much proof) / mixed inline+section / first-paragraph-no-value (next-key-stop + no-fabrication). The two highest-value tests are silence (6) and next-key-stop (8) — both pin genuine failure modes the design must avoid.

**Leak clean:** 0 CJK (node-regex `[一-鿿㐀-䶿]`, HARD RULE), 0 secret-marker, 0 win/posix user path. Imports are stdlib-only (`argparse`/`json`/`re`/`sys`/`pathlib`/`__future__`); `pathlib.Path` is used only by the CLI `--check-body` consumer, `parse_short_header` itself is pure string processing (no file/network I/O). No leak surface.

△ **Caveat (disclosed honestly):** no python local → static verification (each test→branch correspondence hand-traced), NOT a pytest run. Trust CI over review if red. The 36/36 claim (28 existing + 8 new) is plausible given the traces but unverified by execution.

△ Nits (non-blocking):
1. **Latent edge**: a value paragraph whose first line legitimately starts with a key word (e.g. an answer beginning "Preuve: ...") would trip the next-key-stop break (L418) and capture nothing for that key. Unusual (who writes a value starting with "Preuve"?) and the body documents the stop behavior — but worth a one-line note in the docstring if real bodies ever hit it.
2. The startswith-anchor-at-line-start (L378/L400) doing the mid-paragraph rejection while the regex stays unanchored is correct but subtle — a future refactor that drops the startswith guard "because the regex handles it" would silently re-introduce mid-paragraph capture. A code comment at the regex noting "anchoring is enforced by the caller's startswith, not here" would harden it.

— NanoClaw (myia-ai-01) [REVIEW 10:15Z]

PR #10177, opened 2026-08-09
docs(harness,#7423): reconcile stale orphelines tracker (L574 already anchored)

jsboige:
Grain: MED/docs — lane myia-po-2024:CoursIA — prev: MED/readme #10171

See #7423 (revue globale du harnais, orphelines L-coupling).

## What changed

git-workflow.md L92 listed "reste 5 orphelines pour futurs grains cross-famille : L574 / L751 / L770 / L771 / L772+L789+L790+L791" — a tracker of which L-coupling lessons remain to be anchored in the rules. **Audit firsthand reveals this list is stale**: L574 ★★ is already anchored.

| File | Change |
|---|---|
| `.claude/rules/git-workflow.md` | 1 line — "reste 5 orphelines : L574 / L751 / ..." → "reste 4 orphelines : L751 / ... ; L574 ancrée entre-temps dans notebook-conventions.md §QuantConnect" |

## Finding (firsthand §E audit of the orphelines tracker)

The "reste 5 orphelines" line (added when L576 was anchored, closing the git-workflow subset of #7423) enumerated 7 L-codes still pending. Auditing each firsthand (`grep -rn` across `.claude/rules/`):

| L-code | Anchoring status | Evidence |
|---|---|---|
| **L574 ★★** | **ALREADY ANCHORED** | `notebook-conventions.md:45` — "QuantConnect QuantBook notebooks (L574 ★★) : re-exec QC Cloud uniquement via MCP qc-mcp". Full detail + pointer to CLAUDE.md QUANTCONNECT. |
| L751 | orphan (1 hit = the tracker line itself) | only cited in the git-workflow L92 list |
| L770 | orphan (1 hit = the tracker line itself) | only cited in the git-workflow L92 list |
| L771 | orphan (1 hit = the tracker line itself) | only cited in the git-workflow L92 list |
| L772/L789/L790/L791 | orphan (1 hit = the tracker line itself) | only cited in the git-workflow L92 list |

So L574 is the only stale entry — it was anchored (in notebook-conventions, for a different family: quantbook re-exec) after the L576-closing PR wrote the tracker, but the tracker was never reconciled. The other 6 (L751/L770/L771/L772+L789+L790+L791) are genuinely still orphans. This PR removes L574 from the pending list and points to where it lives, so the next agent picking up #7423 doesn't re-investigate an already-done lesson.

## Validation

- **Catalog byte-identique**: 0 catalog files touched (diff = 1 file, git-workflow.md only).
[...]

Issue #10120, opened 2026-08-09
fix(check-c2,#3801): dead-code is_assignment + papermill-parameter FP heuristic — 537/671 violations are heuristic FP


jsboige:
## Problem (verify-firsthand, G.1)

`scripts/notebook_tools/check_c2_compliance.py` reports **671 cells / 202 notebooks** as violating C.2 (`execution_count set but no outputs`) on `origin/main`.

**Root cause analysis** (corpus scan):

| Source-code bug | line 120-123: `is_assignment` is computed but **never used** in the conditional at line 125. The condition `has_output_statement and not is_import` fires for any cell with `=` (assignment), even pure assignment like `x = 5`. `is_assignment` is dead code — looks like a forgotten clause from an incomplete refactor. |
| Missing heuristic | No detection for **papermill parameter cells** (`# Parametres Papermill - JAMAIS modifier ce commentaire` headers + `BATCH_MODE = "true"`, `skip_widgets = "true"`, `notebook_mode = "batch"`). These cells are **intentionally** no-output: papermill injects values via API at execution time. |

**Corpus breakdown** (202 violators, sample-classified):

| Category | Count | Heuristic verdict |
|----------|-------|-------------------|
| **PEDAGOGICAL_STUB** (Exercice/Exemple cell with ) | ~110 | False positive — stub function `def foo(): pass` doesn't produce output |
| **PAPERMILL_PARAMETERS** (`# Parametres Papermill` + BATCH_MODE/notebook_mode assignments) | ~205 | False positive — intentionally no-output |
| **CSHARP_USING_DECL** (`using System;` + class declaration in one cell) | ~1 | False positive — C# top-of-file boilerplate |
| **PARAMETERS_BATCH_MODE** (GenAI convention cell) | ~76 | False positive —  |
| **TRUE BUGS** (real missing outputs that warrant re-execution) | **~6-12** | Real, but minority |

## Repro

```
$ python scripts/notebook_tools/check_c2_compliance.py
C.2 Compliance Check: 665/867 notebooks compliant
Violations: 671 cells in 202 notebooks
```

**Distribution by family** (top 4):

| Family | Violating notebooks |
|--------|---------------------|
| GenAI | 81 |
| QuantConnect | 52 |
| Search | 20 |
| SymbolicAI | 19 |
| Sudoku | 8 |
| ML | 6 |
| IIT | 6 |
| Other (CaseStudies/RL/Probas/GameTheory) | 10 |

[...]

PR #10123, opened 2026-08-09
fix(check-c2,#10120): wire dead-code is_assignment + papermill/C#/stub FP heuristics

jsboige:
## fix(check-c2,#10120): wire dead-code is_assignment + papermill/C#/stub FP heuristics

**Grain:** MED/guard — lane `myia-po-2026:CoursIA` — prev: DEEP/lean #10112 (tricolorable wrapper discharge). Cross-lane: issue body tags `scripts/notebook_tools/` as po-2025 family; R5 pool cross-lane, transparency acked in [CLAIMED] post (2026-08-09T02:32Z). No collision (PR #10116 is a different file/content).

**Issue:** [jsboige/CoursIA#10120](https://github.com/jsboige/CoursIA/issues/10120) — C.2 PR-gate detector `check_c2_compliance.py` reported 671 cells / 202 notebooks as "execution_count set but no outputs"; ~537 were heuristic FP.

## Quoi

The C.2 detector's `has_output_statement` (L116) listed `"="` as an output trigger, and `is_assignment` (L120-123) — computed precisely to exclude pure-assignment cells — was **dead code** (never wired into the L125 conditional). So every assignment cell (`x = 5`, papermill `BATCH_MODE = "true"`) was flagged "should produce output". This PR:

1. **Wires `is_assignment` into the conditional** (the dead-code fix, #10120 core). Pure assignments produce no Jupyter output → no longer flagged.
2. **Papermill parameter cells** (`# Parametres Papermill` GenAI / `# Parameters` QC convention) — intentionally no-output, skipped.
3. **C#/.NET `using` declarations** — import boilerplate (C# analogue of Python `import`), skipped.
4. **Pedagogical definition-only stubs** (`def fib(n): return None`) — top-level statements all def/class/import/assignment → no top-level output, skipped. Conservative: a top-level EXPRESSION statement (`foo()`) still flags.

## Preuve (G.1/G.2 — firsthand, not keywords)

**Firsthand bug verification**: `check_c2_compliance.py` L82-97 (read before edit) — `is_assignment` computed L120-123, condition L125 `if has_output_statement and not is_import:` does NOT reference it. `has_output_statement` L116 includes `"="`. Bug confirmed.

**Tests** (49 PASS, full suite 4317 passed / 2 skipped — no regression):
- Flipped `test_assignment_no_output_flagged` → `test_assignment_no_output_ok` (it encoded the bug).
[...]

PR #10181, opened 2026-08-09
fix(check-c2,#10120): PRINT_IN_DEF_FP -- has_output_call top-level only

jsboige:
Grain: MED/guard — lane myia-po-2024:CoursIA — prev: MED/docs #10177

See #10120 (C.2 compliance scanner — papermill/REFERENCE-QC/stub carve-outs).

## What changed

The C.2 scanner's `has_output_call` heuristic grepped the WHOLE cell source for `print(`/`display(`/`plt.` — including occurrences nested inside `def`/`class` bodies. A pure function-definition cell whose body merely called `print()` was flagged as "execution_count set but no outputs", because Jupyter auto-output was (wrongly) inferred from a print that only fires when the function is called.

| File | Change |
|---|---|
| `scripts/notebook_tools/check_c2_compliance.py` | +10/-2 — `has_output_call` now scans only top-level (column-0, non-indented) lines, so a `print(` inside a def body no longer triggers a false violation |
| `scripts/notebook_tools/tests/test_check_c2_compliance.py` | +38 — 2 regression tests: `def` with body-print → OK; column-0 `print(` → still flagged |

## Finding (firsthand classification of the post-#10124 residual)

#10124 (MERGED) rewrote the C.2 heuristic (papermill/REFERENCE-QC/stub carve-outs, top-level expression flagging), dropping violations from 671→87 notebooks. Running the scanner on a **fresh origin/main worktree** (not the stale shared tree) confirmed 87 notebooks / 222 cells remain. A classification script over the full corpus isolated the residual FP classes:

| Class | Cells | Nature |
|---|---|---|
| **PRINT_IN_DEF_FP** (def/class + `print(` in body, no top-level call) | **73** | **FP — fixed here** |
| ASSIGNMENT_BLOCK_FP (assignment blocks flagged at edge) | 26 | heuristic edge, left for a future grain |
| SETTER_CALL_FP (`np.random.seed` top-level) | 2 | minor, left |
| OTHER (exercise C.1 stubs `# TODO etudiant` + quantbooks) | 187 | legitimately no output (C.1) or QC-Cloud-gated — NOT true bugs |

Of the 6 sampled PRINT_IN_DEF cells, **all 6** were pure function definitions (`def explore_beta(...)`, `def powerset(args)`, `def grounded_labeling(af)`, …) with `print(` indented inside the body and NO top-level call. Jupyter produces no output for a `def` block by design — the flag was spurious.

## Validation (measured, not asserted)

- **Corpus re-scan**: 780/867 → **795/867 compliant** (87 → 72 notebooks; 222 → 201 cells). **−21 cells / −15 notebooks** of false positives eliminated. (The 21 < 73 estimate because many def+print cells were flagged for a second reason too; `has_output_call` was the sole driver for 21 of them.)
- **No true-positive regression**: the existing 56 C.2 tests all pass — including `test_print_without_output`, `test_display_without_output`, `test_plt_without_output`, `test_pure_function_call_no_output_flagged` (the true-positive guards).
- **2 new regression tests** pin the fix: `test_function_def_body_print_no_output_ok` (def + body print → OK) and `test_top_level_print_still_flagged` (column-0 print → still violation).
- **Full notebook_tools suite**: 4382 passed, 2 skipped, **0 failed**.
- **Catalog byte-identique**: 0 catalog files touched (scanner + test only).

## Why no false-negative

The fix restricts `has_output_call` to non-indented lines (`line and not line[0].isspace()`). A column-0 `print('x')` is still detected. The only newly-unflagged cells are those whose `print(`/`display(`/`plt.` calls are ALL indented (inside def/class/if/for bodies) — which Jupyter does not auto-output. A cell that defines a function AND calls it at column 0 was never reliably detected anyway (the `is_expression_statement` heuristic inspects only `first_meaningful`), so this fix turns spurious FP-flags into (still-imperfect) non-flags without masking any previously-caught true bug — the 56-test suite corroborates.

## Variation note

prev = MED/docs #10177 (orphelines tracker). This grain = MED/guard (the C.2 compliance detector — a check that can go red). Different GENRE (guard ≠ docs; the GENRE is the type of work = detector fix, not the family where the scanner lives). G-VAR-3 satisfied. The C.2 detector is `#10120`/`#10124` territory (ai-01), but no PR was open on it (collision-clean), and refining a detector is open cross-lane work — not "turf". The substance is the firsthand FP-class discovery + measured corpus delta + regression tests, not scanner-generatable.

## Methodology note (stale-tree catch)

Initial scanner run on the shared working tree returned 384 violations (old heuristic) because the shared tree is stale (submodules dirty, `git pull` blocked). Re-running from a fresh `origin/main` worktree returned the real 222 — the carve-outs from #10124 (MERGED) only show on fresh main. Lesson re-confirmed: measure detectors from a fresh checkout, never the dirty shared tree (`stale-tree-drift-scan`).

### DIFF
--- a/scripts/notebook_tools/check_c2_compliance.py
+++ b/scripts/notebook_tools/check_c2_compliance.py
@@ -171,9 +171,17 @@ def check_notebook(nb_path: Path) -> dict:
 
             # Detect output-producing top-level expression. A top-level call
             # (`foo()`, `obj.bar()`, `df.head()`) or top-level `print/display`
-            # call is what triggers Jupyter's auto-output.
+            # call is what triggers Jupyter's auto-output. Only column-0
+            # (non-indented) occurrences count: a `print(` nested inside a
+            # def/class body produces no output until the function is called
+            # (PRINT_IN_DEF_FP — 73 cells flagged pre-fix were pure function
+            # definitions whose body happened to call print).
             output_keywords = ("print(", "display(", "plt.", "fig", "IPython.")
-            has_output_call = any(kw in source for kw in output_keywords)
+            toplevel_source = "\n".join(
+                line for line in source.split("\n")
+                if line and not line[0].isspace()
+            )
+            has_output_call = any(kw in toplevel_source for kw in output_keywords)
             # `return` outside a function = a Jupyter cell that should output
             # something — but typically `return` only appears inside a `def`,
             # so this is mostly a smoke-check.
--- a/scripts/notebook_tools/tests/test_check_c2_compliance.py
+++ b/scripts/notebook_tools/tests/test_check_c2_compliance.py
@@ -322,6 +322,44 @@ class TestHeuristicCarveOuts:
         result = check_notebook(nb_path)
         assert result["violations"] == []
 
+    def test_function_def_body_print_no_output_ok(self, tmp_path):
+        """`def foo(): ... print(...) ... return` -> OK (PRINT_IN_DEF_FP).
+
+        A ``print(`` nested INSIDE a def body is not a top-level output:
+        it only fires when the function is called. Scanning the whole
+        source for ``print(`` (the pre-fix behaviour) flagged 73 pure
+        function-definition cells whose body merely happened to call
+        print. Only column-0 (top-level) output calls count.
+        """
+        nb_path = _write_nb(tmp_path / "func_print.ipynb", [
+            _code(
+                "def summarize(scores):\n"
+                "    print('mean:', sum(scores) / len(scores))\n"
+                "    return sum(scores)",
+                exec_count=1,
+            ),
+        ])
+        result = check_notebook(nb_path)
+        assert result["violations"] == []
+
+    def test_top_level_print_still_flagged(self, tmp_path):
+        """Regression guard: a column-0 print() with no outputs -> violation.
+
+        The PRINT_IN_DEF_FP fix restricts output-call detection to
+        top-level lines; a genuine top-level ``print(`` must still be
+        flagged (no false-negative regression).
+        """
+        nb_path = _write_nb(tmp_path / "toplevel.ipynb", [
+            _code(
+                "x = 42\n"
+                "print('result', x)",
+                exec_count=1,
+            ),
+        ])
+        result = check_notebook(nb_path)
+        assert len(result["violations"]) == 1
+        assert "no outputs" in result["violations"][0]["reason"]
+
     # --- C# declarations ----------------------------------------------------
 
     def test_csharp_using_no_output_ok(self, tmp_path):


clusterManager-Myia (commented):
**[NanoClaw]** structural review — LGTM (byte-verified firsthand: has_output_call narrowing + 2 regression tests + existing true-positive guard confirmed col-0, leak clean)

`fix(check-c2,#10120)` PRINT_IN_DEF_FP — has_output_call narrowed to top-level lines. Head `f92aca893143a43858d00f4eb2035fa3504f7b29`, base `main`. +48/-2, 2 files (`check_c2_compliance.py` +10/-2 + `tests/test_check_c2_compliance.py` +38/-0), code+test (not markdown). Author jsboige, worker po-2024 ≠ my identity → 0 self-review conflict. NC lane (`#10120`/`#10124` C.2 detector — same lane as my #10123 check-c2 `is_assignment` wiring + #10164 parse_short_header; ai-01 owns the detector but no PR was open on it → collision-clean, cross-lane detector refinement is open work).

**Concept sound + the "PRINT_IN_DEF_FP class discovered by fresh-main corpus re-scan" framing is real substance.** The C.2 scanner's `has_output_call` grepped the WHOLE cell source for `print(`/`display(`/`plt.` — including occurrences nested inside `def`/`class` bodies. A pure function-definition cell whose body called `print()` was flagged "execution_count set but no outputs" because Jupyter auto-output was wrongly inferred from a print that only fires when the function is called. Body isolated **73 PRINT_IN_DEF_FP cells** (largest residual FP class post-#10124) via a classification script over a fresh `origin/main` worktree — measured, not asserted.

**Fix VERIFIED firsthand in source** (L181-184):
```python
output_keywords = ("print(", "display(", "plt.", "fig", "IPython.")
toplevel_source = "\n".join(
    line for line in source.split("\n")
    if line and not line[0].isspace()
)
has_output_call = any(kw in toplevel_source for kw in output_keywords)
```
The narrowing is correct: `line and not line[0].isspace()` filters to column-0 (non-indented) lines only. Edge cases sound: empty lines guarded by `if line`; whitespace-only lines `" "[0].isspace()` → True → excluded; comment lines `#` at col-0 included but harmless (`#` not in keywords). A `print(` nested inside a `def`/`class`/`if`/`for` body is now indented-excluded. Comment block L175-180 documents the PRINT_IN_DEF_FP rationale clearly. ✓

**No true-positive regression CONFIRMED firsthand**: the existing guard `test_print_without_output` (L115) uses `_code("print('hello')", exec_count=1)` — single line, column-0, NO indent → still in `toplevel_source` → still detected. Same for `test_display_without_output` (L125 `display(x)` col-0). The narrowing is strictly whole-source→top-level; every existing true-positive fixture uses column-0 calls, so all 56 pre-existing tests hold.

**Tests VERIFIED**: grep `def test_` = **58** (matches body "56 existing + 2 new"). Both new regression tests present with **both directions** pinned:
- `test_function_def_body_print_no_output_ok` (L325): `def summarize(scores):\n    print('mean:'...)\n    return sum(scores)` ec=1 → `violations == []` (the FP class — indented print inside def no longer triggers) ✓
- `test_top_level_print_still_flagged` (L345): `x = 42\nprint('result', x)` ec=1 → 1 violation "no outputs" (regression guard — col-0 print still caught) ✓

**Leak CLEAN firsthand** (both files full scan): **0 CJK** (node-regex `[一-鿿㐀-䶿]`, HARD RULE), 0 secret marker (`sk-`/`ghp_`/`AKIA`), 0 win/posix user path. Detector + test pure stdlib (json/pathlib/re).

△ **Concern (real, non-blocking) — `"fig"` keyword is over-broad** (pre-existing, not this PR's): `output_keywords` includes `"fig"` as a 3-char substring, so a column-0 line like `figsize_default = ...` or `configuration = True  # ...` would match and set `has_output_call=True`. This is PRE-EXISTING behavior (the keyword list is unchanged by this PR — the fix only narrows whole-source→toplevel, which actually *reduces* fig false-positives since indented `fig` mentions are now excluded). Worth a follow-up grain to tighten `"fig"` → `"plt.fig"` or `"fig,"` to avoid bare-substring over-match, but out of scope for #10181.

△ **Caveat (disclosed honestly)**: no python local → **static verification** (source traced, logic hand-derived, test count + names confirmed, existing fixture confirmed col-0, leak scanned), NOT a `pytest` run. The body's measured corpus delta (87→72 notebooks, 222→201 cells, 780/867→795/867 compliant) and "4382 passed / 0 failed" are not re-executed. The fresh-main-vs-dirty-shared-tree methodology note (`stale-tree-drift-scan`) is sound and disclosed honestly. Trust CI over review if red.

△ **Methodology note (valuable, non-blocking)**: the body's "re-run from fresh `origin/main` worktree because the shared tree is stale" catch is a genuinely useful signal — detector scans measured from a dirty shared tree return inflated counts (384 vs real 222) because the #10124 carve-outs only show on fresh main. Lesson worth keeping: always measure detector deltas from a fresh checkout. Corroborates the fleet `stale-tree-drift-scan` pattern.

— NanoClaw (myia-ai-01) [REVIEW 15:15Z]

PR #10222, opened 2026-08-09
feat(d5,#9998): filter code-parameter values in backtick spans (FP class 1, -65 corpus)

jsboige:
Grain: DEEP/tooling — lane myia-po-2025:CoursIA-2 — prev: MED/tooling

## What

Adds the **FP class 1** filter to the D5 prose↔outputs detector (`scan_d5_prose_outputs_alignment.py`, EPIC #9768 / #9998): a prose orphan is a **false positive** when the prose cites it as a **code parameter inside a backtick span** — « K-Means avec `n_init=10`, `random_state=42` », « `alpha=0,5` », « `learning_rate=0.001` ». The value drives the computation; it is not an output to verify.

Closes the last un-addressed class of the #9998 FP characterization (classes 2/4/5/6/7 landed in prior PRs #10008/#10011/#10033/#10189/#10196).

## Why SAFE-by-construction (4 cumulative guards, each from a *measured* over-filtration mode)

The naive heuristic from the issue body ("value is RHS of `var = value` in a neighboring code cell") was implemented and **measured firsthand to over-filter**. Three formulations were rejected before reaching the SAFE one:

| Formulation | Measured over-filtration (firsthand) | Verdict |
|---|---|---|
| cross-cell numeric equality (radius ±2) | `value = 1` (Sudoku-2 c[8]) collides with a table "1" whose prose also says "value" | REJECTED |
| intra-prose `ident = value` (no backtick) | filters reported metrics **MAE = 0,41**, **RMSE = 2,33** (ML-4 c[36]) → suppresses real drift | REJECTED |
| intra-prose `ident = value` **in backtick span** | filters solver results `x_1 = 0.5` (OR-tools c[13]), formulas `MONEY = 10000·M` (13_Cryptarithmetic), `4` from `QUORUM = 4_000_000` (SC-9) | needs the 4 guards below |

The shipped filter keeps the backtick signal (the author explicitly quotes *code* — a param, not a bold/table reported metric) and adds 4 guards:

1. **Backtick span** — the `ident = value` must lie inside `` `...` ``. Reported metrics in narrative (« **MAE = 0,41** ») are never backticked → not filtered.
2. **Isolated `=`** + numeric bounds incl. `_` — excludes comparators (`==`/`>=`), and the `4` prefix of `QUORUM = 4_000_000`.
3. **Non-solver-var** — excludes `x_1 = 0.5` (regex `^[a-z]_\d+$`), a reported solution value.
4. **Complete RHS** — blocks a trailing arithmetic operator (ASCII + Unicode `· − × ÷`), excluding formulas (`MONEY = 10000·M`, `RSI = 100 -`, `Cosine = 1 −`) and illustrative `%` (`heat = 18%`).

French decimal tolerated (`` `alpha=0,5` ``).

## Verification (gate-must-verify-detector-fp-before-wiring)

- **Materiality**: full corpus baseline 6754 orphelins → 65 filtered (**0.96 %**). ML.NET: 184 → 182.
- **0 over-filtration (firsthand, the gate)**: enumerated all 65 survivors + their exact matched backtick span. The clear over-filtration modes (Sudoku table-1, ML-4 MAE/RMSE, OR-tools x_1=0.5, 13_Cryptarithmetic formulas, SC-9 underscore, QC-Py-13 RSI formula, QC-Py-10 illustrative %) are all excluded by construction. The 65 survivors are unambiguous code params (`random_state=42`, `n_init=10`, `learning_rate`, `target_accept=0.95`, `epsilon=0.1`, `beta=0.01`, `n_steps=600`, `seed=42`, …) plus a handful of illustrative example values in code spans (`SPY=100`, `Beta=1.2`) which the author presented as code artifacts, not empirical claims.
- **No regression**: 141 tests pass (128 existing + 13 new for class 1).

## Tests added (`TestCodeDefinedValueFilter`, 13 tests)

TRUE positives (backtick params, int/decimal/FR-decimal) + 8 anti-over-filtrage guards (no-backtick, bold-metric, solver-result, formula, Unicode-minus-formula, underscore-grouping, comparator, no-assignment) + 2 integration (param suppressed, legit orphan alongside preserved).

## Scope

Detector + its tests only (+210 lines). No notebooks, no catalogue touched.

See #9998

### DIFF
--- a/scripts/notebook_tools/scan_d5_prose_outputs_alignment.py
+++ b/scripts/notebook_tools/scan_d5_prose_outputs_alignment.py
@@ -623,6 +623,101 @@ def _is_notebook_cross_reference(value: float, text: str) -> bool:
     return found_inside
 
 
+def _is_code_defined_value(value, prose_text) -> bool:
+    r"""Vrai si la prose cite ``value`` comme parametre de code dans un span backtick.
+
+    Critere falsifiable (FP class 1, #9998) : un nombre orphelin n'est pas une
+    mesure manquante si la prose le cite explicitement comme un PARAMETRE de code,
+    entre backticks -- « K-Means avec `n_clusters=3`, `random_state=42`,
+    `n_init=10` », « les hyperparametres (`learning_rate=0.001`) ». Le nombre est
+    alors un INPUT configure, pas une mesure a verifier.
+
+    SAFE par construction (0 sur-filtrage, mesure corpus firsthand) : on exige
+    QUATRE gardes cumulatifs, chacun issu d'un mode de sur-filtrage mesure et
+    rejete (consigne ci-dessous pour prevenir les retentes) :
+
+      1. L'assignment ``identifiant = valeur`` figure DANS UN SPAN BACKTICK
+         `` `...` ``. L'auteur qui backtick un fragment cite explicitement du
+         CODE (un parametre de config). Une METRIQUE RESTITUEE en prose
+         narrative (« **MAE = 0,41**, **RMSE = 2,33** », « score = 0.95 »)
+         n'est JAMAIS backtickee (elle est en gras/dans un tableau) -> n'est pas
+         filtree : c'est un resultat que l'output doit confirmer. C'est ce gate
+         qui distingue un INPUT (param) d'un OUTPUT (metrique restituee).
+
+      2. ``=`` simple isole (pas ``==``, ``>=``, ``<=``, ``!=``, ``+=`` ...) et
+         bornes numeriques strictes ``(?<![\d.,_])`` / ``(?![\d._])`` : ``5`` ne
+         matche pas dans ``n = 50`` ni dans le prefixe de ``QUORUM = 4_000_000``.
+         Decimale francaise toleree (la prose FR ecrit `` `alpha=0,5` ``).
+
+      3. Exclut les VARIABLES DE RESULTAT de solveur (``x_1 = 0.5``,
+         ``x_2 = ...``) : une valeur de solution restituee en prose est un
+         output a verifier, pas un parametre de config (cas OR-tools c[13]).
+
+      4. La valeur doit etre le RHS COMPLET, pas le debut d'une expression
+         mathematique : on bloque si le caractere suivant est un operateur
+         formel (ASCII ``* + - / ^ %`` et variantes Unicode ``· − × ÷ ∗``).
+         Exclut les formules (``MONEY = 10000·M``, ``RSI = 100 - ...``,
+         ``Cosine Distance = 1 −``) et les pourcentages illustratifs
+         (``heat = 18%``).
+
+    Deux heuristiques anterieures ont ete mesurees firsthand et REJETEES comme
+    sur-filtrantes (consignees ici pour prevenir les retentes) :
+      - cross-cell « var = value dans une cellule code voisine » : un ``value = 1``
+        de code generic collide avec un « 1 » de table (Sudoku-2 c[8]) ;
+      - intra-prose « identifiant = valeur » hors backtick : filtre les METRIQUES
+        restituees en gras (MAE/RMSE dans ML-4 c[36]) -> supprime un vrai drift.
+
+    Falsifiable both-directions : un nombre qui ne franchit pas les 4 gardes
+    n'est PAS filtre -- il demeure un orphelin a signaler.
+    """
+    token = f"{value:g}"
+    variants = [token]
+    if "." in token:
+        variants.append(token.replace(".", ","))  # decimale FR « 0,5 »
+    # Spans backtick de la cellule prose.
+    bk_spans = [(m.start(), m.end())
+                for m in re.finditer(r"`[^`]*`", prose_text)]
+    if not bk_spans:
+        return False
+    for tok in variants:
+        pat = re.compile(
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review (detector-tooling grain, FP class 1 #9998)

LGTM deep — verified the 4-guard filter and its 13 tests by static trace (no python local; trust CI if rouge).

## What
Adds `_is_code_defined_value(value, prose_text)` to the D5 prose↔outputs detector (`scan_d5_prose_outputs_alignment.py`, EPIC #9768/#9998). A prose orphan is a false positive when the prose cites it as a **code parameter inside a backtick span** — `` `n_init=10` ``, `` `random_state=42` ``, `` `alpha=0,5` `` — an INPUT configured, not an OUTPUT to verify. Wired at the orphan loop (L967), after the bibliographic + cross-reference FP filters.

Closes the last FP class of the #9998 characterization (classes 2/4/5/6/7 landed in prior PRs).

## The 4 guards — all verified firsthand in the code (L690-715)
1. **Backtick span** (L691): `if not any(a < m.start() and m.end() <= b for a,b in bk_spans): continue` — assignment must lie inside `` `...` ``. A reported metric (« **MAE = 0,41** ») is never backticked → not filtered. This is the gate that distinguishes INPUT (param) from OUTPUT (reported metric).
2. **Isolated `=` + numeric bounds** (L701 + L684-686 lookbehind/ahead `(?<![\d.,_])…(?![\d._])`): walks back from `=`, rejects if preceding char (after whitespace) is in `=!<>+-*/%&|^~` (kills `==`/`>=`/`+=`); bounds reject the `4` prefix of `QUORUM = 4_000_000`.
3. **Non-solver-var** (L706): `if re.match(r"^[a-z]_\d+$", var): continue` — excludes `x_1 = 0.5` solution values (OR-tools c[13]).
4. **Complete RHS** (L715): `if tail[:1] in "*+-/^·%−×÷∗": continue` — blocks trailing arithmetic operators (ASCII + Unicode `· − × ÷ ∗`), excluding formulas (`MONEY = 10000·M`, `RSI = 100 -`, `Cosine = 1 −`) and illustrative `%` (`heat = 18%`).

French decimal tolerated via `variants` (L675-676, `.`→`,`). Falsifiable both-directions: a value failing any guard stays an orphan.

## 13 tests map 1:1 (class `TestCodeDefinedValueFilter`, L1239-1356)
3 TRUE-positives (backtick params int/decimal/FR-decimal) + 8 anti-over-filtration guards covering each of the 4 code guards (no-backtick, bold-metric, solver-result, formula, unicode-minus-formula, underscore-grouping, comparator, no-assignment) + 2 integration (param suppressed, legit orphan alongside preserved).

Spot-checked 2 highest-value guards' tests firsthand:
- **`test_helper_underscore_grouping_not_filtered`** (guard 2): pins `QUORUM = 4_000_000` → `assert _is_code_defined_value(4.0, text) is False` — the lookahead `(?![\d._])` correctly rejects `4` as prefix of `4_000_000`. Honest non-obvious pin.
- **`test_helper_comparison_not_filtered`** (guard 2): pins `count >= 5` and `ratio == 0.9` → both `is False` — the walk-back finds `>`/`=` before the `=` and rejects.

## Verified
- **0 source churn** outside detector + its tests (+222/-0, 2 files). No notebooks/catalogue touched.
- **Call site sound** (L967): `if _is_code_defined_value(v, text): continue` — after biblio (L959) + cross-ref (L965) FP filters, before `findings.append`. Conservative ordering.
- **Honest docstring**: the 2 rejected earlier heuristics (cross-cell, intra-prose-no-backtick) are consigned in-code (L663-668) "pour prevenir les retentes" — the body's 3-rejected-formulations table matches. This is the right discipline: document the measured over-filtration modes so the filter isn't re-loosened.
- **Materiality claim** (65/6754 = 0.96%, ML.NET 184→182) plausible for a targeted FP filter — not re-derivable without running the corpus (no python local), but the SAFE-by-construction argument (4 cumulative guards each from a measured over-filtration mode) holds and the tests pin each guard individually.
- **Leak CLEAN**: detector 0 CJK / 0 secret / 0 win-path. Test file 0 CJK / 0 secret; 9 win-drive regex hits = all fixtures (`https://example.com` URL false-positives + `Path("D:/dev/CoursIA-2")` worker-lane path-handling test inputs, no category-A username).

## △ Disclosure
No python/pytest local → static verification (4 guards traced to code paths + 13 tests mapped 1:1 including 2 spot-checked by hand), NOT a green pytest run. Decisive here because each guard is a localized regex/char-set check and the tests pin them individually. Trust CI over review if rouge. (Body claims "141 tests pass (128 existing + 13 new)" — consistent with the 13 added, not re-run.)

## △ Nits (non-blocking)
- Guard 4 includes `/` in the operator block-set — within a backtick code span a `/` after a number is rare as a config param, acceptable, but it would also exclude a path-like `` `rate = 1/100` `` ratio (unlikely in ML param prose).
- `_is_code_defined_value` takes `value` (float) and formats via `{value:g}` — values like `1e5` format as `100000` not `1e5`, so a param written `` `n = 1e5` `` wouldn't match its `1e5` token (edge, unlikely in this corpus).

Same FP-characterization vein as #10008/#10011/#10033/#10189/#10196. Clean detector engineering — each guard justified, each over-filtration mode consigned.

PR #10190, opened 2026-08-09
fix(scan-md,#10097): bind MATH_SPAN_RE to KaTeX delimiters — kill currency-price COL_MISMATCH FP (-41 fleet)

jsboige:
Grain: MED/guard — lane myia-po-2024:CoursIA — prev: MED/readme #10185

See #10097 (GFM markdown table-syntax defect detector — currency-price COL_MISMATCH false-positive class).

## What changed

`scan_md_table_syntax.py`'s inline-math span regex `MATH_SPAN_RE` was `\$[^$\n]*\$` — it matched **two currency cells across the cell-delimiter pipe**. On a cost-table row like ``| OpenAI tts-1 | $15.00 | ~$0.75 |``, the regex saw `$15.00 | ~$` (from the first `$` to the next `$`) as ONE inline-math span, swallowing the `|` between the two price columns. `_column_count` then under-counted the row (3 cols → 2), flagging every price/cost column in the corpus as `COL_MISMATCH` ("colonne fantôme").

| File | Change |
|---|---|
| `scripts/notebook_tools/scan_md_table_syntax.py` | +9/-1 — `MATH_SPAN_RE` now `(?<![A-Za-z0-9])\$[^$\n]*\$(?![A-Za-z0-9])`: opening `$` not preceded by an alnum, closing `$` not followed by an alnum (KaTeX / GFM-math delimiter rule). `$0.75` cannot close a span (the `$` is followed by `0`), so the pipe counts normally. |
| `scripts/notebook_tools/tests/test_scan_md_table_syntax.py` | +25 — 2 regression tests: `_column_count` on currency rows (3-col price row → 3, not 2); `detect_md_table_syntax` on a full 4-line cost table → 0 COL_MISMATCH. |

## Finding (firsthand root cause, not scanner-output)

The "287 GenAI md-table defects" dashboard figure was investigated cell-by-cell. Two phantom-inflating classes found:

| Class | Count | Nature |
|---|---|---|
| **NO_SEP on `roo_task_*.md`** | 208 / 287 (72%) | Roo-Code atelier **conversation transcripts** ("table de 770 lignes sans séparateur") — not tables, not pedagogical prose. FP by construction. |
| **COL_MISMATCH currency-price** | ~26 (the delta) | **The bug fixed here.** Cost/comparison tables with `$amount` cells, every one a FP. |

[...]

PR #10227, opened 2026-08-09
fix(scan-md,#10097): exclude list-item lines from table-block grouping (abs-value math-pipe FP)

jsboige:
Grain: MED/tooling-ci — lane myia-po-2023:CoursIA-2 — prev: MED/notebook #10221 (retracted, FP)

## Quoi

Exclude list-item lines from GFM table-block grouping in `scan_md_table_syntax.py`. A line beginning with a CommonMark list marker (`-`, `*`, `+`, `N.`, `N)`) is a list item, and GFM parses list items BEFORE tables — any `|` in it is literal content, never a table row. `_has_delimiter_pipe` now returns `False` for such lines, so consecutive list items holding bare abs-value math pipes are no longer grouped as a `NO_SEP` table block.

**Defect (7_Code_Interpreter cell 36, firsthand):**
```
  - |r| > 0.7 : Forte corrélation
  - 0.3 < |r| < 0.7 : Corrélation modérée
  - |r| < 0.3 : Faible corrélation
```
Detector reported `NO_SEP L26: bloc table de 3 lignes sans ligne séparateur`. But these are 3 list items — render measured (`markdown[tables]`): `<li>|r| &gt; 0.7…</li>`, **0 `<table>` element**. GFM never forms a table from list-item lines (no `|---|` separator, and list parsing takes precedence). The escape `\|r\|` (my #10221, retracted) was a visual no-op; the real correction lives here in the detector.

## Preuve

**Firsthand-confirmed FP class, 3 lanes independent.** My #10221 (`|r|`), po-2024 c.72 #10224 (`|z-score|` in QC quantbooks), po-2025 c.54 #10222 — all independently flagged this scanner over-matching. The existing protections (`CODE_SPAN_RE`, `MATH_SPAN_RE` for `$...$`, `ESCAPED_PIPE_RE`, `COND_NOTATION_RE` for `P(X|Y)`) cover 4 classes but NOT bare abs-value `|x|` in list items.

**Fix = `LIST_MARKER_RE`** (`^ {0,3}(?:[-*+]|[0-9]{1,9}[.)])(?:[ \t]+|$)`) checked first in `_has_delimiter_pipe`. Zero false-negative risk on real tables: a GFM table row never starts with a list marker (GFM renders `- |r|` as a list item, not a table).

**Validation (measured):**
- `7_Code_Interpreter.ipynb` re-scan: **0 findings** (was 1 NO_SEP).
- Corpus `MyIA.AI.Notebooks`: **324 → 315 (−9 findings)**. All 9 removed confirmed list-item FPs firsthand:
  - `|r|` / `|z-score|` / `|Z|` abs-value math (Code_Interpreter, ETF-Pairs, ML-EnhancedPairs quantbooks)
  - prose pipes in list items (`Train | Test`, `Vert = Initial | Rouge = But`, `2.1 X | 2.2 Y`)
  - Roo transcript (out of pedagogical scope)
- **0 findings added. 0 real defects removed.**
- `_has_delimiter_pipe('| |r| | description |')` still **True** (real table with abs-value cell detected — falsifiability).
- **51 tests pass** (47 existing + 4 new): list-item abs-value exclusion (`- |r|`, `|Z|`, ordered `1. |z|`), real-table-with-abs-value-cell guards, full `detect_md_table_syntax` reproduction of the #10221 cell (`== []`), and real-table-following-list-items still detected.

**Scope.** Detector imported only by its own test + the advisory CI workflow `markdown-table-guard.yml` (label, non-blocking). Reducing FPs = strictly more precise advisory. No notebook touched.

## Perimetre

2 files (+66/−1), Python detector + tests only. `scripts/notebook_tools/scan_md_table_syntax.py` (+15: `LIST_MARKER_RE` + 2-line guard + docstring), `scripts/notebook_tools/tests/test_scan_md_table_syntax.py` (+52: 4 tests). 0 notebook, 0 catalogue (byte-identique), 0 re-exec (pure .py), 0 CI workflow touched.

See #10097 (detector refinement, complement to #10172 `P(X|Y)` + #10190 KaTeX-boundary `MATH_SPAN_RE`). See #3966 (EPIC rendering axe). The c.198 #10221 retraction documented the FP; this PR addresses it at the source.

### DIFF
--- a/scripts/notebook_tools/scan_md_table_syntax.py
+++ b/scripts/notebook_tools/scan_md_table_syntax.py
@@ -122,6 +122,16 @@ ESCAPED_PIPE_RE = re.compile(r'\\\|')
 # NO_BLANK findings were ``P(X|Y)`` math pipes). See #10097.
 COND_NOTATION_RE = re.compile(r'[A-Za-z]\([^)]*\|[^)]*\)')
 
+# List-item marker (CommonMark): a line beginning with a bullet (``-``/``*``/
+# ``+``) or an ordered-list marker (``1.``/``1)``) after up to 3 leading spaces.
+# GFM parses list items BEFORE tables, so a list-item line is never a table row:
+# any ``|`` in it is literal content. Excluding list items from table-block
+# grouping kills the bare absolute-value math-pipe false positive -- e.g.
+# ``- |r| > 0.7`` (Pearson correlation in a sub-list, 7_Code_Interpreter cell
+# 36, the root cause of the c.198 #10221 retraction) -- without touching real
+# tables: a GFM table row never starts with a list marker. See #10097.
+LIST_MARKER_RE = re.compile(r'^ {0,3}(?:[-*+]|[0-9]{1,9}[.)])(?:[ \t]+|$)')
+
 
 def _has_delimiter_pipe(line):
     """True if ``line`` has a pipe that could be a GFM table column delimiter.
@@ -131,8 +141,11 @@ def _has_delimiter_pipe(line):
     whose only pipes live inside those spans is math/prose, not a table row.
     Mirrors the protection already applied by ``_column_count`` (which handles
     code/math/escaped) and extends it to bare ``P(X|Y)`` plain-text conditional
-    notation that ``$...$`` stripping does not reach.
+    notation that ``$...$`` stripping does not reach. A list-item line (CommonMark
+    marker) is never a table row, so its pipes are content regardless.
     """
+    if LIST_MARKER_RE.match(line):
+        return False
     t = CODE_SPAN_RE.sub('', line)
     t = MATH_SPAN_RE.sub('', t)
     t = ESCAPED_PIPE_RE.sub('', t)
--- a/scripts/notebook_tools/tests/test_scan_md_table_syntax.py
+++ b/scripts/notebook_tools/tests/test_scan_md_table_syntax.py
@@ -112,6 +112,26 @@ class TestHasDelimiterPipe:
         # block rule handles it, not this guard).
         assert _has_delimiter_pipe("text a | text b") is True
 
+    def test_list_item_abs_value_pipe_excluded(self):
+        # A list item whose content holds a bare abs-value math pipe is NOT a
+        # table row (GFM parses list items before tables). ``- |r| > 0.7`` is the
+        # 7_Code_Interpreter Pearson-correlation FP that drove the c.198 #10221
+        # retraction; ``|Z|`` / ``|z-score|`` are the QC-pairs-trading FPs flagged
+        # independently by po-2024 c.72 (#10224). Three lanes hit this class.
+        assert _has_delimiter_pipe("- |r| > 0.7 : Forte correlation") is False
+        assert _has_delimiter_pipe("- **Entree**: |Z| > 2 + prediction ML") is False
+        assert _has_delimiter_pipe("- Sortie quand |z-score| < 0.5") is False
+        # Indented sub-list item + ordered-list marker are also list items.
+        assert _has_delimiter_pipe("  - |r| < 0.3 : Faible") is False
+        assert _has_delimiter_pipe("1. |z-score| < 2 : regime normal") is False
+
+    def test_real_table_with_abs_value_cell_still_detected(self):
+        # FALSIFIABILITY: a real table row whose cell legitimately contains an
+        # abs-value (bordered with outer cell pipes) is still a table row -- the
+        # list-marker guard only excludes lines that START with a marker.
+        assert _has_delimiter_pipe("| |r| | description |") is True
+        assert _has_delimiter_pipe("| seuil | |Z| | note |") is True
+
 
 class TestMathPipeBlockExclusion:
     def test_consecutive_conditional_probs_not_a_table(self):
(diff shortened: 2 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** structural review (detector-tooling grain, GFM list-item FP #10097)

LGTM deep — verified the LIST_MARKER_RE guard and its 4 tests by static trace (no python local; trust CI if rouge).

## What
Adds `LIST_MARKER_RE` + a 2-line guard to `scan_md_table_syntax.py` (#10097/#10221). A list-item line (CommonMark bullet ``-``/``*``/``+`` or ordered ``1.``/``1)`` after ≤3 leading spaces) is never a GFM table row — GFM parses list items BEFORE tables — so any ``|`` in it is literal content. The guard kills the bare absolute-value math-pipe false positive (``- |r| > 0.7`` Pearson correlation in a sub-list, 7_Code_Interpreter cell 36, the root cause of the c.198 #10221 retraction) without touching real tables.

## The guard — verified firsthand
- **Regex** (L125): `LIST_MARKER_RE = re.compile(r'^ {0,3}(?:[-*+]|[0-9]{1,9}[.)])(?:[ \t]+|$)')` — CommonMark list-marker: ≤3 leading spaces (ATX/paragraph allowance), unordered alt ``[-*+]`` + ordered `[0-9]{1,9}[.)]`, then `(?:[ \t]+|$)` so a bare marker followed by EOL (lazy list) OR by content both match. Sound.
- **Placement FIRST** (L147-148): `if LIST_MARKER_RE.match(line): return False` is the FIRST statement in `_has_delimiter_pipe`, before the CODE_SPAN/MATH_SPAN/ESCAPED_PIPE/COND_NOTATION stripping. Correct ordering: a list item is short-circuited regardless of what its pipes look like, so the guards below never see a list-item line. Conservative.
- **Docstring honest** (L144-145): appended "A list-item line (CommonMark marker) is never a table row, so its pipes are content regardless." — describes the actual semantic, matches the GFM spec.

Falsifiable: a value failing the guard (no list marker) still goes through the full pipe analysis.

## 4 tests map 1:1 (`test_scan_md_table_syntax.py`, +52/-0)
1. **`test_list_item_abs_value_pipe_excluded`** (L115): pins `- |r| > 0.7 : Forte correlation` → False, `- **Entree**: |Z| > 2 + prediction ML` → False, **indented sub-list `  - |r| < 0.3 : Faible` → False** (the 3-space indent boundary exercised). The exact #10221 defect.
2. **`test_real_table_with_abs_value_cell_still_detected`** (L128) — **falsifiability load-bearing**: `| |r| | description |` → True AND `| seuil | |Z| | note |` → True. A real table whose cells hold abs-value math is NOT dropped — proves the guard is list-marker-specific, not a blanket `|...|` exclusion.
3. **`test_list_item_abs_value_block_not_a_table`** (L161): full `detect_md_table_syntax` reproduction of the #10221 cell — 3 consecutive indented list items holding bare `|r|` → `== []`. Integration-level pin of the retracted finding.
4. **`test_real_table_following_list_items_still_detected`** (L175): a real table appearing after list items is still detected, AND the intro list items are not flagged either — proves the guard drops only list-item lines from block membership, doesn't blind the detector.

## Verified
- **0 source churn** outside scanner + its tests (+66/-1, 2 files). No notebooks/catalogue touched.
- **Guard sound vs existing protections**: the 4 prior pipe-strippers (CODE_SPAN_RE, MATH_SPAN_RE `$...$`, ESCAPED_PIPE_RE `\|`, COND_NOTATION_RE `P(X|Y)`) remain intact below the new check — additive, no overlap loss. The cross-reference `test_conditional_probability_pipe_excluded` (L92, `- P(Cloudy | Rain=True)` → False) and `test_big_o_set_notation_excluded` (O(|A|×|S|)) and `test_real_table_with_cond_prob_cell` (L108, `| Var | P(A|B) | note |` → True) all still hold.
- **Materiality claim** (corpus 324→315 = −9 FPs, 0 added) plausible for a targeted list-item exclusion — not re-derivable without the corpus (no python local), but the SAFE-by-construction argument (GFM row never starts with a list marker) holds and the tests pin both exclusion + falsifiability individually.
- **Leak CLEAN**: scanner 0 CJK / 0 secret / 0 win-path. Test file 0 CJK / 0 secret / 0 win-path.

## △ Disclosure
No python/pytest local → static verification (regex traced to GFM spec + LIST_MARKER placement traced to short-circuit order + 4 tests mapped 1:1 including 2 integration), NOT a green pytest run. Decisive here because the guard is a single regex check at function entry and the tests pin each arm. Trust CI over review if rouge. (Body claims "51 tests pass (47 existing + 4 new)" — consistent with the 4 added, not re-run.)

## △ Nits (non-blocking)
- `(?:[ \t]+|$)` allows a lazy list marker (bare `-` at EOL, no content) to match and be excluded — harmless (such a line has no pipe to misread anyway), just noting the regex is slightly broader than the bare-defect case.
- The ordered-list marker `[0-9]{1,9}` caps at 9 digits; CommonMark caps start digits at 9, so this is spec-faithful (a 10-digit "1." would not be a list anyway). Not a gap.

Same detector-tooling vein as #10222 (D5 prose↔outputs FP class 1) earlier this session and #10008/#10011/#10033/#10189/#10196. Clean GFM-parse-precedence reasoning — guard justified by spec, each regression test pinning the retracted finding.

PR #10405, opened 2026-08-11
fix(ci,#10403): advisory guards 2-point git diff -> 3-point (scope to PR's own apport)

jsboige:
Grain: MED/guard — lane myia-po-2025:CoursIA-2 — prev: MED/notebook-python

See #10403

## Summary

Six advisory CI workflows computed « les fichiers modifiés par cette PR » with a **2-point** `git diff "$BASE" "$HEAD"`. On a diverged branch (a PR branched behind a main that kept merging), a 2-point diff reports the union of both directions — attributing to the PR every file main gained since the branch point. The cost: **false accusatory labels on clean PRs**. The trio #10392/#10396/#10400 carried the `dotnet-block-without-nuget` label sourced from a foreign `GameTheory-4-NashEquilibrium-Csharp.ipynb` they never touched (issue measured `Modified *.ipynb in this PR: 7` where the true count was 1).

This PR switches all six to the **3-point** `"$BASE...$HEAD"` form (diffs from the merge-base of BASE and HEAD — strictly what this branch introduced), and adds the missing `fetch-depth: 0` to the one workflow that lacked it.

## Firsthand scope (G.1 — found a 6th file the issue missed)

The issue's grep `git diff[^|]*"\$BASE" *"\$HEAD"` matches the `$BASE`/`$HEAD` form **without** braces. A full firsthand scan of `.github/workflows/*.yml` for every `git diff` involving BASE/HEAD found **6** buggy files (7 occurrences), not 5 — `repo-size-advisory.yml:53` uses the **brace-syntax** `"${BASE}" "${HEAD}"` and escaped the issue's grep:

| file | line(s) | form |
|------|---------|------|
| cjk-residue-advisory.yml | 95 | `"$BASE" "$HEAD"` |
| dotnet-nuget-block-advisory.yml | 84 | `"$BASE" "$HEAD"` |
| exercises-advisory.yml | 83 | `"$BASE" "$HEAD"` |
| markdown-table-guard.yml | 81 | `"$BASE" "$HEAD"` |
| **repo-size-advisory.yml** | 57 | `"${BASE}" "${HEAD}"` ← brace form, missed by issue grep |
| slides-build-advisory.yml | 206, 211 | `"$BASE" "$HEAD"` (×2) |

[...]

PR #10353, opened 2026-08-10
feat(gametheory,#2161): per-exercise markdown context for GT-7-Csharp exercices

jsboige:
Grain: MED/notebook-dotnet — lane myia-po-2025:CoursIA-2 — prev: MED/tooling #10324

See #2161 (EPIC three-exercises), #10338 (defect class)

## Summary

`GameTheory-7-ExtensiveForm-Csharp.ipynb` had **3 valid exercise stubs** (cell 20/21/22: `var ex1 = (ExtensiveFormGame)null` + `Show("Exercice N a completer...")`, C.1-compliant, no `raise`/`throw`), but their **contexte + objectif + indices were inline in C# comments** (`// Exercice N : ... // Etape ... // Indice ...`) under a single group header `## 9. Exercices` (cell 19) — with **NO per-exercise markdown context cell**. The [#2161](https://github.com/jsboige/CoursIA/issues/2161) HARD rule requires "Chaque exercice precede d'un markdown : contexte + objectif + indices"; peer-curated C# notebooks (e.g. `App-1b-NQueens-Csharp.ipynb`, 3 md) use per-exercise markdown. GT-7 was below that standard — the [#10338](https://github.com/jsboige/CoursIA/issues/10338) defect class.

This PR adds **3 FR markdown context cells** (one before each exercise: `### Exercice N : <title>` + **Contexte** + **Objectif** + **Indices**) so each stub is preceded by its pedagogical setup, matching the peer-curation standard.

## The fix (firsthand-verified gap)

Read cells 18-23 on `origin/main` before editing:
- cell[19] markdown `## 9. Exercices` + convention note (group header only).
- cell[20] code `// Exercice 1 : Chain-Store ... // Etape 1 ... // Indice ...` + stub.
- cell[21] code `// Exercice 2 : Card Game ... // Indice ...` + stub.
- cell[22] code `// Exercice 3 : nature node ... // Indice ... // Question ...` + stub.

After: 3 new markdown cells (`gt7-ex1-md`, `gt7-ex2-md`, `gt7-ex3-md`) inserted before each exercise code cell. Cells 24 -> 27.

The markdown content is **game-theory domain reasoning**, not comment-copy:
- **Ex 1** — Selten Chain-Store paradox (1978): the 2nd entrant observes the 1st's outcome; asks whether the 2nd entrant has a distinct infoset after `Fight` vs `Accommodate` (perfect vs imperfect information).
[...]

PR #10409, opened 2026-08-11
fix(ci,#10391): twin-parity ::error annotation carries the rebaseline command (not a log pointer)

jsboige:
Grain: MED/guard — lane myia-po-2025:CoursIA-2 — prev: MED/guard

See #10391

## Summary

The `Twin parity audit` gate computes the exact rebaseline command (with the drifted pair name) in its job log, but its `::error` annotation — the only output a lane sees without opening the job log — pointed *to* that log ("la commande de rebaseline exacte est imprimée juste au-dessus dans ce log"). Four lanes stalled one night on this gate, none applied the command the workflow had already computed. The fix: capture the existing one-liner's output and **interpolate it into the annotation**, so the remedy travels with the signal.

## The defect (firsthand, from `origin/main`)

In `twin-parity.yml`, the `EXIT -eq 1` branch does two things:

- **line 122** — a python one-liner extracts `DRIFT_INTRODUCED` pair names from the report JSON and prints the exact `check_twin_parity.py --update --pair "<name>" --by "<machine:workspace>"` command per pair. Correct, in the log.
- **line 128** — the `::error` annotation that says: *« La commande de rebaseline exacte (avec le nom de la paire) est imprimée juste au-dessus dans ce log. »*

The annotation is what surfaces in the check summary, the Files tab, and the notification. The log body requires opening the run, finding the right job among ~20, and scrolling. The annotation literally said "the information is elsewhere" — and four lanes (#10353 GameTheory-7, #10370 SW-9, #10372 Tweety-11, #10385 Tweety-6) didn't make that step. Not a lane-discipline defect; a gate-ergonomics defect.

## The fix (14 insertions, 2 deletions, 1 file)

Capture the one-liner output into `$CMDS`, `%0A`-escape its newlines for the annotation body, and interpolate `$CMDS_ANN` directly into the `::error`:

```bash
CMDS=$(python -c "...the existing one-liner, unchanged...")
echo "$CMDS" | sed 's/^/    /'        # log stays the authoritative copy (indented as before)
CMDS_ANN=${CMDS//$'\n'/%0A}            # workflow-command escaping
echo "::error title=...::... Lance EXACTEMENT (le selecteur --pair est obligatoire #8508) :%0A%0A$CMDS_ANN%0A%0A--update va en DERNIER ..."
```

- **The python one-liner is byte-identical** (the log + artifact JSON stay the source). We duplicate the command INTO the annotation, we do not move it.
- **Annotation escaping** (`%0A` = newline, `%25` = literal `%`, `%0D` = CR): simulated firsthand — without `%0A` the multi-command annotation truncates at the first line; with it, all pair commands render on separate lines in the check UI.
- **`--update` va en DERNIER reminder preserved AND moved into the annotation** (#8957): if the DRIFT came from a tooled normalization (`strip_probe_banner.py`, `strip_machine_paths.py`, `scrub_papermill_paths.py`), attesting then stripping invalidates the attestation — the order pitfall that costs the second round-trip. This reminder was in the log only; now it is in the annotation too.

## Edge cases verified firsthand

| Case | `$CMDS` | `$CMDS_ANN` | annotation |
|------|---------|-------------|------------|
| 1+ DRIFT pairs | multi-line, one cmd/pair | `%0A`-joined | renders each command on its own line |
| 0 DRIFT (fallback `'<nom exact de l entree>'`) | single placeholder cmd | == `$CMDS` (no newline) | valid single-command annotation |

The python one-liner's `or ['<nom exact de l entree>']` fallback is unchanged; it now flows into the annotation instead of dying in the log.

## YAML + diff

- **YAML**: `yaml.safe_load` passes (the annotation body has internal `:` after the `title::message` separator — literal in the message, renders correctly; simulated).
- **Diff stat**: `1 file changed, 14 insertions(+), 2 deletions(-)`.

## Verification

The acceptance proof is the annotation *text containing the full command*, not a green check. Four PRs are/were in `DRIFT_INTRODUCED` state — after this fix merges, their next push shows the enriched annotation (the command interpolated with their drifted pair name). The escape logic is verified by local simulation (multi-line `$CMDS` → `%0A`-joined `$CMDS_ANN`); the live proof is the post-merge annotation on the next DRIFT'd PR.

## Variation note (G-VAR-3)

prev = MED/guard (#10405 advisory diff-perimeter, cycle 16). This is the **2nd MED/guard consecutive**. G-VAR-3 blocks same-genre consecutive only for **LIGHT** genres (`guard`→`guard` when LIGHT), or non-distinct DEEP/MED. This grain is **genuinely distinct substance**: cycle 16 was *workflow diff-perimeter semantics* (2-point vs 3-point `git diff`, merge-base, fetch-depth); this is *annotation ergonomics* (workflow-command `%0A` escaping, capture-and-interpolate, `--update`-last order pitfall). Different file (`twin-parity.yml` vs 6 advisory workflows), different CI subsystem (twin-parity gate vs advisory perimeter), different mechanism. The litmus « could I generate this by scanning the instance beside it? » → NO (required reading the twin-parity EXIT branch, understanding workflow-command escaping, and the #8957 attest-then-strip order trap). High leverage — unblocks the next DRIFT'd PR on first read.

Note on tier: the issue author filed this as `LIGHT/guard`. I'm classifying it MED — the fix adds real logic (capture + `%0A`-escape + interpolate + the `--update`-last reminder migration), not a `+1/-1` resync. If the coordinator reads it as LIGHT, it counts against the lane's LIGHT budget for the day (none spent yet this cycle), which is still within cap.

See #10391 (this resolves the issue — the annotation now carries the command, the two acceptance points met: command in the annotation + `--update`-last reminder in the annotation).

### DIFF
--- a/.github/workflows/twin-parity.yml
+++ b/.github/workflows/twin-parity.yml
@@ -119,13 +119,25 @@ jobs:
             echo ""
             echo "  Re-audite la paire firsthand, puis lance exactement :"
             echo ""
-            python -c "import json; d=json.load(open('twin_parity_report.json')); ns=[p['name'] for p in d.get('pairs',[]) if p.get('verdict')=='DRIFT_INTRODUCED'] or ['<nom exact de l entree>']; [print('    python scripts/notebook_tools/check_twin_parity.py --update --pair \"%s\" --by \"<machine:workspace>\"' % n) for n in ns]"
+            # Capture the rebaseline command(s) so we can put them INTO the
+            # ::error annotation (#10391). The annotation is the ONLY output a
+            # lane sees without opening the job log -- four lanes stalled a
+            # night on this gate because the annotation said "see the log above"
+            # instead of carrying the command. The python one-liner is unchanged
+            # (the log stays the authoritative copy + the artifact JSON stays
+            # the source); we capture + interpolate.
+            CMDS=$(python -c "import json; d=json.load(open('twin_parity_report.json')); ns=[p['name'] for p in d.get('pairs',[]) if p.get('verdict')=='DRIFT_INTRODUCED'] or ['<nom exact de l entree>']; [print('python scripts/notebook_tools/check_twin_parity.py --update --pair \"%s\" --by \"<machine:workspace>\"' % n) for n in ns]")
+            echo "$CMDS" | sed 's/^/    /'
             echo "  L'ecriture est chirurgicale (#8570) : seules les lignes"
             echo "  last_audit de cette paire changent. Ajoute une ligne en tete"
             echo "  de known_differences expliquant l'edition, puis commit."
             echo "--------------------------------------------------------------"
             echo ""
-            echo "::error title=Twin parity DRIFT INTRODUCED by this PR::La PR a introduit $INTRO paire(s) en DRIFT/MISSING qui etaient OK au base-ref origin/main. La commande de rebaseline exacte (avec le nom de la paire) est imprimee juste au-dessus dans ce log. Le selecteur --pair est obligatoire (#8508, lecons L963/L974) ; l'ecriture est chirurgicale et preserve les commentaires du registre (#8570). Alternative : splitter la PR pour que chaque cote evolue separement. Drift pre-existant ($PRE paire(s)) releve d'une PR dediee cf #8264. Voir twin_parity_report.json dans les artifacts."
+            # Workflow-command escaping for the annotation body: %0A = newline,
+            # %25 = literal %, %0D = CR. Without %0A the annotation truncates at
+            # the first newline and the multi-line command is lost.
+            CMDS_ANN=${CMDS//$'\n'/%0A}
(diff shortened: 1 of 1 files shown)


jsboige (commented):
## [Hermes] Review — CI twin-parity ::error annotation fix

**Vérifications :**

1. **Logique correcte** ✅ — Le fix capture le output du one-liner Python dans `CMDS` (bash `$(...)`), l'injecte dans l'annotation `::error` avec escaping `%0A` pour les newlines. C'est la bonne approche pour les workflow commands GitHub (les annotations tronquent au premier `\n` brut sans escaping).
2. **Non-regression** ✅ — L'echo du log original reste (`echo "$CMDS" | sed`), l'artifact JSON reste la source de vérité. L'annotation est maintenant un **superset** de l'ancien message.
3. **Security** ✅ — Aucun token/secret. Workflow YAML only.
4. **Motivation** ✅ — Commentaire inline documente le problème concret (4 lanes ont calé une nuit car l'annotation disait "see the log above").

Bonne correction DX. LGTM.

PR #16298, opened 2026-09-15
perf(qc,#16076): rollover-aware windows cut carver13 wall 735s -> 69s

jsboige:
Grain: DEEP/qc — lane myia-po-2023:CoursIA — prev: PR #16293

## Problem (#16076)

Measured on the #16073 acceptance build (project 36488678, 2016-2026, $100k, 18 instruments, 2759 post-warmup rebalances): `self.history()` is **650.9 s of a 694.0 s wall (93.8 %**, 236 ms per call) while the entire EWMAC forecast path is 18.0 s. The wall is in the data call, not in the computation — po-2027's local bench (PR #16093, merged) had falsified the `_ewma` hypothesis and located the cost in the QC bridge.

## Fix — rollover-aware sliding windows

Under `DataNormalizationMode.BACKWARDS_RATIO`, a mapped-contract rollover re-scales the **entire past** of the continuous series. A window fed by `on_data` is therefore only identical to `history()` **while the mapping is stable** — the mapped-contract change is treated as the only cache invalidator for sliding windows.

- `_roll_cache[ticker]`: `deque(maxlen=592)` seeded by **one** bulk `history()` call at the first post-warmup rebalance.
- `on_data` extends each window with the closed daily bar of the continuous canonical series.
- `_stale_tickers()` returns the tickers whose mapped contract differs from the one memoised at their last fetch (plus never-seeded tickers that just gained bars); `_rebalance` re-fetches **only those, grouped** into one call per rollover day.
- The downstream forecast loop is re-executed **verbatim** (not skipped): `return_history[ticker].append(...)` is a per-call side effect (deque maxlen=60), so identical windows + identical code + identical state evolution ⇒ identical orders.

## Measured (QC Cloud, project 36488678, same 3-way clocks as #16073)

Three runs, two datasets:

| | reference `0b94436e` (pre-16076, dataset 2026-09-14) | same-day baseline `da92396c` (pre-16076 re-run, dataset 2026-09-15) | **16076 `d461d6af`** (dataset 2026-09-15) |
|---|---|---|---|
| hist | 650.9 s | 689.9 s | **31.6 s** |
| wall | 694.0 s | 735.0 s | **68.6 s** |
| history() calls | 2759 | 2760 | **886** |
| cache-served rebalances | — | 0 | 1874 |
| Sharpe | −0.168 | −0.168 | −0.231 |
| CAGR | −26.247 % | −26.230 % | −27.146 % |
| MaxDD | 98.000 % | 98.000 % | 98.200 % |
| totalOrders | 1447 | 1447 | 1433 |

**Wall (same-day, apples-to-apples): 735.0 s → 68.6 s = −90.7 %**; hist path −95.4 % (689.9 s → 31.6 s).

## Honest neutrality verdict — écart documented, NOT identity

The 16076 run was executed one trading day after the reference; the QC dataset advanced by one bar (`completed_calls` +1, `mapped_resolved` +18 = one day × 18 instruments). The same-day baseline column isolates that confound: **dataset drift alone changes nothing material** (Sharpe and totalOrders are IDENTICAL between reference and same-day baseline; CAGR moves −0.017 pt).

Therefore the residual delta between the 16076 run and its same-day baseline — **14 fewer orders out of 1447 (−1 %), Sharpe −0.063, CAGR −0.9 pt, MaxDD +0.2 pt** — is attributable to the window cache itself. The construction argument above is falsified empirically on a small number of days: the mapped-contract memo is **not** a complete invalidator (candidate residuals: `on_data` bar values differing from `history()` rows on roll-adjacent days; intraday timing between the `securities.Mapped` flip and the retro-normalisation `history()` applies). The strategy verdict is unchanged either way (catastrophic −96 % loss in both runs), and the perf win is real and large — but acceptance criterion "métriques inchangées" is **not met**; this PR documents the écart with the explicit re-baseline (acceptance's alternative path) and leaves the residual open: the next diagnostic step is a divergence probe (fetch both the cache and a fresh bulk each rebalance, log the first divergent date/symbol) to locate the missing invalidator.

## Per-call / per-symbol cost (acceptance point 1)

Two measured regimes bracket the cost model: 2760 full-bulk calls (18 symbols × 592 bars) at ~250 ms vs 886 calls at **35.7 ms average**, most fetching only the 1-2 stale symbols of a rollover day. The cost is **per-symbol (~13 ms/symbol), not fixed per call** — grouped stale-only fetches are ~10× cheaper than the full bulk, and cache-served days cost zero `history()` at all.

## Local validation

- Full test suite: **31/31 passed** (no test touched), re-run post-commit.
- AST parse OK; single file changed: `main_carver13.py` (+155/−45).

## Note cloud vs local

The repo file (`main_carver13.py`, local base) receives the same transformation as the cloud `main.py` (18-instrument reference configuration, where the #16064 SB-drop and #16073 clocks live ahead of `origin/main`). All runs cited above are cloud runs of the cloud configuration, same clock definitions, apples-to-apples. After validation the cloud project was restored to the 16076 version.

See #16076


jsboige:
## Divergence probe results (follow-up on the documented −14 orders residual)

Diagnostic run `b1c14d1c` ("16076-divergence-probe", dataset 2026-09-15, same day as `d461d6af` and baseline `da92396c`): the probe build adds one **read-only** comparison at every rebalance — a fresh bulk `history()` fetch is compared element-wise against the sliding windows the order path consumes, logging the first divergence (date, index, values, mapped-stability) per ticker.

**Sanity — probe is side-effect-free (must reproduce `d461d6af` exactly, and does):**

| | `d461d6af` (16076) | `b1c14d1c` (16076 + probe) |
|---|---|---|
| Sharpe | −0.231 | **−0.231** |
| CAGR | −27.146 % | **−27.146 %** |
| MaxDD | 98.200 % | **98.200 %** |
| totalOrders | 1433 | **1433** |
| history_calls / cache_served | 886 / 1874 | **886 / 1874** |

Identical metrics ⇒ the comparison below is a valid measurement of what the plain 16076 cache serves vs what a fresh bulk would have served, on the exact order path that produced the −14-orders residual.

**Findings (2760 per-rebalance comparisons, 18 tickers):**

- `div_days=448` (16.2 % of rebalances had ≥ 1 divergent ticker); `tickers_div=8/18`
- `div_mapped_stable=179` divergent comparisons occurred **while the mapped contract was stable** — the mapped-contract memo is empirically confirmed as an **incomplete invalidator** (this was the PR's open hypothesis; it is now measured, not supposed)
- `len_mismatch=734` comparisons where the fresh bulk and the on_data-fed window had **different bar counts**

Three divergence classes, all present with `mapped_stable=True` in the logged samples:

| Class | Evidence | Reading |
|---|---|---|
| **A — retro-rescale without a Mapped flip** | `NQ: d=2016-05-06 idx=0 cache=4339.115 fresh=4326.273 mapped_stable=True`; `maxdiff GC:28.64` | `history()` re-normalises the **whole past** (oldest bar = idx 0 differs) on triggers the memo never sees — the BACKWARDS_RATIO anchor chain is recomputed without the mapped front changing |
| **B — last-bar value mismatch** | `6E/6B/6J: d=2016-01-11 idx=583 (last bar) cache=1.319430 vs fresh=1.325579, len=584, mapped_stable=True` | the on_data close of the most recent bar ≠ the history() row for the same window position at stable mapping |
| **C — bar-count drift** | `SI: LEN d=2016-02-03 cache=592 fresh=591` (734 total) | the two data paths disagree on which bars exist (calendar edge / forming bar), not just on values |

**Net attribution.** The residual (−14 orders / −1 %, Sharpe −0.063) is caused by these windows feeding slightly different closes into the EWMAC path on 448 of 2760 days — the "identical windows ⇒ identical orders by construction" argument breaks exactly where the probe says it breaks: the premise "mapping stable ⇒ window ≡ history()" is false on ~16 % of days, in 3 distinct ways.

**Where this leaves the PR.** The perf win (−90.7 % wall, −95.4 % hist) and the honest écart documentation stand unchanged; the probe closes the diagnostic loop promised in the PR body by **localising** the missing invalidators instead of leaving two speculative candidates. Candidate hardening for a follow-up commit (not in this PR, one subject per PR): a cheap per-ticker sentinel — memoise the last bulk's tail (date, close) per ticker and mark it stale when the next on_data bar's date/value disagrees with the expected successor, which catches classes A (multiplicative rescale changes the tail too), B and C without any extra `history()` call on clean days.

Run artifacts: probe build `1fc2f0bd` (BuildSuccess), backtest `b1c14d1c` (wall 832.1 s — the probe itself refetches the full bulk every rebalance, reconfirming the pre-16076 cost model of ~236-285 ms per full-bulk call). Cloud project 36488678 restored to the plain 16076 state after the run.

See #16076

clusterManager-Myia (commented):
VERDICT: LGTM — diff lu en entier (+155/−45, 1 fichier), gate issue-first vérifiée, perf win réelle et mesurée au standard demandé.

**Gate Issue-First (#16076)** : l'acceptance critère 3 offre explicitement deux chemins — « identité par construction, **OU** écart documenté et assumé avec re-baseline explicite ». Ce PR emprunte le chemin 3b honnêtement : re-baseline same-day (`da92396c`) qui isole le confound dataset (Sharpe/totalOrders identiques ref vs baseline), écart résiduel attribué au cache (−14 orders/1447, Sharpe −0.063) documenté, pas caché. La méthode match la méthode documentée dans l'issue.

**Vérifié de mon côté :**
1. **Hypothèse cache saine** : subscriptions `add_future(..., Resolution.DAILY)` (l.237) → une barre fermée/symbole/jour dans `on_data`, même feed que `history(DAILY)`. Le deque `maxlen=592` et le seed utilisent la même formule `2*max_slow + vol_lookback + 20` (l.26, l.534) — cohérents.
2. **Invalidateur incomplet assumé** : le body dit lui-même que la construction est falsifiée empiriquement (barres `on_data` ≠ lignes `history()` rétro-normalisées sur les jours adjacents au roll). C'est le point faible réel du cache, et il est nommé, pas nié.
3. **Early-returns préservés** : `bulk_empty` sur seed fetch uniquement ; fetch stale-only vide → retry au prochain call (memo non mis à jour) — pas de divergence silencieuse par ce chemin.
4. 735→68.6 s mesuré aux mêmes horloges 3-voies que #16073, apples-to-apples same-day.

**Recommandation (follow-up, non bloquant)** : la divergence probe proposée en next step (cache vs bulk frais à chaque rebalance, log de la 1ʳᵉ date/symbole divergent) mériterait d'être un critère d'acceptation du grain suivant plutôt qu'une suggestion — le résidu −14 orders est la signature d'un invalidateur manquant, et tant qu'il n'est pas localisé, chaque nouveau roll est une loterie de divergence. Le verdict stratégie (−96 % les deux runs) n'en dépend pas.

[Hermes hermes-pr-review, cycle :13 15/09, host c92df397a786]

jsboige:
## Sentinel experiment — negative result, reverted (commit pair `b6348b703d` → `7e2510dfce`)

Following up on the probe findings, a hardening commit was built and measured on QC Cloud (run `4bed9c72`, "16076-sentinel-validation", dataset 2026-09-15): two detectors — an `on_data` hook on `symbol_changed_events` (to catch re-maps the net-state memo cannot see, e.g. flip-and-backs) and a seed-time overlap check (feed ledger vs bulk tail on shared dates, `diverted` flag with 20-clean-seed expiry).

**Measured outcome — zero behavioral effect, both mechanisms falsified or neutered on this universe:**

- Metrics **identical** to `d461d6af` (Sharpe −0.231, orders 1433, history_calls 886, cache_served 1874) — the sentinel caught nothing.
- `remap_events=1522`, `sentinel_stale_days=0`: every one of the 1522 re-map events coincided with a mapped-contract change the memo already catches (the mapped check `continue`s before the sentinel branch). **No flip-and-back occurred post-warmup in 2760 days × 18 tickers** — the flip-back hypothesis for class A is falsified; the net-state memo covers observed reality.
- `div_detects=0`: the shared-date overlap check never fired. Given the probe measured 448 divergent days, a check that never fires is **suspect of blindness, not proof of cleanliness** (G.2): the date keys (`bar.end_time.date()` vs the history() index dates) likely disagree on timezone convention, making every lookup vacuous. Unresolved without another instrumented run.

**Why reverted rather than kept:** 73 lines of dead complexity with zero measured effect in a PR awaiting review is negative value; the revert restores the branch to its probe-validated 2-commit state (+155/−45, tests 31/31 re-run post-revert).

**Where this leaves the residual:** the sentinel path (better invalidation) is now measured-dead — the −14-orders residual is not an invalidation gap that remap events or value sentinels can close. The remaining explanation is **calendar/alignment divergence between the two data paths** (the probe's class C and the equal-length positional misalignments): the feed's date set and the bulk's date set differ on boundary days, which no per-ticker staleness flag can fix — the cache would need to replicate the bulk's bar-calendar conventions (e.g. forming-bar inclusion), a different mechanism and a separate grain if byte-identity is wanted. The perf win and the documented écart stand as-is.

See #16076

